PR #8, opened 2016-11-23
build: Generate wrapper functions on the fly

ueno:
Generate wrapper functions in p11-kit/virtual.c from the
definitions in common/pkcs11.h and common/pkcs11i.h, rather than
hard-coding them.

This is a spin-off patch from:
https://bugs.freedesktop.org/show_bug.cgi?id=97611


ueno:
I withdraw this, as I realized that there are some cases we need to adjust the generated wrapper functions.

PR #9, opened 2016-11-24
build: Make libffi closure optional

ueno:
libffi's closure support is not available on all platforms and may fail
at run time if running under a stricter SELinux policy.  Fallback to
pre-compiled closures if it is not usable.

This depends on p11-glue/p11-kit#8.


nmav:
A question I had is what should happen when the limit is reached? I've extended test-managed to cover for that, but I reach an assert() inside the library. Is that expected, or am I doing something wrong?
[limit-test.txt](https://github.com/p11-glue/p11-kit/files/616742/limit-test.txt)

nmav:
The gnutls test suite passes with the changes above.

ueno:
Thanks for the review.  I have updated the patches reflecting the comments, and incorporated the test into this PR.

configure.ac
```diff
@@ -195,22 +195,19 @@ if test "$with_libffi" != "no"; then
...
-	                  ])],
-	[], [AC_MSG_ERROR([the libffi on this system has no support for closures.])])
-	CFLAGS="$SAVE_CFLAGS"
-
 	with_libffi="yes"
 fi
 
 AM_CONDITIONAL(WITH_FFI, test "$with_libffi" = "yes")
 
+AC_ARG_WITH([fixed-closure],
```
ueno: Yes, changed to ```--with-max-managed-modules```.

p11-kit/virtual.c.template
```diff
@@ -46,7 +46,7 @@
 #include <stdlib.h>
 #include <string.h>
```
ueno: Updated.

ueno:
As I abandoned #8, I rewrote this without the generation script.

PR #14, opened 2016-11-25
trust: Avoid confusion in DER/PEM decoding

ueno:
Previously p11-kit-trust.so tried to interpret certificate as PEM format
first.  This could cause potential conflict if the certificate were
actually in DER format and contained a PEM marker strings.

https://bugs.freedesktop.org/show_bug.cgi?id=92063


nmav:
The test-compat seems to fail on f25 because /tmp/ is mounted as nosuid.

ueno:
Thanks for the review.

PR #16, opened 2016-11-28
Fixes in TMPDIR handling

nmav:
This patch set allows using alternatives to /tmp for writing temp files and switches travis build to ./ for temp directory.

### DIFF
--- a/common/test.c
+++ b/common/test.c
@@ -483,11 +483,12 @@ char *
 p11_test_copy_setgid (const char *input)
 {
 	gid_t groups[128];
-		char *path;
-		gid_t group = 0;
-		int ret;
-		int fd;
-		int i;
+	char *path;
+	gid_t group = 0;
+	int ret;
+	int fd;
+	int i;
+	char *tmpdir;
 
 	ret = getgroups (128, groups);
 	for (i = 0; i < ret; ++i) {
@@ -501,8 +502,11 @@ p11_test_copy_setgid (const char *input)
 		return NULL;
 	}
 
-	path = strdup ("/tmp/test-setgid.XXXXXX");
-	assert (path != NULL);
+	tmpdir = getenv ("TMPDIR");
+	if (tmpdir == NULL)
+		tmpdir = "/tmp";
+
+	assert (asprintf (&path, "%s/test-setgid.XXXXXX", tmpdir) >= 0);
 
 	fd = mkstemp (path);
 	assert (fd >= 0);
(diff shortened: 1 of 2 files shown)


common/test.c
```diff
@@ -501,7 +503,12 @@ p11_test_copy_setgid (const char *input)
 		return NULL;
 	}
 
-	path = strdup ("/tmp/test-setgid.XXXXXX");
+	tmpdir = getenv("TMPDIR");
+	if (tmpdir == NULL)
+		tmpdir = "/tmp";
+
+	snprintf(file, sizeof(file), "%s/test-setgid.XXXXXX", tmpdir);
+	path = strdup (file);
```
ueno: The construction of ```path``` could be simplified using ```asprintf```, which we use extensively and provide a fallback definition for portability.

nmav:
Updated to use asprintf() instead.

common/test.c
```diff
@@ -483,11 +483,13 @@ char *
...
-		gid_t group = 0;
-		int ret;
-		int fd;
-		int i;
+	char *path;
+	gid_t group = 0;
+	int ret;
+	int fd;
+	int i;
+	char file[256];
```
ueno: This is no longer necessary.

common/test.c
```diff
@@ -501,8 +503,11 @@ p11_test_copy_setgid (const char *input)
 		return NULL;
 	}
 
-	path = strdup ("/tmp/test-setgid.XXXXXX");
-	assert (path != NULL);
+	tmpdir = getenv("TMPDIR");
+	if (tmpdir == NULL)
+		tmpdir = "/tmp";
+
+	assert(asprintf(&path, "%s/test-setgid.XXXXXX", tmpdir) >= 0);
```
ueno: It would be good to avoid side-effect in ```assert```, since the code can be omitted at compile time, when NDEBUG is set. Also please put a space after function name for style consistency.

nmav: I've added the space, however I am not sure about the second point. I see that the test suite is built around assert(), and that construction in most of the tests. Does it make sense to avoid assert on this particular test?

common/test.c
nmav: Removed

common/test.c
ueno: I meant to change it to:
```c
ret = asprintf (...);
assert (ret >= 0);
```
so that asprintf is still called even if the assert statement becomes no-op.

Anyway, thank you for the update, though I'm leaning toward #17.

nmav:
#17 seems to be simpler.

PR #17, opened 2016-11-28
test-conf: don't create the setuid copy in /tmp

lkundrak:
The temporary directory is often mounted with nosuid, thus whatever runs
from there doesn't get AT_SECURE in auxv.

### DIFF
--- a/common/test.c
+++ b/common/test.c
@@ -501,7 +501,7 @@ p11_test_copy_setgid (const char *input)
 		return NULL;
 	}
 
-	path = strdup ("/tmp/test-setgid.XXXXXX");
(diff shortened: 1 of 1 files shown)


ueno:
Actually #16 fixes the same issue by replacing ```/tmp``` with ```TMPDIR```, while in this context it might make more sense to hard-coding it.

lkundrak:
Indeed. I first didn't notice the repository moved (maybe you could add a commit that replaces everything with a notice to the original repository?) and the old repository didn't contain Nikos' commit.

Nevertheless, this is still probably not a good use of a temporary directory. Even the mkstemp() is probably superfluous.

PR #15, opened 2016-11-25
Add socket based server

ueno:
This pull request adds a new ```p11-kit server``` command, which provides a Unix domain socket based server to expose access to smartcards.

PR #18, opened 2016-11-28
Improve usefullness of remoting

lkundrak:
This is in general done to allow forwarding the modules from the desktop session (such as the GNOME keyring or physical tokens the console user can access) to NetworkManager helpers (think VPN daemons or wpa_supplicants) running in system scope.

Related: https://bugzilla.gnome.org/show_bug.cgi?id=767872

### DIFF
--- a/common/compat.c
+++ b/common/compat.c
@@ -171,7 +171,7 @@ p11_mutex_init (p11_mutex_t *mutex)
 	int ret;
 
 	pthread_mutexattr_init (&attr);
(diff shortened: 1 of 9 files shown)


lkundrak:
I've removed the fd remote from the patchset. Stef suggested using a UNIX socket instead and that in fact might be a better idea.

I'm keeping the rest here until it's ready (and the fd remote will probably be completely unnecessary): https://github.com/NetworkManager/p11-kit/tree/lr/NM-remoting2

lkundrak:
Discarding this. It has a large overlap with #15 -- let's just continue there.

PR #26, opened 2016-12-14
filter: New virtual wrapper for access control

ueno:
This is a dependency of #15.  When we forward the Unix domain socket to another machine, it would be good to limit the access to a specific token, rather than exposing the entire module and all the tokens managed by that module.

Maybe this could be used together with the PKCS#11 URIs, something like:
```
p11-kit server <token URI> ...
```
instead of:
```
p11-kit server <module filename> ...
```

### DIFF
--- a/doc/manual/Makefile.am
+++ b/doc/manual/Makefile.am
@@ -52,6 +52,7 @@ IGNORE_HFILES= \
 	conf.h \
 	debug.h \
 	dict.h \
+	filter.h \
 	log.h \
 	mock.h \
 	modules.h \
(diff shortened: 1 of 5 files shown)


ueno:
Other use-case could be to blacklist certain driver in a PKCS#11 module.  That could be done through a module configuration, something like "allow-token" or "deny-token" with a PKCS#11 URI as the argument.

ueno:
Rewritten using the new P11KitIter API (#28) for enumerating tokens.

ueno:
Merged this, as it is merely an internal change.

PR #28, opened 2016-12-19
iter: Enable iteration over slots/tokens/modules

ueno:
While PKCS#11 URI can identify slots/tokens/modules, P11KitIter is only
capable of iterating over objects.

This patch adds new behaviors to P11KitIter to support iterations over
slots/tokens/modules, using the C coroutine trick as described in:
http://www.chiark.greenend.org.uk/~sgtatham/coroutines.html


ueno:
Merging this, assuming the API change doesn't introduce any incompatibility.

PR #31, opened 2016-12-28
uri: fix dealing with the query part

lkundrak:
The current code accepts and produces the query attributes as path attributes. That's wrong (and perhaps dangerous, when used recklessly). Hopefully I'm improving it a bit here.

Note that this potentially breaks compatibility with existing (broken!) URIs. I think that's okay, but perhaps need some consideration.

Please consult the grammar in section 2.3. for details: https://tools.ietf.org/html/rfc7512#section-2.3

### DIFF
--- a/p11-kit/uri.c
+++ b/p11-kit/uri.c
@@ -753,21 +753,32 @@ p11_kit_uri_new (void)
 	return uri;
 }
 
+enum uri_sep {
+	sep_path = '\0',
+	sep_pattr = ';',
+	sep_query = '?',
+	sep_qattr = '&',
+};
+
 static void
 format_name_equals (p11_buffer *buffer,
-                    bool *is_first,
+                    enum uri_sep *sep,
                     const char *name)
 {
-	if (!*is_first)
-		p11_buffer_add (buffer, ";", 1);
+	if (*sep)
+		p11_buffer_add (buffer, sep, 1);
 	p11_buffer_add (buffer, name, -1);
 	p11_buffer_add (buffer, "=", 1);
-	*is_first = false;
+
+	if (*sep == sep_path)
+		*sep = sep_pattr;
+	else if (*sep == sep_query)
+		*sep = sep_qattr;
 }
 
 static bool
 format_raw_string (p11_buffer *buffer,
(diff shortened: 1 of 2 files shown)


p11-kit/uri.c
```diff
@@ -1433,9 +1436,44 @@ p11_kit_uri_parse (const char *string, P11KitUriType uri_type,
...
+		}
+
+		epos = strchr (string, '=');
+		if (epos == NULL || spos == string || epos == string || epos >= spos) {
+			free (allocated);
+			return P11_KIT_URI_BAD_SYNTAX;
+		}
+
+		ret = 0;
+		if (ret == 0)
```
ueno: This seems redundant.

p11-kit/uri.c
```diff
@@ -1393,17 +1393,22 @@ p11_kit_uri_parse (const char *string, P11KitUriType uri_type,
 	uri->pin_value = NULL;
 	uri->slot_id = (CK_SLOT_ID)-1;
 
+	/* Parse the path. */
```
lkundrak: Well, I can't really tell how to merge them :(
i basically split it because it started to look bad to do a lot of special casing there.

p11-kit/uri.c
```diff
@@ -1393,17 +1393,22 @@ p11_kit_uri_parse (const char *string, P11KitUriType uri_type,
...
 	for (;;) {
 		spos = strchr (string, ';');
+		if (!spos || (strchr (string, '?') && strchr (string, '?') < spos)) {
```
lkundrak: Could I have not ever heard of strcspn()?

p11-kit/uri.c
lkundrak: Right.

lkundrak:
Pushed an updated branch.

ueno:
Thank you, merged.

By the way, the URI parser in [engine_pkcs11](https://github.com/OpenSC/engine_pkcs11/wiki) seems to have the same issue with [pin-value](https://github.com/OpenSC/libp11/blob/master/src/eng_parse.c#L306), which I guess would be nice to fix.

PR #35, opened 2017-01-16
systemd: add per-user remoting socket

lkundrak:
This allows daemons outside user's session to use per-user PKCS#11
modules. Useful for letting VPN daemons or wpa_supplicant use
certificates stored in user's GNOME keyring, etc.

### DIFF
--- a/p11-kit/Makefile.am
+++ b/p11-kit/Makefile.am
@@ -93,12 +93,22 @@ install-exec-hook:
 	done
 	$(MKDIR_P) $(DESTDIR)$(p11_package_config_modules)
 
+install-data-hook:
+	$(MKDIR_P) $(DESTDIR)$(systemduserdir)/sockets.target.wants
+	$(LN_S) -f ../p11-kit-remote.socket $(DESTDIR)$(systemduserdir)/sockets.target.wants/p11-kit-remote.socket
+
 uninstall-local:
 	for i in so dylib; do \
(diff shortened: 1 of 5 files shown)


lkundrak:
This has some overlap with 4e05b6d64. The key difference is that this exports all tokens via the proxy module, so that the consumer can pick the right token.

Also, this doesn't do credential validation, but the UNIX file mode seems to do the right job allowing only the socket user or superuser with DAC override capability to connect. On the other hand, this offloads the job of managing the socket to systemd and makes it possible to only spawn the p11-kit-remote on-demand, only when needed.

p11-kit/Makefile.am
```diff
@@ -93,13 +93,23 @@ install-exec-hook:
...
 	for i in so dylib; do \
 		rm -f $(DESTDIR)$(libdir)/p11-kit-proxy.$$i; \
 	done
+	rm -f $(DESTDIR)$(systemduserdir)/sockets.target.wants/p11-kit-remote.socket
 
 endif
 
+systemduserdir = $(prefix)/lib/systemd/user
+systemduser_DATA = \
+	p11-kit/p11-kit-remote.socket \
```
ueno: I think this file (p11-kit/p11-kit-remote.socket) should be added to EXTRA_DIST, while p11-kit-remote\@.service.in is automatically added to the distribution as it is listed as AC_CONFIG_FILES.

lkundrak: Correct. Added it.

PR #38, opened 2017-01-18
rpc: New rpc_unix transport based on Unix socket

ueno:
This is split from #15 and basically identical to https://github.com/NetworkManager/p11-kit/commit/fbd884b40f1763c1ef54d3eb41d1f4eba368a726.  The only differences are:

- rpc_unix_connect() does not return CKR_DEVICE_ERROR, so make rpc_C_Initialize() happy
- use '/' as the prefix, instead of 'unix:path=', as suggested in https://lists.freedesktop.org/archives/p11-glue/2014-November/000516.html

### DIFF
--- a/common/path.c
+++ b/common/path.c
@@ -38,9 +38,11 @@
 
 #include "config.h"
 
+#include "buffer.h"
 #include "debug.h"
 #include "message.h"
 #include "path.h"
+#include "url.h"
 
 #include <assert.h>
 #include <errno.h>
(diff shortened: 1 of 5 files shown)


lkundrak:
While I agree with the mailing list post that '/' would be better than '@', but I still maintain that 'unix:path=' would be better (as mentioned in the standalone server pull request). Hopefully I'm not bikeshedding too much.

The [D-Bus transport name](https://dbus.freedesktop.org/doc/dbus-specification.html#transports) specifiers are already familiar to users and would make it easy to add new transports that may come useful in future, such as `unix:tcp` or `unix:runtime` (or even add `unixexec:` alias for `|` for no reason at all).

ueno:
Yes, having a readable prefix for transport names sounds reasonable.  However, for Unix domain sockets, IMO only "path" is relevant (in the D-Bus spec, "tmpdir" and "runtime" are shortcut for servers, and "abstract" doesn't really make sense these days; also "tcp" is not really Unix specific).  So I have changed it from '/' to "unix=/".

Also added a standalone test case.

ueno:
I ended up with "unix:path=" for consistency with DBUS_*_ADDRESS.

PR #40, opened 2017-01-23
uri: Relax pin-* parsing for compatibility

ueno:
While 'pin-source' and 'pin-value' are defined as query atttribute, they
are defined as path attribute in old drafts (-13 or earlier), and some
implementations still stick to it.

For backward compatibility, accept those in path attributes when
parsing (but not when formatting).

Reported by Andreas Metzler in:
https://lists.freedesktop.org/archives/p11-glue/2017-January/000637.html

### DIFF
--- a/p11-kit/uri.c
+++ b/p11-kit/uri.c
@@ -1390,9 +1390,9 @@ parse_module_info (const char *name_start, const char *name_end,
 }
 
 static int
-parse_extra_info (const char *name_start, const char *name_end,
-		  const char *start, const char *end,
-		  P11KitUri *uri)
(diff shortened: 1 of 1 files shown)


nmav:
Seems fine and seems to address the test suite failure. @lkundrak would that compatibility conflict with what you were trying to achieve on cbf1e42e39c030edb3e2c72ae9b4d7dd7ccf3eea? Should we set some transition period, or some flag to disable that compatibility if that could cause problems in applications like nm?

ueno:
Assuming this doesn't break compatibility with any consumers at present, I'm merging this.

PR #44, opened 2017-01-29
trust: Implement a 'trust dump' command

stefwalter:
This dumps all the PKCS#11 objects in the internal .p11-kit persistence format.
    
This is part of the trust command and tooling, even though at some point it could go in the p11-kit command. The reason for this is that the code related to the internal .p11-kit objects is in the trust code, and consumed solely by the trust related modules.


trust/list.c
```diff
@@ -253,6 +253,7 @@ p11_trust_list (int argc,
 	if (!p11_enumerate_ready (&ex, "trust-policy"))
 		exit (1);
 
+	ex.flags |= P11_ENUMERATE_CORRELATE;
```
ueno: Some tests are failing due to this behavior change:
https://travis-ci.org/p11-glue/p11-kit/jobs/196340100
Good to amend this commit with something like ueno/p11-kit@1d9bf793dd2e22853a62988dd736693beae70a77.

doc/manual/trust.xml
```diff
@@ -347,6 +350,42 @@ $ trust extract-compat
 
 </refsect1>
 
+<refsect1 id="trust-list">
```
ueno: "trust-dump"

kaie:
Thanks a lot for this patch, it's very helpful!

doc/manual/trust.xml
stefwalter: Thanks.

trust/list.c
stefwalter: Whoops. Obviously I forgot to run 'make check' :O

stefwalter:
Thanks @kaie ... I've amended this patch and force pushed.

PR #46, opened 2017-02-02
NSS attribute CKA_NSS_MOZILLA_CA_POLICY in p11-kit files

kaie:
Support loading new NSS attribute CKA_NSS_MOZILLA_CA_POLICY from .p11-kit files.
See also NSS bug https://bugzilla.mozilla.org/show_bug.cgi?id=1334976
and p11-kit bug https://bugs.freedesktop.org/show_bug.cgi?id=99453


ueno:
Merged this, as the new attribute is already added to NSS.

PR #47, opened 2017-02-13
trust: Revert to the original 'extract' behavior

ueno:
Since commit f4384a40, due to a missing ex->flags setting, the 'trust
extract' command didn't retrieve correlation between related objects and
that was causing assertion failure when writing PEM files.

https://bugs.freedesktop.org/show_bug.cgi?id=99795


kaie:
Thank you, I confirm it fixes the crash from bug 99795, and the trust extract command produces the expected output in my test scenario.

PR #51, opened 2017-02-17
trust: Honor "modifiable" setting in persist file

ueno:
Previously, all objects read from p11-kit persist files are marked as
modifiable when parsing, regardless of the explicit "modifiable: false"
setting in the file.

Reported by Kai Engert in:
https://bugs.freedesktop.org/show_bug.cgi?id=99797


kaie:
Thanks, the patch works for me.

ueno:
Thank you for confirming that.

PR #53, opened 2017-02-20
rpc: Try $XDG_CACHE_HOME before ~/.cache

grawity:
This is unset on most systems, but might as well follow the Base Directory spec properly.


ueno:
Thank you, merged.

PR #56, opened 2017-02-23
FreeBSD build fixes

novel:
A couple of FreeBSD-related build fixes for issues I stopped with 0.23.4.


ueno:
Oops, I should have run the FreeBSD build before the release.  Merged, thanks!

PR #61, opened 2017-03-01
Release 0.23.5

ueno:


novel:
It builds fine on FreeBSD for me, `make check` also passes.

There are some warnings about unused variables and functions though, but that's not important I guess.

ueno:
Thank you for testing.  Yes, I indeed see such warnings on FreeBSD, but all seem to be harmless.

PR #63, opened 2017-03-05
correct text for --user-config option

grooverdan:


ueno:
Thank you for spotting this.

grooverdan:
pretty easy when duplicate entries show up with configure  --help. Thanks for merging. Keep up the good work - tools like this help keep the CA mess closer to under control for users. Thanks for all your efforts.

PR #66, opened 2017-05-02
trust: Mark p11-kit files installed under DATA_DIR read-only

ueno:
The certificates installed by the ca-certificates package are supposed
to be only modifiable by package manager, not by administration
command.

Previously, we utilized the ```modifiable``` attribute in ```.p11-kit``` files
to enforce that, but it turned out to make it impossible to remove
custom certificates.

This patch reverts commit 8eed1e60, which introduced the behavior, and
adds a check based on the file locations.  That is, only files
installed under ```/etc``` (i.e. in "System Trust") are marked as
modifiable.

Suggested by Kai Engert in:
https://bugzilla.redhat.com/show_bug.cgi?id=1445212#c7


ueno:
I withdraw this change in favor of #70.

PR #67, opened 2017-05-11
rpc: Convert attribute value for portability

ueno:
When using the RPC across multiple architectures, where data models
are different, say LP64 vs ILP32, there can be unwanted truncation of
attribute values.

This patch converts the values into portable format for the known
attributes.

### DIFF
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -35,6 +35,7 @@
 
 #include "config.h"
 
+#define P11_DEBUG_FLAG P11_DEBUG_RPC
 #include "debug.h"
 #include "library.h"
 #include "message.h"
@@ -43,6 +44,8 @@
 
 #include <assert.h>
 #include <string.h>
+
+#define ELEMS(x) (sizeof (x) / sizeof (x[0]))
 
 void
 p11_rpc_message_init (p11_rpc_message *msg,
@@ -255,8 +258,6 @@ p11_rpc_message_write_attribute_array (p11_rpc_message *msg,
                                        CK_ULONG num)
 {
 	CK_ULONG i;
-	CK_ATTRIBUTE_PTR attr;
-	unsigned char validity;
 
 	assert (num == 0 || arr != NULL);
 	assert (msg != NULL);
@@ -268,22 +269,8 @@ p11_rpc_message_write_attribute_array (p11_rpc_message *msg,
 	/* Write the number of items */
 	p11_rpc_buffer_add_uint32 (msg->output, num);
 
-	for (i = 0; i < num; ++i) {
-		attr = &(arr[i]);
-
-		/* The attribute type */
-		p11_rpc_buffer_add_uint32 (msg->output, attr->type);
-
-		/* Write out the attribute validity */
-		validity = (((CK_LONG)attr->ulValueLen) == -1) ? 0 : 1;
-		p11_rpc_buffer_add_byte (msg->output, validity);
-
-		/* The attribute length and value */
-		if (validity) {
-			p11_rpc_buffer_add_uint32 (msg->output, attr->ulValueLen);
-			p11_rpc_buffer_add_byte_array (msg->output, attr->pValue, attr->ulValueLen);
-		}
-	}
+	for (i = 0; i < num; ++i)
+		p11_rpc_buffer_add_attribute (msg->output, &(arr[i]));
 
 	return !p11_buffer_failed (msg->output);
 }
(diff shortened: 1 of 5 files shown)


p11-kit/rpc-message.c
```diff
@@ -767,3 +755,486 @@ p11_rpc_buffer_get_byte_array (p11_buffer *buf,
...
+
+typedef size_t (*p11_rpc_attribute_value_encoder) (p11_buffer *, const void *, size_t);
+
+static size_t
+p11_rpc_buffer_add_attribute (p11_buffer *buffer, const CK_ATTRIBUTE *attr)
+{
+	unsigned char validity;
+	size_t total = 0;
+	static const p11_rpc_attribute_value_encoder encoders[] = {
+		p11_rpc_buffer_add_attribute_value_byte,
```
nmav: A suggestion to make that easier to read; there is a (C99) form that allows to make the mappings apparent:
```
static const p11_rpc_attribute_value_encoder encoders[] = {
    [P11_RPC_ATTR_BYTE] = p11_rpc_buffer_add_attribute_value_byte,
    [P11_RPC_ATTR_ULONG] = p11_rpc_buffer_add_attribute_value_ulong,
...
}
```
That form guarrantees that any unset values will be zero.

p11-kit/rpc-message.c
```diff
@@ -767,3 +755,486 @@ p11_rpc_buffer_get_byte_array (p11_buffer *buf,
...
+	if (buffer)
+		p11_rpc_buffer_add_byte (buffer, validity);
+	total++;
+
+	if (!validity)
+		return total;
+
+	/* The attribute length and value */
+	p11_rpc_buffer_add_uint32 (buffer, attr->ulValueLen);
+	encoder = encoders[map_attribute_type_to_value_type (attr->type)];
```
nmav: That seems like error prone on future revisions (when new RPC types are added), with a jump to an invalid address confusing debuggers. Maybe separate into an assign part to allow error checking? e.g.
```
if (attr->type > P11_RPC_ATTR_MAX)
  error();

mapval = map_attribute_type_to_value_type (attr->type);
encoder = encoders[mapval]
if (encoder == NULL)
  error();

```

p11-kit/rpc-message.c
```diff
@@ -767,3 +755,486 @@ p11_rpc_buffer_get_byte_array (p11_buffer *buf,
...
+		attr->type = type;
+		return true;
+	}
+
+	if (!p11_rpc_buffer_get_uint32 (buffer, offset, &length))
+		return false;
+
+	/* Decode the attribute value */
+	value_type = map_attribute_type_to_value_type (type);
+	decoder = decoders[value_type];
```
nmav: Same comment as for the encoder for safety.

p11-kit/rpc-message.c
```diff
@@ -767,3 +755,486 @@ p11_rpc_buffer_get_byte_array (p11_buffer *buf,
...
+}
+
+static size_t
+p11_rpc_buffer_add_attribute_value_byte (p11_buffer *buffer,
+					 const void *value,
+					 size_t value_length)
+{
+	CK_BYTE byte_value;
+
+	memcpy (&byte_value, value, value_length);
```
nmav: What if value_length > 1?

p11-kit/rpc-message.c
```diff
@@ -767,3 +755,486 @@ p11_rpc_buffer_get_byte_array (p11_buffer *buf,
...
+	encoder = encoders[map_attribute_type_to_value_type (attr->type)];
+	total += encoder (buffer, attr->pValue, attr->ulValueLen);
+
+	return total;
+}
+
+static bool
+p11_rpc_buffer_get_attribute_value_byte (p11_buffer *buffer,
+					 size_t *offset,
+					 uint32_t expected,
```
nmav: Are both expected and value_length needed?

p11-kit/rpc-server.c
```diff
@@ -306,34 +302,31 @@ proto_read_attribute_array (p11_rpc_message *msg,
...
-			attrs[i].ulValueLen = value;
 		} else {
 			attrs[i].pValue = NULL;
 			attrs[i].ulValueLen = -1;
 		}
+
+		msg->parsed = offset;
 	}
 
 	*result = attrs;
```
nmav: I think a test suite of the new changes would be quite needed. I attach a minimal one for the byte and ulong types. 
[test-suite.patch.gz](https://github.com/p11-glue/p11-kit/files/996119/test-suite.patch.gz)


Something related, is that since we go through systems we would also need a fuzzer for the server and client to ensure there is no compromise possible; we can use AFL or even oss-fuzz project for that. An example fuzzer that can be used as base is at gnutls repo devel/fuzz/gnutls_server_fuzzer.cc.

ueno: Thank you for the review.  It's embarrassing that the patch didn't have any additional tests.  I have merged the patch and added tests for other value types.  I will give it a try with AFL.

p11-kit/rpc-message.c
ueno: Actually, the expected length is tracked in buffer, so I removed expected.

p11-kit/rpc-message.c
ueno: I added a couple of asserts there.

p11-kit/rpc-message.c
```diff
@@ -767,3 +755,486 @@ p11_rpc_buffer_get_byte_array (p11_buffer *buf,
...
+
+static size_t
+p11_rpc_buffer_add_attribute_value_ulong (p11_buffer *buffer,
+					  const void *value,
+					  size_t value_length)
+{
+	CK_ULONG ulong_value;
+
+	memcpy (&ulong_value, value, value_length);
+	if (ulong_value > 0xFFFFFFFFFFFFFFFF)
```
ueno: Here the code converts the value to the wire format in two steps: first it converts ```void *``` into ```CK_ULONG``` and then ```CK_ULONG``` to ```uint64_t```.  So there should be separate checks for those.  I have added them to all encoding functions.

ueno:
Merging this.  The fuzzer based on AFL was added as #73 .

PR #81, opened 2017-06-07
Add Solaris support

grobian:
These commits add support for compilation on Solaris-like platforms.  We need some special libs for socket funcs, a slightly different invocation of getpwnam_r, we can use ucred for p11_get_upeer_id, and we need to work around a linkage problem with dlclose.

### DIFF
--- a/common/unix-peer.c
+++ b/common/unix-peer.c
@@ -43,6 +43,10 @@
 #include <sys/uio.h>
 #include <sys/errno.h>
 
+#ifdef HAVE_UCRED_H
+#  include <ucred.h>
+#endif
+
 /* Returns the unix domain socket peer information.
  * Returns zero on success.
  */
@@ -87,6 +91,21 @@ p11_get_upeer_id (int cfd, uid_t *uid, uid_t *gid, pid_t *pid)
 	if (pid)
 		*pid = -1;
 
+#elif defined(HAVE_GETPEERUCRED)
+	/* *Solaris/OpenIndiana */
+	ucred_t *ucred = NULL;
+
+	if (getpeerucred(cfd, &ucred) == -1)
+		return -1;
+
+	ret = ( (uid && (*uid = ucred_geteuid(ucred)) == -1) ||
+			(gid && (*gid = ucred_getrgid(ucred)) == -1) ||
+			(pid && (*pid = ucred_getpid(ucred)) == -1)  );
+
+	ucred_free(ucred);
+
+	if (ret)
+		return -1;
 #else
 #error "Unsupported UNIX variant"
 #endif
--- a/configure.ac
+++ b/configure.ac
@@ -96,7 +96,7 @@ if test "$os_unix" = "yes"; then
 	])
 
 	# These are thngs we can work around
-	AC_CHECK_HEADERS([locale.h sys/resource.h])
+	AC_CHECK_HEADERS([locale.h sys/resource.h ucred.h])
 	AC_CHECK_MEMBERS([struct dirent.d_type],,,[#include <dirent.h>])
 	AC_CHECK_FUNCS([getprogname getexecname basename mkstemp mkdtemp])
 	AC_CHECK_FUNCS([getauxval issetugid getresuid secure_getenv])
@@ -104,6 +104,7 @@ if test "$os_unix" = "yes"; then
 	AC_CHECK_FUNCS([fdwalk])
 	AC_CHECK_FUNCS([setenv])
 	AC_CHECK_FUNCS([getpeereid])
+	AC_CHECK_FUNCS([getpeerucred])
 
 	# Required functions
 	AC_CHECK_FUNCS([gmtime_r],
--- a/configure.ac
+++ b/configure.ac
@@ -88,6 +88,13 @@ if test "$os_unix" = "yes"; then
 		AC_MSG_ERROR([could not find dlopen])
 	])
 
+	# for Solaris we need -lsocket -lnsl for socket stuff, gethostbyname
+	# is just a dummy to find -lnsl
+	AC_SEARCH_LIBS([gethostbyname], [nsl])
+	AC_SEARCH_LIBS([connect], [socket], [], [
+		AC_MSG_ERROR([could not find socket])
+	])
+
 	# These are thngs we can work around
 	AC_CHECK_HEADERS([locale.h sys/resource.h])
 	AC_CHECK_MEMBERS([struct dirent.d_type],,,[#include <dirent.h>])
--- a/common/compat.c
+++ b/common/compat.c
@@ -278,12 +278,6 @@ p11_dl_error (void)
 	return msg_buf;
 }
 
-void
-p11_dl_close (void *dl)
-{
-	FreeLibrary (dl);
-}
-
 int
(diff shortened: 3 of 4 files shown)


ueno:
Thank you.

> On Linux there is no 5-arg getpwnam_r, which is fine.

I don't think this is true; without this change the configure test actually fails with:
```
conftest.c:48:12: error: 'NULL' undeclared (first use in this function)
 getpwnam_r(NULL,NULL,NULL,0,NULL);
            ^~~~
```
so you could simply add ```#include <stdlib.h>``` or pass 0.

By the way, where is ```getpwnam_r``` is used?  I couldn't find any in your branch.

grobian:
Hmmm, that's odd.  My branch went back a long while, and I can't remember why I had to add that.  This looks very stupid now.  Especially since it took me a while to get it working.

I tried compiling without, and got this:

libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I. -I./common "-DBINDIR=\"/usr/local/bin\"" "-DBUILDDIR=\"/net/ptah/export/gentoo/working-repos/p11-kit\"" "-DDATA_DIR=\"/usr/local/share\"" "-DPRIVATEDIR=\"/usr/local/libexec/p11-kit\"" "-DSRCDIR=\"/net/ptah/export/gentoo/working-repos/p11-kit\"" "-DSYSCONFDIR=\"/usr/local/etc\"" -DP11_KIT_FUTURE_UNSTABLE_API -g -O2 -g -Wall -Wstrict-prototypes -Wmissing-declarations -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wdeclaration-after-statement -Wformat=2 -Winit-self -Waggregate-return -Wno-missing-format-attribute -fno-strict-aliasing -fno-common -Wmissing-include-dirs -Wundef -MT common/url.lo -MD -MP -MF common/.deps/url.Tpo -c common/url.c  -fPIC -DPIC -o common/.libs/url.o
common/path.c: In function ‘expand_homedir’:
common/path.c:127:4: warning: missing sentinel in function call [-Wformat=]
    return p11_path_build (env, remainder + 8, NULL);
    ^
common/path.c:132:3: warning: missing sentinel in function call [-Wformat=]
   return p11_path_build (env, remainder, NULL);
   ^
common/path.c:143:9: error: too many arguments to function ‘getpwuid_r’
   ret = getpwuid_r (getuid (), &pws, buf, sizeof (buf), &pwd);
         ^
In file included from common/path.c:54:0:
/usr/include/pwd.h:166:23: note: declared here
 extern struct passwd *getpwuid_r(uid_t, struct passwd *, char *, int);
                       ^
common/path.c:143:7: warning: assignment makes integer from pointer without a cast [-Wint-conversion]
   ret = getpwuid_r (getuid (), &pws, buf, sizeof (buf), &pwd);
       ^
common/path.c:154:3: warning: missing sentinel in function call [-Wformat=]
   return p11_path_build (pwd->pw_dir, remainder, NULL);
   ^
make[2]: *** [Makefile:2874: common/path.lo] Error 1
make[2]: *** Waiting for unfinished jobs....

So, it seems I managed to mix up some things.
a) it's getpwuid_r
b) on Solaris by default it wants 4, unless POSIX_PTHREAD_SEMANTICS is set, then it wants 5 like Linux (and Darwin)

grobian:
I think I would like to merge the three commits in one, but I'll have to do some git-research if that's possible at all without creating a new PR.

grobian:
Travis failed due to timeouts on their side, and I can't restart the jobs.  Can you please restart them so the code is actually tested?

common/unix-peer.c
```diff
@@ -92,6 +96,21 @@ p11_get_upeer_id (int cfd, uid_t *uid, uid_t *gid, pid_t *pid)
...
 
+#elif defined(HAVE_GETPEERUCRED)
+	/* *Solaris/OpenIndiana */
+	ucred_t *ucred = NULL;
+	(void)ret;  /* pacify compiler */
+
+	if (getpeerucred(cfd, &ucred) == -1)
+		return -1;
+	if (uid && (*uid = ucred_geteuid(ucred)) == -1)
+		return -1;
```
ueno: I see a minor leak of ``ucred``` here and in the following if-clauses.

configure.ac
```diff
@@ -100,6 +100,19 @@ if test "$os_unix" = "yes"; then
...
 
+	AC_MSG_CHECKING([whether getpwuid_r() accepts 5 arguments])
+	AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <pwd.h>]],
+	                                   [[getpwuid_r(0,0,0,0,0);]])],
+					  [],
+					  [AC_COMPILE_IFELSE(
+					   [AC_LANG_PROGRAM([[#define _POSIX_PTHREAD_SEMANTICS
+#include <pwd.h>]],
+									    [[getpwuid_r(0,0,0,0,0);]])],
+					   [AC_DEFINE([_POSIX_PTHREAD_SEMANTICS], [1],
```
ueno: Maybe we could do this unconditionally without checking? For extra safety, it might be good to emit an ```#ifndef ... #undef ... #endif``` guard with [some AH_* macros](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Header-Templates.html#Header-Templates).

grobian: Reason I did this, was to ensure the define is added only when it "helps".  I will try using AC_USE_SYSTEM_EXTENSIONS (2.60+) which should define this one, and more.  Hopefully that won't conflict.  Then the entire check can go, I presume.
What do you want to guard in the end?  The extra argument, or the entire call?

common/unix-peer.c
grobian: "When  successful,  getpeerucred() stores the pointer to a freshly allo‐
cated ucred_t in the memory location pointed to by the  ucred  argument
if  that memory location contains the null pointer. If the memory loca‐
tion is non-null, it will reuse the existing ucred_t."

So, I will allocate on the stack instead so we don't have to worry about this.

configure.ac
grobian: ok, I guess the answer is no :)
```
./common/compat.h:44:2: error: #error Make the crap stop. _GNU_SOURCE is completely unportable and breaks all sorts of behavior
 #error Make the crap stop. _GNU_SOURCE is completely unportable and breaks all sorts of behavior
  ^
In file included from ./common/library.h:41:0,
                 from p11-kit/proxy-init.c:40:
```

grobian:
Is this approach more favourable?

configure.ac
ueno: That's unfortunate.  Looking at the git log, it was apparently disabled for the strerror_r compatibility with glibc.  I have opened a pull request #82 to re-enable it.

grobian:
When #82 is merged, the first commit can go, I just kept it in order to keep the Solaris part compiling with the current head.

ueno:
I have merged #82.  Could you rebase the patches against the git master?

grobian:
Anything I can do here?  Do you have any comments/remarks/suggestions?

ueno:
Sorry, I thought that there was a conflict - it was my misunderstanding.  Merged the patches, thanks!

PR #84, opened 2017-06-23
Be silent by default and do not print messages on stderr

nmav:
As p11-kit is a library there are cases where it is not desirable
to log on stderr by default. See for example this report
https://bugzilla.redhat.com/show_bug.cgi?id=1464490
where wget prints an error due to an unconfigured pkcs11 module.

Signed-off-by: Nikos Mavrogiannopoulos <nmav@redhat.com>

### DIFF
--- a/common/message.c
+++ b/common/message.c
@@ -58,7 +58,7 @@
 #include <stdio.h>
 #include <string.h>
 
-static bool print_messages = true;
+static bool print_messages = false;
 
 static char *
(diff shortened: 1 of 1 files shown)


nmav:
This is currently untested.

Safari77:
Thanks. It prevents stderr logs, but I still get the two systemd journal log lines.

ueno:
Thanks, merged.

PR #90, opened 2017-07-31
pkcs11: added definitions of GOST mechanisms

nmav:
This follows the definitions in PKCS#11 v2.40:
  http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/os/pkcs11-curr-v2.40-os.html


nmav:
The reason for this addition is that I now use pkcs11.h from p11-kit to auto-generate the textual form of a token's mechanism and I found out that these mechanisms were missing from it.

nmav:
Makes sense. Done in updated patch.

PR #91, opened 2017-08-01
Add OTP-related definitions

nmav:
It seems that we don't include any PKCS#11 mechanisms related to OTP. I simply included them following PKCS#11 2.40 however, I have no use case of it at the moment.

### DIFF
--- a/common/pkcs11.h
+++ b/common/pkcs11.h
@@ -12,12 +12,12 @@
    the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
    PURPOSE.  */
 
-/* Please submit changes back to the Scute project at
-   http://www.scute.org/ (or send them to marcus@g10code.com), so that
+/* Please submit any changes back to the p11-kit project at
+   https://github.com/p11-glue/p11-kit/, so that
    they can be picked up by other projects from there as well.  */
 
 /* This file is a modified implementation of the PKCS #11 standard by
-   RSA Security Inc.  It is mostly a drop-in replacement, with the
+   OASIS group.  It is mostly a drop-in replacement, with the
    following change:
 
    This header file does not require any macro definitions by the user
--- a/common/pkcs11.h
+++ b/common/pkcs11.h
@@ -1,6 +1,7 @@
 /* pkcs11.h
    Copyright 2006, 2007 g10 Code GmbH
    Copyright 2006 Andreas Jellinghaus
+   Copyright 2017 Red Hat, Inc.
 
    This file is free software; as a special exception the author gives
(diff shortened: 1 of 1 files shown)


common/pkcs11.h
```diff
@@ -12,12 +12,12 @@
...
 
-/* Please submit changes back to the Scute project at
-   http://www.scute.org/ (or send them to marcus@g10code.com), so that
+/* Please submit any changes back to the p11-kit project at
+   https://github.com/p11-glue/p11-kit/, so that
    they can be picked up by other projects from there as well.  */
 
 /* This file is a modified implementation of the PKCS #11 standard by
-   RSA Security Inc.  It is mostly a drop-in replacement, with the
+   OASIS group Security Inc.  It is mostly a drop-in replacement, with the
```
ueno: Please remove "Security Inc" here.

common/pkcs11.h
```diff
@@ -12,12 +12,12 @@
    the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
    PURPOSE.  */
 
-/* Please submit changes back to the Scute project at
-   http://www.scute.org/ (or send them to marcus@g10code.com), so that
+/* Please submit any changes back to the p11-kit project at
+   https://github.com/p11-glue/p11-kit/, so that
```
ueno: I tend to agree with this change, though Scute seems to have gained a new maintainer.

common/pkcs11.h
```diff
@@ -754,6 +809,25 @@ struct ck_mechanism_info
   ck_flags_t flags;
 };
 
+typedef unsigned long ck_param_type;
+
+typedef struct ck_otp_param {
+   ck_param_type type;
+   void *pValue;
+   unsigned long ulValueLen;
```
ueno: For consistency with the GNU convention, I would suggest to rename members in snake case without type prefix: ulValueLen -> value_len.

common/pkcs11.h
nmav: Done.

common/pkcs11.h
nmav: Done. I've also renamed the rest (ulCount and pParams)

common/pkcs11.h
nmav: It's up to you. I think that scute has more limited scope of pkcs11 (gnupg cards) thus they may have no reason to support new mechanisms, though they could benefit from the pkcs11 file here also. The fork of that file goes way back to past and I'm not sure that this file is being updated since then:
https://github.com/gpg/scute/blob/master/src/pkcs11.h

PR #99, opened 2017-08-22
trust: Respect anyExtendedKeyUsage in CA certificates

ueno:
A CA certificate can include anyExtendedKeyUsage (2.5.29.37.0) in the extended key usage extension, if the CA doesn't want to restrict usages of the key.   However, previously the trust module treated it as a custom key purpose and failed to match if the search criteria included a key purpose.


ueno:
Updated commit reflecting comments from @kaie:
- remove unnecessary critical check
- add comment on the encoded value in the test

Issue #95, opened 2017-08-16
0.23.8 build failure on macOS


ilovezfs:
```
libtool: compile:  clang -DHAVE_CONFIG_H -I. -I. -I./common -DBINDIR=\"/usr/local/Cellar/p11-kit/0.23.8/bin\" -DBUILDDIR=\"/private/tmp/p11-kit-20170815-64820-14apsw1/p11-kit-0.23.8\" -DDATA_DIR=\"/usr/local/Cellar/p11-kit/0.23.8/share\" -DPRIVATEDIR=\"/usr/local/Cellar/p11-kit/0.23.8/libexec/p11-kit\" -DSRCDIR=\"/private/tmp/p11-kit-20170815-64820-14apsw1/p11-kit-0.23.8\" -DSYSCONFDIR=\"/usr/local/etc\" -DP11_KIT_FUTURE_UNSTABLE_API -g -O2 -g -Wall -Wstrict-prototypes -Wmissing-declarations -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wdeclaration-after-statement -Wformat=2 -Winit-self -Waggregate-return -Wno-missing-format-attribute -fno-strict-aliasing -fno-common -Wmissing-include-dirs -Wundef -c common/debug.c  -fno-common -DPIC -o common/.libs/debug.o
common/compat.c:498:6: error: use of undeclared identifier 'SIZE_MAX'
        if (SIZE_MAX / nmemb < size) {
            ^
1 error generated.
make[2]: *** [common/compat.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
[...]

PR #100, opened 2017-08-26
Include stdint.h for SIZE_MAX

novel:
On some systems (e.g. FreeBSD 10.x) build fails with:

common/compat.c:498:6: error: use of undeclared identifier 'SIZE_MAX'
        if (SIZE_MAX / nmemb < size) {

Fix by including stdint.h.

### DIFF
--- a/common/compat.c
+++ b/common/compat.c
@@ -52,6 +52,7 @@
 #include <errno.h>
 #include <stdint.h>
 #include <stdlib.h>
+#include <stdint.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


ueno:
Thank you, but it has already been fixed as #95. Perhaps we should make a new bug-fix release.

novel:
Oh, I've somehow overlooked it. Yeah, new release would be nice, because 0.23.8 also fails to build without debug (fixed in master).

PR #110, opened 2017-10-30
p11_kit_override_system_files: introduced new function

nmav:
That allows overriding the default module and configuration
locations, for use in test suites, etc.

Signed-off-by: Nikos Mavrogiannopoulos <nmav@gnutls.org>

### DIFF
--- a/p11-kit/modules.c
+++ b/p11-kit/modules.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (C) 2008 Stefan Walter
  * Copyright (C) 2011 Collabora Ltd.
+ * Copyright (C) 2017 Red Hat, Inc.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -184,6 +185,51 @@ const char *p11_config_user_modules = P11_USER_CONFIG_MODULES;
 /* -----------------------------------------------------------------------------
  * P11-KIT FUNCTIONALITY
  */
+
+/**
+ * p11_kit_override_system_files:
+ * @system_conf: the system configuration file (default: system_config_dir/pkcs11.conf)
+ * @user_conf: the user configuration file (default: ~/.config/pkcs11/pkcs11.conf)
+ * @package_modules: location of modules shipped by p11-kit (default: system_config/modules)
+ * @system_modules: location of system pkcs11 modules (default: system_config/modules)
+ * @user_modules: location of user modules (default: ~/.config/pkcs11/modules)
+ *
+ * Overrides the default system configuration files. The
+ * provided values should be accessible for the lifetime
+ * of p11-kit usage.
+ *
+ * When the value %NULL is provided for any of the locations,
+ * it will not be updated.
+ *
+ * This is function intended to be used in test suites and
+ * not production, and as such %P11_KIT_FUTURE_UNSTABLE_API
+ * must be defined before including p11-kit.h.
+ *
+ * Since: 0.23.10
+ *
+ */
+void
+p11_kit_override_system_files (const char *system_conf,
+                               const char *user_conf,
+                               const char *package_modules,
+                               const char *system_modules,
+                               const char *user_modules)
+{
+	if (system_conf)
+		p11_config_system_file = system_conf;
+
+	if (user_conf)
+		p11_config_user_file = user_conf;
+
+	if (package_modules)
+		p11_config_package_modules = package_modules;
+
+	if (system_modules)
+		p11_config_system_modules = system_modules;
+
+	if (user_modules)
+		p11_config_user_modules = user_modules;
+}
 
 static CK_RV
 create_mutex (CK_VOID_PTR_PTR mut)
--- a/p11-kit/p11-kit.h
+++ b/p11-kit/p11-kit.h
@@ -111,6 +111,11 @@ void                   p11_kit_be_loud                      (void);
 
 void                   p11_kit_set_progname                 (const char *progname);
 
+void                   p11_kit_override_system_files        (const char *system_conf,
+                                                             const char *user_conf,
+                                                             const char *package_modules,
+                                                             const char *system_modules,
+						             const char *user_modules);
 #endif
 
 const char *           p11_kit_message                      (void);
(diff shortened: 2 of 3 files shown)


p11-kit/modules.c
```diff
@@ -185,6 +185,50 @@ const char *p11_config_user_modules = P11_USER_CONFIG_MODULES;
...
+ * @user_conf: the user configuration file (default: ~/.config/pkcs11/pkcs11.conf)
+ * @package_modules: location of modules shipped by p11-kit (default: system_config/modules)
+ * @system_modules: location of system pkcs11 modules (default: system_config/modules)
+ * @user_modules: location of user modules (default: ~/.config/pkcs11/modules)
+ *
+ * Overrides the default system configuration files. The
+ * provided values should be accessible for the lifetime
+ * of p11-kit usage.
+ *
+ * When the value %P11_KIT_IGNORE_PTR is provided for any of the locations,
```
ueno: Can't we simply use NULL, given all the variables are preset to non-NULL strings?

p11-kit/p11-kit.h
```diff
@@ -111,6 +111,12 @@ void                   p11_kit_be_loud                      (void);
 
 void                   p11_kit_set_progname                 (const char *progname);
 
+# define P11_KIT_IGNORE_PTR ((void*)-1)
+void                   p11_kit_override_system_files        (const char *system_conf,
+                                                             const char *user_conf,
+                                                             const char *package_modules,
+                                                             const char *system_modules,
+						             const char *user_modules);
```
ueno: Here is an indentation mismatch.

p11-kit/test-transport.c
```diff
@@ -87,8 +88,9 @@ setup_remote (void *unused)
 	data = "remote: |" BUILDDIR "/p11-kit/p11-kit" EXEEXT " remote " BUILDDIR "/.libs/mock-five" SHLEXT "\nx-init-reserved: initialize-arg";
 	p11_test_file_write (test.user_modules, "init-arg.module", data, strlen (data));
 
-	p11_config_user_modules = test.user_modules;
-	p11_config_user_file = test.user_config;
+	p11_kit_override_system_files(P11_KIT_IGNORE_PTR, test.user_config,
```
ueno: Add a space before "(".

p11-kit/modules.c
```diff
@@ -185,6 +185,50 @@ const char *p11_config_user_modules = P11_USER_CONFIG_MODULES;
...
+ * it will not be updated.
+ *
+ * This is function intended to be used in test suites and
+ * not production, and as such %P11_KIT_FUTURE_UNSTABLE_API
+ * must be defined before including p11-kit.h.
+ *
+ * Since: 0.23.10
+ *
+ */
+void p11_kit_override_system_files (const char *system_conf,
```
ueno: Add a newline after "void".

p11-kit/modules.c
nmav: What if we want to set some files as being null? (e.g., set the system file and make the user file null)

p11-kit/test-transport.c
nmav: done

p11-kit/p11-kit.h
nmav: fixed

p11-kit/modules.c
nmav: fixed

p11-kit/modules.c
ueno: As far as I can see, the code assumes all those variables non-null.  For example:
```p11_config_system_file``` is passed to ```_p11_conf_load_globals```:
https://github.com/p11-glue/p11-kit/blob/master/p11-kit/modules.c#L571
```_p11_conf_load_globals``` calls ```_p11_conf_parse_file```:
https://github.com/p11-glue/p11-kit/blob/master/p11-kit/conf.c#L201
which has a non-NULL assertion on the argument:
https://github.com/p11-glue/p11-kit/blob/master/p11-kit/conf.c#L95

Setting the other 4 variables to NULL also leads to assertion failure.  Maybe you could add some assertions in ```p11_kit_override_system_files``` to fail early?

Comment (nmav) [reply p11-kit/modules.c]: Done. I've also added some text on why P11_KIT_IGNORE_PTR is there instead of NULL.

Comment (ueno) [reply p11-kit/modules.c]: Sorry, that's not what I meant; disregard the last paragraph in my previous comment.  Given that the library doesn't accept NULL as the value of those 5 variables, we could use NULL as an indicator of using a default value, instead of introducing P11_KIT_IGNORE_PTR.

File: p11-kit/test-transport.c [hunk 5]
```diff
@@ -87,8 +88,9 @@ setup_remote (void *unused)
 	data = "remote: |" BUILDDIR "/p11-kit/p11-kit" EXEEXT " remote " BUILDDIR "/.libs/mock-five" SHLEXT "\nx-init-reserved: initialize-arg";
 	p11_test_file_write (test.user_modules, "init-arg.module", data, strlen (data));
 
-	p11_config_user_modules = test.user_modules;
-	p11_config_user_file = test.user_config;
+	p11_kit_override_system_files( P11_KIT_IGNORE_PTR, test.user_config,
```
ueno: I meant to add a space before "(", not after it.

p11-kit/modules.c
nmav: I was thinking of future extensibility where null could be accepted, but you are right, no reason to overthink of it. Let's use null. Updated.

p11-kit/test-transport.c
nmav: Done

p11-kit/modules.c
```diff
@@ -185,6 +186,51 @@ const char *p11_config_user_modules = P11_USER_CONFIG_MODULES;
...
+ *
+ */
+void
+p11_kit_override_system_files (const char *system_conf,
+                               const char *user_conf,
+                               const char *package_modules,
+                               const char *system_modules,
+                               const char *user_modules)
+{
+	assert (system_conf);
```
ueno: This is causing a test failure in travis:
https://travis-ci.org/p11-glue/p11-kit/jobs/295657233#L4820
I guess this (and the following assignments) should be something like:
```
if (system_conf)
        p11_config_system_file = system_conf;
```

nmav: fixed. sorry for that.

PR #120, opened 2018-01-16
proxy: Reuse the existing slot ID mapping after fork

ueno:
While the proxy module reassigns slot IDs in ```C_Initialize()```, some applications assume that valid slot IDs should never change across multiple calls to ```C_Initialize()```.  This patch mitigates this by preserving the slot IDs, if they are known to the proxy module.

Note that this change is experimental.  @ansasaki, @nmav this is what I was talking yesterday.


ueno:
I talked with @rjrelyea about this and it seems that the remapping is really a responsibility of a module.

PR #124, opened 2018-01-22
builder: Print warning if cert loaded from anchor location is not CA

ueno:
Reported in:
https://bugzilla.redhat.com/show_bug.cgi?id=1492700

### DIFF
--- a/trust/builder.c
+++ b/trust/builder.c
@@ -562,6 +562,7 @@ calc_certificate_category (p11_builder *builder,
                            CK_ULONG *category)
 {
 	CK_ATTRIBUTE *label;
+	CK_ATTRIBUTE *origin;
 	unsigned char *ext;
 	size_t ext_len;
 	bool is_ca = 0;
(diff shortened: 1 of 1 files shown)


ueno:
Thank you for the review; I guess I need to write a test to land this.

ZoltanFridrich:
We have decided to close this MR.

The rationale is: while anchors are typically a CA certificate, there may be a use-case of marking end entity certificates trusted or distrusted, as it was possible with the trust assertions (the predecessor of trust policies):
https://p11-glue.github.io/p11-glue/doc/pkcs11-trust-assertions/

The design document of trust policies also has this paragraph:
"At one end of a certificate chain is the end entity certificate, which is the certificate that is being validated. At the other end the certificate chain is anchored by a trust anchor. This is a public key that is explicitly trusted by the local system, either by default or by a deliberate configuration choice. Usually this public key is represented as a certificate. **The anchor is usually, but not always, a root self-signed certificate authority.**"

PR #128, opened 2018-01-31
po: Update translations from transifex

### LINKED ISSUE #62: Pulling translations from Transifex
It’s been more than a year without pulling updated translations from Transifex. It would be great if it could be made a part of the release process.

ueno:
Fixes #62.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -1,6 +1,7 @@
 # Set of available languages.
 ar
 as
+ast
 az
 bg
 bn_IN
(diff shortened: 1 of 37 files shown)


po/LINGUAS
```diff
@@ -69,3 +69,5 @@ zh_HK
 zh_TW
 oc
 et
+ast
+fur
```
piotrdrag: Nitpick, but maybe we could have it sorted alphabetically?

ueno: Right, fixed.

PR #134, opened 2018-03-17
Treat CKR_CRYPTOKI_ALREADY_INITIALIZED correctly

phoenix-frozen:
In p11_kit_modules_initialize(), treat a return code
of CKR_CRYPTOKI_ALREADY_INITIALIZED as identical to
CKR_OK.

### DIFF
--- a/p11-kit/modules.c
+++ b/p11-kit/modules.c
@@ -2115,7 +2115,7 @@ p11_kit_modules_initialize (CK_FUNCTION_LIST **modules,
 
 	for (i = 0, out = 0; modules[i] != NULL; i++, out++) {
 		rv = modules[i]->C_Initialize (NULL);
-		if (rv != CKR_OK) {
+		if (rv != CKR_OK && rv != CKR_CRYPTOKI_ALREADY_INITIALIZED) {
 			name = p11_kit_module_get_name (modules[i]);
 			if (name == NULL)
 				name = strdup ("(unknown)");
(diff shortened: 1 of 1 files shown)


phoenix-frozen:
This patch fixes issue 133

phoenix-frozen:
Good point, I'll try that and push an update shortly.

phoenix-frozen:
Actually. This turns out to be *hard* -- the bug really only shows up where p11-kit gets dlclose()d and then dlopen()d again -- as happens in wpa_supplicant when it throws away its TLS context and starts again -- but its the modules it's managing don't. Writing a test that does this turns out to be quite the challenge.

phoenix-frozen:
(And in particular, none of the existing test harness seems to support it.)

nmav:
What is the actual scenario. Is it (1) p11-kit is dlopened() and default modules are initialized, (2) p11-kit is dlclosed, (3) p11-kit is dlopened again and same modules are initialized?

phoenix-frozen:
Yes, I'm pretty sure that's what's going on. (I'm not saying "yes, absolutely" because actually understanding wpa_supplicant's behavior is also hard.)

ueno:
Though I don't understand wpa_supplicant either, I guess it wouldn't be hard to write a test for the code change: create a mock module that returns `CKR_CRYPTOKI_ALREADY_INITIALIZED` and call `p11_kit_modules_initialize` twice.

I added an expected-to-fail test in #135, namely:
https://github.com/p11-glue/p11-kit/pull/135/commits/c5bc9a53a58cf029cf1c4560146d0438561a3dc0#diff-c5be56620bf93a272f063fb58a64bc58R478

phoenix-frozen:
Yeah, that should do it.

ueno:
Thank you for the patch.  As for the actual issue with wpa_supplicant, @lkundrak perhaps might know what's happening there?

phoenix-frozen:
Woohoo! Thanks!

Issue #101, opened 2017-09-01
Failed build on IRIX


gwfisher:
I'm trying to build P11-kit on IRIX so I can build gnutls. However, it's giving me this error when I compile it. I've tried it with both MIPSPro (SGI's proprietary compiler) and GCC with no such luck. As I understand this also happens on AIX.

Is there anyway I can get this to work?

Any help is appreciated!

bash-4.3$ gmake
gmake  all-recursive
gmake[1]: Entering directory `/usr/people/wfisher/p11-kit-0.23.3'
Making all in .
gmake[2]: Entering directory `/usr/people/wfisher/p11-kit-0.23.3'
  CC       common/compat.lo
common/compat.c: In function 'getprogname':
common/compat.c:118:3: error: #error No way to retrieve short program name
common/compat.c: In function 'strerror_r':
common/compat.c:854:3: error: #error no strerror_r implementation
common/compat.c: In function 'fdwalk':
common/compat.c:894:8: warning: implicit declaration of function 'dirfd' [-Wimplicit-function-declaration]
common/compat.c:894:8: warning: nested extern declaration of 'dirfd' [-Wnested-externs]
common/compat.c: In function 'strerror_r':
common/compat.c:856:1: warning: control reaches end of non-void function [-Wreturn-type]
gmake[2]: *** [common/compat.lo] Error 1
gmake[2]: Leaving directory `/usr/people/wfisher/p11-kit-0.23.3'
gmake[1]: *** [all-recursive] Error 1
gmake[1]: Leaving directory `/usr/people/wfisher/p11-kit-0.23.3'
gmake: *** [all] Error 2

nmav:
If you don't use p11-kit functionality in gnutls, then simply disable it. The system seems to miss important posix functionality such as strerror_r(). Closing it, but please feel free to open a merge request to address such portability issues.

PR #137, opened 2018-03-28
Extract edk2/cacerts.bin (RHBZ#1559580)

lersek:
Please refer to https://bugzilla.redhat.com/show_bug.cgi?id=1559580 for the feature description.

I have successfully tested this against firmware code I've written (and in part pushed) for https://bugzilla.redhat.com/show_bug.cgi?id=1536624 . Testing instructions are attached to that BZ: https://bugzilla.redhat.com/attachment.cgi?id=1413960

The intended invocation by `update-ca-trust` is spelled out in the commit message of the second patch.

Thanks!
Laszlo

### DIFF
--- a/trust/extract-edk2.c
+++ b/trust/extract-edk2.c
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2018, Red Hat Inc.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the
+ *       following disclaimer.
+ *     * Redistributions in binary form must reproduce the
+ *       above copyright notice, this list of conditions and
+ *       the following disclaimer in the documentation and/or
+ *       other materials provided with the distribution.
+ *     * The names of contributors to this software may not be
+ *       used to endorse or promote products derived from this
+ *       software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * Author: Laszlo Ersek <lersek@redhat.com>
+ */
+
+#include "config.h"
+
+#include "extract.h" /* p11_extract_edk2_cacerts() */
+
+bool
+p11_extract_edk2_cacerts (p11_enumerate *ex,
+                          const char *destination)
+{
+	return false;
+}
--- a/trust/extract-edk2.c
+++ b/trust/extract-edk2.c
@@ -34,11 +34,167 @@
 
 #include "config.h"
 
+#include "buffer.h"  /* p11_buffer */
+#include "debug.h"   /* return_val_if_fail() */
+#include "message.h" /* p11_message() */
 #include "extract.h" /* p11_extract_edk2_cacerts() */
+
+#include <stdint.h>  /* UINT32_MAX */
+#include <limits.h>  /* SSIZE_MAX */
+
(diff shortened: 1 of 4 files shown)


gnutlsmirror:
Thank you! I'll make some comments inline. Would you like to add a test which extracts a fixed set of certs and compares the result with a set file? That would prevent future regressions in that code.

trust/extract-edk2.c
```diff
@@ -34,11 +34,167 @@
...
+#include "buffer.h"  /* p11_buffer */
+#include "debug.h"   /* return_val_if_fail() */
+#include "message.h" /* p11_message() */
 #include "extract.h" /* p11_extract_edk2_cacerts() */
 
+#include <stdint.h>  /* UINT32_MAX */
+#include <limits.h>  /* SSIZE_MAX */
+
+/* types from the UEFI 2.7 spec, section "31.4.1 Signature Database" */
+#pragma pack (1)
```
gnutlsmirror: I that pragma supported by gcc or clang? Isn't `__attribute__((packed))` a better alternative for the compilers we use?

trust/extract-edk2.c
```diff
@@ -34,11 +34,167 @@
...
+	0x4aa7,
+	{ 0x87, 0xb5, 0xab, 0x15, 0x5c, 0x2b, 0xf0, 0x72 }
+};
+
+/*
+ * the GUID identifying this extractor as "agent"
+ * (DCDD3B50-F405-43FD-96BE-BD33B1734776, generated with "uuidgen"), in host
+ * byte order
+ */
+static const efi_guid agent_guid_host = {
```
gnutlsmirror: Where is that GUID being use? Does it need to be unique per extraction?

lersek:
I'm not really used to github pull requests (I work primarily on mailing lists), so please forgive if I don't get the format right etc... I'm unsure how I can respond to comments directly, so I'll do some manual quoting here.

(1) Who is "gnutlsmirror"? May I know the person behind the persona? :)

(2) *Would you like to add a test which extracts a fixed set of certs and compares the result with a set file?*

Looking at `trust/test-extract.in`, I think I could extend the same logic to this extractor. That is, first check that all three certificates are present in the binary output file (`grep` works on binary files as well), and then check that the blacklisted cert is missing from the binary output file. I think `assert_contains` and `assert_not_contains` can be reused as they are.

Would this work for you?

(3) Regarding `#pragma pack`:

Before I added that, I grepped the tree for `__attribute__((packed))`, `#pragma pack`, or any home-grown macros that might wrap these. To my surprise, while the tree lacks any use of the former, it does use `#pragma pack` already (see `common/pkcs11.h`).

So, I'm fine either way, but `#pragma pack` seems to fit both edk2 and p11-kit better.

Also, gcc has supported `#pragma pack` essentially forever; in RHEL-5 (that's not a typo) we ship `gcc-4.1.2`, and the documentation for that upstream gcc release lists the pragma:

https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Structure_002dPacking-Pragmas.html

(4) The "agent GUID" is used to populate the `EFI_SIGNATURE_DATA.SignatureOwner` field. The UEFI spec documents this field as follows: *An identifier which identifies the agent which added the signature to the list.*

Given that this extractor is the "same agent" for each of the extracted certificates, the same value should be used. (Obviously I tested this against the firmware and it works fine.)

Uniqueness is not required; for example Microsoft uses the same `77FA9ABD-0359-4D32-BD60-28F4E78F784B` SignatureOwner GUID for at least three of their certificates:
* `Microsoft Corporation KEK CA 2011`
* `Microsoft Windows Production PCA 2011`
* `Microsoft Corporation UEFI CA 2011`

(For completeness I shoudl mention that those CA certificates are not for web server authentication but for Secure Boot driver authentication.)

lersek:
OK, I'll check out the C language testers, thanks for the reference.

Regarding the avoidance of packed structures for better portability, I disagree. Packed structures are universally supported. They are used by all major projects that I'm involved with to various extents (edk2, QEMU, libvirt, kernel). Those projects build with multiple toolchains and for multiple OSes and architectures.

For example, edk2 (which uses `#pragma pack`) builds with every release of Visual Studio since VS2003, every release of GCC since 4.4, CLANG 3.5 and 3.8, GCC 4.3 on Cygwin, Intel C 9.1 and 11.1, ARM C 5.0, and XCODE 5.

Replacing the packed structures with stand-alone field writes doesn't just give up on "improvements" to code readability; it turns the code into an unreadable mess, and loses any resemblance of the UEFI spec (which is what *defines* the format).

A good example for such bad practice is `trust/extract-jks.c`, which:
* references a file by pathname in the "java sources" (whatever those might be) for the java keystore format, and
* writes naked fields with natural language comments such as `/* The creation date: current time */`.

This leaves the reader completely in the dark about the actual structure of the output file.

lersek:
@ueno what I can do -- if you *really* dislike structure packing -- is:
* keep the structures but drop `pragma #pack`
* keep the code mostly unchanged (i.e., populate the structures in memory)
* when serializing the structures to the P11 buffer, serialize them field-wise, rather than whole-sale. After all, `p11_buffer_add()` works on individual fields too, not just on the full struct level.

This will take care of any unexpected padding between fields and at the end of the structures, and still keep the code readable. Is this acceptable?

lersek:
BTW I think such custom encoders would be greatly helped if `common/buffer.h` offered convenience functions such as:
* p11_buffer_add_u8()
* p11_buffer_add_u16()
* p11_buffer_add_u32()

Can I add them?

ueno:
I don't strongly oppose to using packed structures, but my concern is that we sometimes receive bug reports on AIX or IRIX (#101) and I don't know if that pragma is available on such platforms.

Anyway, given that those structures are not too complicated, I was thinking something like:
```c
static bool buffer_add_uint16 (p11_buffer *buffer, uint16_t value);
static bool buffer_add_uint32 (p11_buffer *buffer, uint32_t value);
static bool buffer_add_efi_guid (p11_buffer *buffer, efi_guid *value);
static bool buffer_add_efi_signature_list (p11_buffer *buffer, efi_signature_list *value);
```
and:
```c
static bool
buffer_add_efi_guid (p11_buffer *buffer, efi_guid *value)
{
    if (!buffer_add_uint32 (buffer, value->data1)
      return false;
    ...
}
```
There are similar code in p11-kit/rpc-message.c.

lersek:
OK, thanks.

nmav:
@gnutlsmirror was me, sorry for the confusion. I was logged with the wrong account.

PR #139, opened 2018-03-29
Extract edk2/cacerts.bin (RHBZ#1559580) v2

lersek:
This is the followup pullreq to #137 (v1). Updates in this version:
* patch nr.1: no change
* patch nr.2: replace `#pragma pack` and the `XXX_to_le()` converters with `buffer_add_XXX()` helpers, covering both scalars (with conversion to LE) and structures (which need no packing this way) @ueno
* patch nr.3: add a unit test that extracts two certificates (new patch) @ueno @nmav 

Thanks.


lersek:
Thank you!

PR #140, opened 2018-04-05
modules: Add option to control module visibility from proxy

ueno:
This enables to control the visibility of a module loaded from the proxy module.  The configuration reuses the `enable-in` and `disable-in` options.  To enable/disable a module when loaded from the proxy module, set `p11-kit-proxy` as the value of the option.


ueno:
Thank you for the review.  Yes, it will not be loaded at all.  I have made it clearer in the commit log and the documentation.

PR #142, opened 2018-04-06
 trust: Prevent trust module being loaded by proxy module 

ueno:
Otherwise, when the proxy module were registered in NSS database, the
trust module would be loaded twice and affect search performance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,12 @@ sudo: required
 language: c
 
 env:
-  - BUILD_OPTS="--prefix=/usr --libdir=/usr/lib64" SRCDIR=/srcdir BUILDDIR=/builddir
-  - BUILD_OPTS="--prefix=/usr --libdir=/usr/lib64 --enable-coverage" COVERAGE=yes SRCDIR=/coverage BUILDDIR=/coverage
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -22,26 +22,34 @@ before_install:
   - docker exec $CONTAINER dnf -y install libasan libubsan
   - docker exec $CONTAINER dnf -y install clang-analyzer
   - docker exec $CONTAINER dnf -y install mingw64-gcc mingw64-libffi mingw64-libtasn1 wine
+  - docker exec $CONTAINER dnf -y install opensc openssl
```
tomato42: why not all installations are done at the same time? wouldn't it make them faster?

ueno: That's a good point; I arranged the `dnf install` invocation in https://github.com/p11-glue/p11-kit/pull/142/commits/05c257e5e71c78599cc6970c7a7438fd79bb863c.

jamescassell:
How is the trust module otherwise loaded?  It it automatically loaded by some other mechanism, or are users expected to register it explicitly in the nssdb separate from the proxy module (which is automatically registered.)

PR #152, opened 2018-05-14
Improve const correctness for P11KitUri

npmccallum:
This does not improve const for the getters. The reason for this is that
they are usually passed into the PKCS#11 APIs directly and these APIs
are not const correct. Trying to force const correctnesss here would
result in pain for library consumers.

This is an API and ABI compatible change.

### DIFF
--- a/p11-kit/uri.h
+++ b/p11-kit/uri.h
@@ -95,13 +95,13 @@ typedef struct p11_kit_uri p11_kit_uri;
 
 CK_INFO_PTR         p11_kit_uri_get_module_info             (P11KitUri *uri);
 
-int                 p11_kit_uri_match_module_info           (P11KitUri *uri,
-                                                             CK_INFO_PTR info);
+int                 p11_kit_uri_match_module_info           (const P11KitUri *uri,
+                                                             const CK_INFO *info);
 
 CK_SLOT_INFO_PTR    p11_kit_uri_get_slot_info               (P11KitUri *uri);
 
-int                 p11_kit_uri_match_slot_info             (P11KitUri *uri,
-                                                             CK_SLOT_INFO_PTR slot_info);
+int                 p11_kit_uri_match_slot_info             (const P11KitUri *uri,
+                                                             const CK_SLOT_INFO *slot_info);
 
 CK_SLOT_ID          p11_kit_uri_get_slot_id                 (P11KitUri *uri);
 void                p11_kit_uri_set_slot_id                 (P11KitUri *uri,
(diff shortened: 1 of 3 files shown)


npmccallum:
I believe this is a false positive. See here: https://stackoverflow.com/questions/5083765/does-changing-fmystruct-a-to-fconst-mystruct-a-breaks-api-abi-in-c

nmav:
I also think that const-addition changes don't affect ABI. abi-compliance-checker may be more explicit about it.

p11-kit/uri.h
```diff
@@ -130,16 +130,16 @@ int                 p11_kit_uri_set_attributes              (P11KitUri *uri,
...
 
-const char*         p11_kit_uri_get_pin_value              (P11KitUri *uri);
+const char*         p11_kit_uri_get_pin_value               (const P11KitUri *uri);
 
-void                p11_kit_uri_set_pin_value              (P11KitUri *uri,
+void                p11_kit_uri_set_pin_value               (P11KitUri *uri,
                                                              const char *pin);
 
-const char*         p11_kit_uri_get_pin_source              (P11KitUri *uri);
+const char*         p11_kit_uri_get_pin_source              (const P11KitUri *uri);
```
ueno: Should we add `const` to other `const char *` getters, such as `p11_kit_uri_get_module_name()` and `p11_kit_uri_get_module_path()`?

npmccallum:
@ueno I have added const to the following functions:
* `p11_kit_uri_get_pinfile()`
* `p11_kit_uri_get_module_name()`
* `p11_kit_uri_get_module_path()`
* `p11_kit_uri_get_vendor_query()`

I believe this now covers all of the low hanging fruit.

ueno:
Thank you!

asavah:
This breaks building gnutls 3.6.2 with p11-kit
```
In file included from ./../pkcs11_int.h:34:0,
                 from x509.c:37:
/home/asavah/kross/build/rpi3/rootfs/usr/include/p11-kit-1/p11-kit/uri.h:99:68: error: unknown type name 'CK_INFO'
                                                              const CK_INFO *info);
                                                                    ^~~~~~~
/home/asavah/kross/build/rpi3/rootfs/usr/include/p11-kit-1/p11-kit/uri.h:104:68: error: unknown type name 'CK_SLOT_INFO'
                                                              const CK_SLOT_INFO *slot_info);
                                                                    ^~~~~~~~~~~~
/home/asavah/kross/build/rpi3/rootfs/usr/include/p11-kit-1/p11-kit/uri.h:113:68: error: unknown type name 'CK_TOKEN_INFO'
                                                              const CK_TOKEN_INFO *token_info);
                                                                    ^~~~~~~~~~~~~
/home/asavah/kross/build/rpi3/rootfs/usr/include/p11-kit-1/p11-kit/uri.h:134:68: error: unknown type name 'CK_ATTRIBUTE'
                                                              const CK_ATTRIBUTE *attrs,
                                                                    ^~~~~~~~~~~~
Makefile:1512: recipe for target 'x509.lo' failed

```

nmav:
Ouch, indeed this patch also changes the types while adding the const. It seems that the block inside `#ifdef CRYPTOKI_GNU` has to be updated as well for the types used.  It seems that the testsuite doesn't compile any program with this definition on. Changing test-uri.c to define that, catches the issue.

PR #154, opened 2018-05-18
Address few existing documentation issues

nmav:


Makefile.am
```diff
@@ -122,5 +122,10 @@ upload-release: $(DIST_ARCHIVES)
 	gpg --detach-sign --local-user 'stef@thewalter.net' $<
 	scp $< $<.sig $(WEBHOST):$(WEBBASE)/releases/
 
+README:
+	cp README.md README
+
 transifex:
 	cd $(srcdir) && sh build/tx-update
+
+.PHONY: README
```
ueno: I don't think `.PHONY` is appropriate here; it's for non-filename target as far as I understand it.

Makefile.am
```diff
@@ -122,5 +122,10 @@ upload-release: $(DIST_ARCHIVES)
 	gpg --detach-sign --local-user 'stef@thewalter.net' $<
 	scp $< $<.sig $(WEBHOST):$(WEBBASE)/releases/
 
+README:
+	cp README.md README
```
ueno: Maybe this could be done as part of `dist-hook` defined above?

Makefile.am
nmav: removed

ueno:
Thank you!

PR #155, opened 2018-05-23
trust: Provide a drop-in replacement of gnome-keyring user trust storage

ueno:
GCR has an API to store user certificates on a token:
https://developer.gnome.org/gcr/3.20/gcr-Trust-Storage-and-Lookups.html

The implementation of the API relies on the fact that there is a token backed by per-user storage.

Although previously such token was provided by gnome-keyring, the upstream decided to stop installing the p11-kit configuration for the gnome-keyring's PKCS#11 module in 3.28:
https://bugzilla.gnome.org/show_bug.cgi?id=791401
which leads to a regression in applications such as Geary:
https://bugzilla.gnome.org/show_bug.cgi?id=736640

This patch exposes the "User Trust" token as a drop-in replacement of gnome-keyring's trust storage.

### DIFF
--- a/trust/builder.c
+++ b/trust/builder.c
@@ -922,8 +922,10 @@ const static builder_schema trust_schema = {
 	}, common_populate
 };
 
+/* CKO_X_TRUST_ASSERTION objects can be stored through PKCS #11, for
+ * compatibility reasons with gcr */
 const static builder_schema assertion_schema = {
-	GENERATED_CLASS,
+	NORMAL_BUILD,
 	{ COMMON_ATTRS,
 	  { CKA_X_PURPOSE, REQUIRE | CREATE },
 	  { CKA_X_CERTIFICATE_VALUE, CREATE },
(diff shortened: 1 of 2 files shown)


nmav:
Should that be documented in the manual or it is only for compatibility with gnomekeyring?

ueno:
Yes, this is primarily for compatibility with the applications which indirectly call gnome-keyring-pkcs11.so through the GCR API.

After thinking again, I have made the following changes to the previous version:
- don't enable "User Trust" by default, but leave it to user specifying `--with-trust-paths`; because the solution is temporary, it would make more sense to keep the feature not discoverable
- allow creating CKO_X_TRUST_ASSERTION objects through PKCS #11 (which were previously rejected)

ueno:
This will not happen in p11-kit; we have a better proposal:
https://gitlab.gnome.org/GNOME/gcr/-/issues/51

PR #160, opened 2018-05-27
pkcs11: Exercise GNU calling convention at compile time

### LINKED ISSUE #159: GNU calling convention broken
This "[Improve const correctness for P11KitUri](https://github.com/p11-glue/p11-kit/commit/e42dcf5283a5537c196147c9a2468ee537b9da7b)" commit appears to break compilation of `gnutls` due to use of `CRYPTOKI_GNU`.

ueno:
Fixes #159.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,6 +99,7 @@ x86_64-w64-mingw32
 /p11-kit/pkcs11.conf.example
 /p11-kit/p11-kit-remote@.service
 /p11-kit/virtual-fixed.c
+/p11-kit/*.gnu.c
 
 /p11-kit-remote
 /p11-kit-server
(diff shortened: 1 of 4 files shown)


p11-kit/pkcs11-gnu.c
```diff
@@ -0,0 +1,48 @@
...
+
+#include <stdlib.h>
+
+#define CRYPTOKI_GNU 1
+#define P11_KIT_FUTURE_UNSTABLE_API 1
+
+#include <p11-kit/p11-kit.h>
+#include <p11-kit/iter.h>
+#include <p11-kit/pin.h>
+#include <p11-kit/uri.h>
```
nmav: I'm wondering whether the compiler could ignore some type errors if the code is not used. nevertheless this seems to catch the existing issue.

ueno: I added references to the defined symbols so that the declarations are actually used.
Also this file is split into 3 (generated): iter.gnu.c, pin.gnu.c, and uri.gnu.c, so that the type declarations in other files do not take effect.

PR #161, opened 2018-05-27
Fix typedefs in the CRYPTOKI_GNU case

npmccallum:
Since we now use the non-PTR variants in function definitions, we need
to ensure that we properly typedef them if CRYPTOKI_GNU is set.

See: #152


ueno:
Thank you, but it's already been tracked as #160.

PR #164, opened 2018-05-28
common: do not use /proc on MacOS

lbschenkel:
I'm the maintainer for `p11-kit` in MacPorts. I'm submitting a patch we have downstream since 2014, to make it buildable on OS X 10.7 and earlier, and I believe it would be worth it to have it incorporated upstream. The relevant `opendir` call will always fail on macOS due to the absence of `/proc`, so nothing is lost by just `#ifdef`ing the code out.

### DIFF
--- a/common/compat.c
+++ b/common/compat.c
@@ -912,6 +912,7 @@ fdwalk (int (* cb) (void *data, int fd),
 	struct rlimit rl;
 #endif
 
+#ifdef __linux__
 	dir = opendir ("/proc/self/fd");
 	if (dir != NULL) {
 		while ((de = readdir (dir)) != NULL) {
@@ -934,6 +935,7 @@ fdwalk (int (* cb) (void *data, int fd),
 		closedir (dir);
 		return res;
 	}
+#endif
 
 	/* No /proc, brute force */
 #ifdef HAVE_SYS_RESOURCE_H


ueno:
Although the change looks okay, I still don't see why the original code doesn't compile on OS X 10.7 or earlier.  If `/proc` is absent, doesn't `opendir()` return NULL and the block is completely skipped, no?

lbschenkel:
Unfortunately I don't have a 10.7 machine to check myself, and I was not the maintainer when the patch was introduced. The original commit that introduces this patch mentions "breaks build on 10.7 and earlier", so assuming the message is accurate I am presuming that one of the methods in that block was not implemented in those macOS versions. I can get in touch with the original author of the patch to ask for clarification.

lbschenkel:
@dbevans: Do you happen to remember what exactly broke in macOS <= 10.7 without the `#ifdef`?

lbschenkel:
Unfortunately I could not get any additional clarification on the history behind this change. But I do believe that it's unlikely that somebody would have bothered to introduce this patch if they didn't have experienced a breakage as stated.

@ueno: Would you oppose merging this patch, given it's innocuous? It's a marginal thing, but it would help maintenance  if I could reduce the number of downstream patches to a minimum.

ueno:
I generally prefer not to add a code without a justification why it's necessary; otherwise upstream maintenance would be harder.

However, in this particular case, it might make some sense to skip the `opendir()` on non-Linux platforms, as it eliminates an unnecessary `access()` syscall.

lbschenkel:
I have amended the original commit and rebased. Please let me know if there any other suggestions.

PR #171, opened 2018-06-08
Fix cross compiling

neheb:
This broke OpenWrt building.


neheb:
there's a different solution. closing.

ueno:
would you be able to check if #170 works for you? if it does I can merge that PR.

neheb:
yeah it works.

PR #173, opened 2018-06-20
server: Enable socket activation through systemd

ueno:
This enables socket activation of "p11-kit server" through systemd.
The feature provided is essentially the same as commit
a4fb2bb5 (reverted), but implemented with "p11-kit server" and
libsystemd API instead of wrapping "p11-kit remote" in the unit file.

Note that, while it exposes all tokens through the socket, it doesn't
increase attack surface beyond the PKCS#11 binary interface provided
by p11-kit-proxy.so, because the service is per-user.


ueno:
Thank you for the review.

PR #174, opened 2018-06-27
 rpc-server: p11_kit_remote_serve_tokens: Allow exporting all modules 

ueno:
This amends #173. Previously, `p11-kit server` and `p11-kit remote` were not capable of serving tokens from across multiple modules, even if the user specify `p11-kit-proxy.so` as the argument (due to the recursive loading check).

This patch removes the limitation and changes the default behavior of those commands to serve tokens from all modules.

### DIFF
--- a/p11-kit/proxy.c
+++ b/p11-kit/proxy.c
@@ -95,7 +95,7 @@ typedef struct _State {
 } State;
 
 static State *all_instances = NULL;
-static State global = { { { { -1, -1 }, NULL, }, }, NULL, NULL, FIRST_HANDLE, NULL };
+static State global = { { { { -1, -1 }, NULL, }, }, NULL, NULL, NULL, FIRST_HANDLE, NULL };
 
 #define PROXY_VALID(px) ((px) && (px)->forkid == p11_forkid)
 #define PROXY_FORKED(px) ((px) && (px)->forkid != p11_forkid)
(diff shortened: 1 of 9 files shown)


p11-kit/proxy.c
```diff
@@ -248,7 +248,8 @@ modules_dup (CK_FUNCTION_LIST **modules)
 }
 
 static CK_RV
-proxy_create (Proxy **res, Mapping *mappings, unsigned int n_mappings)
+proxy_create (Proxy **res, CK_FUNCTION_LIST **loaded,
```
tomato42: I'm assuming this doesn't change API

p11-kit/rpc-server.c
```diff
@@ -2086,124 +2087,200 @@ p11_kit_remote_serve_token (CK_FUNCTION_LIST *module,
...
+static void
+token_set_write_protected (CK_TOKEN_INFO *token, P11KitUri *uri)
+{
+	const char *write_protected =
+		p11_kit_uri_get_vendor_query (uri, "write-protected");
+	if (write_protected &&
+	    _p11_conf_parse_boolean (write_protected, false))
+		token->flags |= CKF_WRITE_PROTECTED;
+	else
+		token->flags &= ~CKF_WRITE_PROTECTED;
```
tomato42: I'm assuming that token->flags and CFK_WRITE_PROTECTED are the same here...

p11-kit/test-server.c
```diff
@@ -101,19 +108,23 @@ setup_server (void *arg)
...
+	p11_array_push (args, NULL);
 
 	test.pid = fork ();
 
 	/* The child */
 	if (test.pid == 0) {
 		close (STDOUT_FILENO);
 		dup2 (fds[0], STDOUT_FILENO);
-		execv (BUILDDIR "/p11-kit-server-testable", argv);
+		execv (BUILDDIR "/p11-kit-server-testable", (char **)args->elem);
```
tomato42: error checking missing

p11-kit/test-server.c
```diff
@@ -101,19 +108,23 @@ setup_server (void *arg)
 	if (envvar)
 		setenv ("LD_PRELOAD", envvar, 1);
 
-	argv[5] = test.socket_path;
-	argv[6] = (char *)arg;
+	p11_array_push (args, "-n");
+	p11_array_push (args, test.socket_path);
+	p11_array_push (args, fixture->token);
+	p11_array_push (args, NULL);
```
tomato42: error checking missing for `p11_array_push` in the test server

p11-kit/test-server.c
ueno: Fixed.

p11-kit/test-server.c
ueno: Fixed.

p11-kit/proxy.c
ueno: Yes, this shouldn't affect the API, although the semantics of `p11_kit_remote_serve_tokens()` becomes slightly different.  Previously it served only the first matching module, while it now serves all modules; I'd expect it's an acceptable change.

p11-kit/rpc-server.c
ueno: Sorry I don't get it; the `else` clause is to turn off the flag in `token->flags`.

ueno: I don't think integer promotion is possible here; as `token->flags` is defined as `CK_ULONG` and `CKF_WRITE_PROTECTED` is defined as `1UL << 1`.

ueno:
Thank you for the review!

PR #177, opened 2018-07-20
Remove hard dependency on libpthread

ueno:
This implements the suggestion in: https://bugzilla.redhat.com/show_bug.cgi?id=985495#c0

### DIFF
--- a/common/compat.c
+++ b/common/compat.c
@@ -169,8 +169,9 @@ getprogname (void)
 #include <fcntl.h>
 #include <unistd.h>
 
+#ifndef PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP
 void
-p11_mutex_init (p11_mutex_t *mutex)
+p11_recursive_mutex_init (p11_mutex_t *mutex)
 {
 	pthread_mutexattr_t attr;
 	int ret;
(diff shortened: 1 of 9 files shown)


ueno:
@fweimer would you take a look at this when you have time?

common/compat.c
```diff
@@ -169,6 +169,14 @@ getprogname (void)
 #include <fcntl.h>
 #include <unistd.h>
 
+#ifdef PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP
+static pthread_mutex_t mutex_initializer = PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP;
+void
+p11_mutex_init (p11_mutex_t *mutex)
+{
+	memcpy (mutex, &mutex_initializer, sizeof (pthread_mutex_t));
```
fweimer: I think this is technically undefined behavior. Copying of mutex objects is not supported.

ueno: Thank you.  The initializer is only applicable to statically allocated mutexes, right?  In p11-kit there are a few places where dynamically allocated mutexes are used; how could we initialize them?

ueno: After going through the uses of mutexes, I realized that only `p11_library_mutex` (and possibly `p11_virtual_mutex`) needs to be recursive (see 563606efe17cbf3b84679f5e54f60b8d68ba9015). I have changed all other initialization to use the default attribute.

ueno:
Thank you for the review. I have replaced the use of `pthread_atfork()` with `__register_atfork()`, which glibc provides,

PR #181, opened 2018-08-14
Update pkcs11 header to allow SoftHSMv2 to compile

abbra:
This patch updates common/pkcs11.h with constants and structure definitions that are required for SoftHSMv2 to be compiled using a copy of this header.

It would be nice to have these changes in p11-kit as well, to avoid maintenance effort.

### DIFF
--- a/common/pkcs11.h
+++ b/common/pkcs11.h
@@ -193,6 +193,20 @@ extern "C" {
 #define source_data pSourceData
 #define source_data_len ulSourceDataLen
 
+#define counter_bits ulCounterBits
+#define iv_ptr pIv
+#define iv_len ulIvLen
+#define iv_bits ulIvBits
+#define aad_ptr pAAD
+#define aad_len ulAADLen
+#define tag_bits ulTagBits
+#define shared_data_len ulSharedDataLen
+#define shared_data pSharedData
+#define public_data_len ulPublicDataLen
+#define public_data pPublicData
+#define string_data pData
+#define string_data_len ulLen
+#define data_params pData
 #endif	/* CRYPTOKI_COMPAT */
 
 
@@ -371,6 +385,24 @@ typedef unsigned long ck_key_type_t;
 #define CKK_AES			(0x1fUL)
 #define CKK_BLOWFISH		(0x20UL)
 #define CKK_TWOFISH		(0x21UL)
+#define CKK_SECURID		(0x22UL)
+#define CKK_HOTP		(0x23UL)
+#define CKK_ACTI		(0x24UL)
+#define CKK_CAMELLIA		(0x25UL)
+#define CKK_ARIA		(0x26UL)
+#define CKK_MD5_HMAC		(0x27UL)
+#define CKK_SHA_1_HMAC		(0x28UL)
+#define CKK_RIPEMD128_HMAC	(0x29UL)
+#define CKK_RIPEMD160_HMAC	(0x2aUL)
+#define CKK_SHA256_HMAC		(0x2bUL)
+#define CKK_SHA384_HMAC		(0x2cUL)
+#define CKK_SHA512_HMAC		(0x2dUL)
+#define CKK_SHA224_HMAC		(0x2eUL)
+#define CKK_SEED		(0x2fUL)
+#define CKK_GOSTR3410		(0x30UL)
+#define CKK_GOSTR3411		(0x31UL)
+#define CKK_GOST28147		(0x32UL)
+#define CKK_EC_EDWARDS		(0x40UL)
 #define CKK_VENDOR_DEFINED	((unsigned long) (1UL << 31))
 
 
@@ -381,6 +413,7 @@ typedef unsigned long ck_certificate_type_t;
 #define CKC_WTLS		(2UL)
 #define CKC_VENDOR_DEFINED	((unsigned long) (1UL << 31))
 
+#define CKC_OPENPGP		(CKC_VENDOR_DEFINED|0x504750UL)
 
 typedef unsigned long ck_attribute_type_t;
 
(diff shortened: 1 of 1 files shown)


abbra:
@ueno these constants are as in the latest 2.40 errata 1: http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/include/pkcs11-v2.40/pkcs11t.h, your link is for a draft of 2.40, not a final one.

common/pkcs11.h
```diff
@@ -371,16 +371,33 @@ typedef unsigned long ck_key_type_t;
...
+#define CKK_SHA384_HMAC		(0x2cUL)
+#define CKK_SHA512_HMAC		(0x2dUL)
+#define CKK_SHA224_HMAC		(0x2eUL)
+#define CKK_SEED		(0x2fUL)
+#define CKK_GOSTR3410		(0x30UL)
+#define CKK_GOSTR3411		(0x31UL)
+#define CKK_GOST28147		(0x32UL)
 #define CKK_VENDOR_DEFINED	((unsigned long) (1UL << 31))
-
+#define CKK_EC_EDWARDS		(CKK_VENDOR_DEFINED|0x8003UL)
```
nmav: This seems to be part of PKCS#11 3.0.

common/pkcs11.h
```diff
@@ -748,27 +835,41 @@ typedef unsigned long ck_mechanism_type_t;
...
+#define CKM_CAMELLIA_CBC_ENCRYPT_DATA	(0x557UL)
+
+#define CKM_AES_KEY_WRAP		(0x2109UL)
+#define CKM_AES_KEY_WRAP_PAD		(0x210aUL)
+
+#define CKM_RSA_PKCS_TPM_1_1		(0x4001UL)
+#define CKM_RSA_PKCS_OAEP_TPM_1_1	(0x4002UL)
+
+#define CKM_EC_EDWARDS_KEY_PAIR_GEN	(CKM_VENDOR_DEFINED|0x9040UL)
+#define CKM_EDDSA			(CKM_VENDOR_DEFINED|0x9041UL)
```
nmav: That's also a value to be defined in PKCS#11 3.0

common/pkcs11.h
nmav: https://github.com/oasis-tcs/pkcs11/blob/master/working/3-00-wd-01/pkcs11t.h

common/pkcs11.h
```diff
@@ -428,6 +446,7 @@ typedef unsigned long ck_attribute_type_t;
 #define CKA_EXPONENT_1			(0x126UL)
 #define CKA_EXPONENT_2			(0x127UL)
 #define CKA_COEFFICIENT			(0x128UL)
+#define CKA_PUBLIC_KEY_INFO		(0x129UL)
```
nmav: This is also defined in pkcs11x.h. It makes sense to have it here, but we may need to protect it in case a file includes both headers.

abbra: pkcs11x.h is wrapping its definition with `#ifndef CKA_PUBLIC_KEY_INFO`. In files where both `pkcs11.h` and `pkcs11x.h` are included, the order is to have `pkcs11.h` first.

common/pkcs11.h
bellgrim: CKK_EC_EDWARDS was added to SoftHSM before PKCS#11 3.0. That is why it was a vendor defined value. Nothing that should be spread to other implementations. SoftHSM should be updated with the values from 3.0.

common/pkcs11.h
bellgrim: Yes, SoftHSM should be updated with values from 3.0

common/pkcs11.h
nmav: Also according to the chair of the working group these are the final values, and are not expected to change.

abbra: Great, thanks for verifying, @nmav. I'll move vendor-specific values for EC EDWARDS and EDDSA to use 3.0 version and resubmit.

common/pkcs11.h
```diff
@@ -856,6 +958,61 @@ struct ck_rsa_pkcs_oaep_params {
   unsigned long source_data_len;
 };
 
+struct ck_aes_ctr_params {
+  unsigned long ulCounterBits;
```
ueno: nit: we use the GNU coding convention for structure members; see `ck_rsa_pkcs_pss_params.hash_alg` and [the renaming macro](https://github.com/p11-glue/p11-kit/pull/181/commits/cf669c242a1a9a646128d65440edf883a00a36fd#diff-79ec9ebeae42b2c808fb40814768ecc3R191) for example.

abbra:
@ueno, I did change structures where possible. Do you foresee more changes?

common/pkcs11.h
ueno: Sorry for not responding earlier; it's not a blocker (just looked different from other part of the header).  Given that the only consumer of this convention is GnuTLS and it doesn't (yet) use the newly added structs, I guess it would be fine to keep the original names.

ueno:
@abbra I see a few mix-ups of `iv` and `pIv` there, but otherwise it looks good to me

abbra:
As I said, they are not mix-ups because we cannot define `iv` to be `pIv` in the compatibility mode as it will override what should be `iv` member of the other structures. We have to live with those, I guess.

nmav:
What about using ivec defined to pIv? (in general all these definitions seem risky)

ueno:
@abbra I meant this line:
https://github.com/p11-glue/p11-kit/pull/181/commits/ec6100e4e0bf0bd349ca0d04f03aa3d63cfd0d79#diff-79ec9ebeae42b2c808fb40814768ecc3R982
and:
https://github.com/p11-glue/p11-kit/pull/181/commits/ec6100e4e0bf0bd349ca0d04f03aa3d63cfd0d79#diff-79ec9ebeae42b2c808fb40814768ecc3R1020
https://github.com/p11-glue/p11-kit/pull/181/commits/ec6100e4e0bf0bd349ca0d04f03aa3d63cfd0d79#diff-79ec9ebeae42b2c808fb40814768ecc3R1026

ueno:
Ah, sorry I was completely lost; so both `iv` and `pIv` are in the reference header, though the former is an array while the latter is a pointer. Then I suppose something like `iv_ptr` might be a better name.

> in general all these definitions seem risky

I agree. Now that the file is diverging from the scute one, I wonder if it might make sense to generate a separate header with the GNU style definitions from the original definitions at compile time.

abbra:
I changed `pIv` to `iv_ptr` and did the same to `pAAD` by changing it to `aad_ptr`.

PR #180, opened 2018-08-13
 proxy: Avoid invalid memory access when unloading proxy module

ueno:
When loading and unloading p11-kit-proxy.so with pkcs11-tool, it
accesses already free'd memory area:
```
$ valgrind pkcs11-tool --module p11-kit-proxy.so -L
==25173== Invalid read of size 8
==25173==    at 0x64BF493: p11_proxy_module_cleanup (proxy.c:1724)
==25173==    by 0x64BD028: _p11_kit_fini (proxy-init.c:65)
==25173==    by 0x401477C: _dl_close_worker (in /usr/lib64/ld-2.27.so)
==25173==    by 0x4014E1D: _dl_close (in /usr/lib64/ld-2.27.so)
==25173==    by 0x5E08C4E: _dl_catch_exception (in /usr/lib64/libc-2.27.so)
==25173==    by 0x5E08CDE: _dl_catch_error (in /usr/lib64/libc-2.27.so)
==25173==    by 0x58B1724: _dlerror_run (in /usr/lib64/libdl-2.27.so)
==25173==    by 0x58B1113: dlclose (in /usr/lib64/libdl-2.27.so)
==25173==    by 0x11E5A7: ??? (in /usr/bin/pkcs11-tool)
==25173==    by 0x110023: ??? (in /usr/bin/pkcs11-tool)
==25173==    by 0x5CF624A: (below main) (in /usr/lib64/libc-2.27.so)
==25173==  Address 0x61231c8 is 552 bytes inside a block of size 584 free'd
==25173==    at 0x4C2FDAC: free (vg_replace_malloc.c:530)
==25173==    by 0x6548492: p11_virtual_unwrap (virtual.c:2902)
==25173==    by 0x64BF492: p11_proxy_module_cleanup (proxy.c:1723)
```

PR #182, opened 2018-08-15
Resolves: rhbz#1615616 crash during __run_exit_handlers

caolanm:
in p11_proxy_module_cleanup on calling p11_virtual_unwrap state has the same
value as wrapper->virt has in p11_virtual_unwrap when its passed to
wrapper->destroyer, so dereferencing state after callling p11_virtual_unwrap
results in junk


ueno:
Thank you for reporting this; it seems the fix is almost identical to #180. Would you take a look at it?

PR #186, opened 2018-08-21
README.md: Mention contact method for security issues

ueno:
This is a mandatory criterion for [the CII Best Practices](https://bestpractices.coreinfrastructure.org/en). As far as I know github doesn't support confidential issues, so I am assigning myself as the contact for now.


nmav:
Makes sense

PR #188, opened 2018-08-24
travis: Fix errors after gcc update in fedora:latest image

ueno:


ueno:
Thank you for the review!

PR #194, opened 2018-10-16
Fix issues spotted by coverity

ueno:


p11-kit/proxy.c
```diff
@@ -612,7 +612,8 @@ proxy_C_OpenSession (CK_X_FUNCTION_LIST *self,
 				sess->wrap_slot = map.wrap_slot;
 				sess->real_session = *handle;
 				sess->wrap_session = ++state->last_handle; /* TODO: Handle wrapping, and then collisions */
-				p11_dict_set (state->px->sessions, &sess->wrap_session, sess);
+				if (!p11_dict_set (state->px->sessions, &sess->wrap_session, sess))
```
nmav: I'd have suggested to use (void) cast here, but then I saw this:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425

trust/save.c
```diff
@@ -105,7 +106,9 @@ p11_save_open_file (const char *path,
 	if (asprintf (&temp, "%s%s.XXXXXX", path, extension) < 0)
 		return_val_if_reached (NULL);
 
+	mode = umask (066);
```
nmav: It would be nice to have the intended permissions documented explicitly as the `066` is not very explanatory. btw. `066` seems wrong as I'd have expected 3 digits after zero. It also ignores the running user umask which may be stricter than what you set there (not know if that could be an issue, but it is not apparent to me how this function is used).

common/debug.h
```diff
@@ -71,13 +71,13 @@ void              p11_debug_precond             (const char *format,
 #endif
 
 #define return_val_if_fail(x, v) \
-	do { if (!(x)) { \
+	do { if (x) { } else {							\
```
nmav: Seems ok but ugly. Is it necessary to address such false-positives?

nmav:
LGTM with a question mark on umask.

common/debug.h
ueno: Yes, the same pattern can be found in the corresponding glib macros.

trust/save.c
ueno: I agree that we should document it (as a comment at least). For other two points: the C standard only says that a literal starting with `0` is treated as octal (so 066 == 0066). Also, I thought that a new umask is ORed with the existing umask so it would never become more permissive.

ueno: Sorry, the last part was my misunderstanding. I guess using `0077` would be better, as it is strictest and self-explanatory.

nmav: ok

ueno:
Thank you for the review!

PR #195, opened 2018-10-18
virtual: Add envvar to disable libffi closure

ueno:
On some occasions (e.g., p11-kit-proxy is loaded from server
programs), it is known that libffi closures are always unusable, and
that causes unnecessary SELinux AVCs in journal.

This patch adds a means to disable it through the
P11_KIT_DISABLE_LIBFFI_CLOSURE envvar.


ueno:
Superseded by #196

PR #196, opened 2018-10-19
virtual: Prefer fixed closures to libffi closures

ueno:
On some circumstances (such as when loading p11-kit-proxy from httpd),
it is known that creation of libffi closure always fails, due to
SELinux policy.  Although this is harmless, it pollutes the journal
and gives wrong hints when troubleshooting.  This patch changes the
order of preference of libffi vs pre-compiled closures to avoid that.


simo5:
LGTM

ueno:
Thank you for the review.

PR #206, opened 2018-12-23
trust: Improve error handling if backed trust file is corrupted

### LINKED ISSUE #205: trust pem file misparsed
Hello,

this is https://bugs.debian.org/914199 submitted by Sam Morris:
XXXXXXXX Quote
    $ grep BEGIN /etc/ssl/certs/ca-certificates.crt  | wc -l
    154

    $ trust list; echo $0
    0

That's from p11-kit 0.23.14-2. If I use the version from stable there is at
least a clue that something is amiss:

    # trust list; echo $?
    p11-kit: ca-certificates.crt: BEGIN ...: pem block before p11-kit section header
    p11-kit: ca-certificates.crt: BEGIN ...: pem block before p11-kit section header
    0

It turns out that, in-between the PEM-encoded certificates in
ca-certificates.crt, I have some lines:

    ...
    -----END CERTIFICATE-----
    # This file was created by IPA. Do not edit.

    [p11-kit-object-v1]
    class: certificate
    certificate-type: x-509
    certificate-category: authority
    label: "CN%3Dipa-CA%2CDC%3Dipa%2CDC%3Dexample%2CDC%3Dcom"
    subject: "..."
    issuer: "..."
    serial-number: "..."
    x-public-key-info: "..."
    trusted: true
    -----BEGIN CERTIFICATE-----
    ...

These are in turn taken from the file that ipa-client-install dropped into
/usr/local/share/ca-certificates/ipa-ca.crt.

IMO p11-kit should treat these extra lines as comments since other tools
(openssl, gnutls) are perfectly happy to ignore them.

It would also be nice if it printed some more useful output to help users
debug issues such as these, and not exit with status 0 if problems are
detected. :)
XXXXXXXX End Quote

On Debian p11-kit is configured with `--with-trust-paths=/etc/ssl/certs/ca-certificates.crt`. Looking at the documentation for the trust module https://p11-glue.github.io/p11-glue/p11-kit/manual/trust-module.html afaict p11-kit should indeed interpret these as RFC 7468 PEM files, ignoring anything outside BEGIN/END markers.

ueno:
This PR does 3 things: first, tell p11-kit-trust.so the tool's verbosity so it can correctly print messages to stderr; secondly, make C_FindObjectsInit fail early on parse error, and finally, if the trust file is not parsable in the p11-kit persistent format proceed with the next formats (DER and PEM).

Fixes #205

### DIFF
--- a/trust/module.c
+++ b/trust/module.c
@@ -287,6 +287,11 @@ parse_argument (char *arg,
 		free (gl.paths);
 		gl.paths = value ? strdup (value) : NULL;
 
+	} else if (strcmp (arg, "verbose") == 0) {
+		if (strcmp (value, "yes") == 0)
+			p11_message_loud ();
+		else if (strcmp (value, "no") == 0)
+			p11_message_quiet ();
 	} else {
 		p11_message ("unrecognized module argument: %s", arg);
 	}
--- a/trust/module.c
+++ b/trust/module.c
@@ -1198,11 +1198,16 @@ sys_C_FindObjectsInit (CK_SESSION_HANDLE handle,
 				indices[n++] = session->index;
 			if (want_token_objects) {
 				if (!session->loaded)
-					p11_token_load (session->token);
-				session->loaded = CK_TRUE;
-				indices[n++] = p11_token_index (session->token);
+					if (p11_token_load (session->token) < 0)
+						rv = CKR_FUNCTION_FAILED;
+				if (rv == CKR_OK) {
+					session->loaded = CK_TRUE;
+					indices[n++] = p11_token_index (session->token);
+				}
 			}
+		}
 
+		if (rv == CKR_OK) {
 			find = calloc (1, sizeof (FindObjects));
 			warn_if_fail (find != NULL);
 
--- a/common/message.c
+++ b/common/message.c
@@ -58,7 +58,7 @@
 #include <stdio.h>
 #include <string.h>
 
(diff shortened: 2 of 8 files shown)


trust/module.c
```diff
@@ -287,6 +287,11 @@ parse_argument (char *arg,
 		free (gl.paths);
 		gl.paths = value ? strdup (value) : NULL;
 
+	} else if (strcmp (arg, "verbose") == 0) {
+		if (strcmp (value, "yes") == 0)
+			p11_message_loud ();
+		else if (strcmp (value, "no") == 0)
```
ep69: Wouldn't it have more sense to call `p11_message_quiet` in every other case other that `value` being `"yes"`? Or is it guaranteed somewhere that the values could only be `"yes"` or `"no"`?

ep69:
Actually, now I also realize that I wasn't able to spot changes related to "if the trust file is not parsable in the p11-kit persistent format proceed with the next formats (DER and PEM)".

trust/module.c
ueno: There is technically no reason to not accept other values, but I think it would be good to be consistent with  [`_p11_conf_parse_boolean()`](https://github.com/p11-glue/p11-kit/blob/master/p11-kit/conf.c#L500) which considers other values invalid.

ueno:
Thank you for the review!

PR #208, opened 2019-01-09
trust: Fix alias generation in JKS extractor

ueno:
When there is a duplicate, the JKS extractor previously assigned somewhat obscure name `"-<digit>"` (not `"<name>-<digit>"`).


ueno:
Thank you for the review!

PR #217, opened 2019-02-27
modules: fix a SIGSEGV checking gl.modules before iterates on it when freeing

stefano-garzarella:
In some circumstances, as described in the BZ, can happen that
free_modules_when_no_refs_unlocked() is called multiple times
when the module destructor is invoked.
We should check gl.modules before iterates on it in the
free_modules_when_no_refs_unlocked() functions, to avoid
a SIGSEGV.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1680963

### DIFF
--- a/p11-kit/modules.c
+++ b/p11-kit/modules.c
@@ -797,14 +797,16 @@ init_globals_unlocked (void)
 static void
 free_modules_when_no_refs_unlocked (void)
 {
-	Module *mod;
-	p11_dictiter iter;
+	if (gl.modules) {
+		Module *mod;
+		p11_dictiter iter;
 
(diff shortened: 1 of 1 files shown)


ueno:
Thank you; looks good to me.
It would be nice if this comes with a test, but I suppose it would be tricky.

stefano-garzarella:
> Thank you; looks good to me.
> It would be nice if this comes with a test, but I suppose it would be tricky.

I agree. I didn't know your test suite, but I'll try to implement a test for this case.
BTW, I tried the test-proxy and it also was affected to the segfault. Now it exits well, but a lot of assert fails (I'm not sure it is updated).

PR #218, opened 2019-03-09
Fix Win32 p11_dl_error crash

tomsci:
Caused by returning a buffer that wasn't allocated with malloc and
needed to be freed with LocalFree() instead. The fix is to strdup
msg_buf so what's returned can be free()d.


ueno:
Nice catch, thank you!

PR #219, opened 2019-03-12
modules: Fix index used in call to p11_dict_remove()

simonjhaggett:
This fixes a call to p11_dict_remove() in managed_steal_sessions_inlock() to use
the correct index in the stolen array (i, rather than at). This avoids an
assert, which was encountered on a host serving a PKCS#11 module to a remote
Linux client.

Signed-off-by: Simon Haggett <simon.haggett@gmail.com>


ueno:
Indeed a good catch, thank you!

PR #224, opened 2019-05-15
proxy: Refresh slot list on every C_GetSlotList call

### LINKED ISSUE #221: Proxy: Update slot list during each call to C_GetSlotList()
When an application calls `C_Initialize()` in the proxy module, the proxy calls `C_GetSlotList()` in all of the p11-kit modules and assigns each slot a "wrapped" slot ID. The proxy never checks for new slots in the p11-kit modules after this; it always returns the same slots whenever the application calls `C_GetSlotList()`. (However the proxy will filter the returned slot list if the `tokenPresent` parameter is `TRUE`; to do this it will call `C_GetSlotInfo()` for each slot to determine if a token is present.)

Although this is [valid behavior in PKCS#11](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959741), it imposes a usability limitation that does not exist if the individual PKCS#11 modules (such as OpenSC) are installed directly in the NSS database instead.

For example, if Firefox and/or Thunderbird are open, and the user now plugs in a Yubikey or attaches the laptop to a docking station connected to a USB keyboard with a smart card reader, these will not be usable by Firefox or Thunderbird via the proxy until those applications are completely closed and re-opened.

ueno:
Previously, the proxy module calculated the slot list only once at the `C_Initialize()` call.  That was causing a usability limitation when the user attaches HSM after starting an application.

Suggested by David Ward.

Fixes #221.

### DIFF
--- a/p11-kit/Makefile.am
+++ b/p11-kit/Makefile.am
@@ -408,7 +408,8 @@ check_LTLIBRARIES += \
 	mock-three.la \
 	mock-four.la \
 	mock-five.la \
-	mock-seven.la
+	mock-seven.la \
+	mock-eight.la
 
 mock_one_la_SOURCES = p11-kit/mock-module-ep.c
 mock_one_la_LIBADD = libp11-test.la libp11-common.la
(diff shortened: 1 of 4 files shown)


ep69:
Looks good to me, r+

ueno:
Thank you for the review.

PR #225, opened 2019-05-16
proxy: Support C_WaitForSlotEvent() if CKF_DONT_BLOCK is specified

### LINKED ISSUE #222: Proxy: Handle non-blocking calls to C_WaitForSlotEvent()
Currently, all calls to `C_WaitForSlotEvent()` in the proxy module [return `CKR_FUNCTION_NOT_SUPPORTED`](https://github.com/p11-glue/p11-kit/blob/793cc3b78f17bb5a3c151eba1144b73a5d51be3e/p11-kit/proxy.c#L576).

While _blocking_ calls might not be practical to handle, is anything known that would prevent _non-blocking_ calls from being handled correctly (where `CKF_DONT_BLOCK` is set in `flags`)?

Some applications appear not to work if the PKCS#11 module cannot handle non-blocking `C_WaitForSlotEvent()`. In particular I have noticed this with VMware Horizon Client for Linux.

ueno:
While fully implementing `C_WaitForSlotEvent()` would require a separate thread to monitor events, it is straightforward to implement the function if the `CKF_DONT_BLOCK` flag is given.

Suggested by David Ward.

Fixes #222

### DIFF
--- a/p11-kit/proxy.c
+++ b/p11-kit/proxy.c
@@ -604,7 +604,40 @@ proxy_C_WaitForSlotEvent (CK_X_FUNCTION_LIST *self,
                           CK_SLOT_ID_PTR slot,
                           CK_VOID_PTR reserved)
 {
-	return CKR_FUNCTION_NOT_SUPPORTED;
+	State *state = (State *)self;
+	Proxy *py = state->px;
+	CK_FUNCTION_LIST_PTR *f;
+	CK_FUNCTION_LIST_PTR funcs;
+	CK_SLOT_ID real_slot;
+	unsigned int i;
+	CK_RV rv = CKR_NO_EVENT;
+
+	/* Only the non-blocking case is supported. */
+	if ((flags & CKF_DONT_BLOCK) == 0)
+		return CKR_FUNCTION_NOT_SUPPORTED;
+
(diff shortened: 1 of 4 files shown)


p11-kit/proxy.c
```diff
@@ -604,7 +604,40 @@ proxy_C_WaitForSlotEvent (CK_X_FUNCTION_LIST *self,
...
+
+	p11_lock ();
+
+	for (f = py->inited; *f; ++f) {
+		funcs = *f;
+		assert (funcs != NULL);
+
+		rv = (funcs->C_WaitForSlotEvent) (flags, &real_slot, reserved);
+		if (rv == CKR_NO_EVENT)
+			continue;
```
simo5: This looks like a busy waiting loop, would it make sense to add some sleeping with a exponential backoff ?

ueno: I think it should be okay, because the loop is calling `C_WaitForSlotEvent` for different modules one after another.

simo5: Ah sorry,  imissed that this cycles through all modules only once. Nevermind

ueno:
Thank you for the review!

PR #227, opened 2019-05-19
build: Update gettext infrastructure to 0.20.1

ueno:
This removes languages with no translation, as the new gettext infrastructure doesn't tolerate them:
https://lists.gnu.org/archive/html/bug-gettext/2019-05/msg00124.html

Reported by Don Lawrence in:
https://lists.freedesktop.org/archives/p11-glue/2019-May/000700.html


ueno:
I withdraw this, as the msgmerge behavior turned out to be not intended.

PR #229, opened 2019-05-22
doc: Add 'server' command in help

rmedaer:
`server` is the last common command which is not in CLI help.
IMHO, adding this small documentation could help to promote usage of pkcs11 forwarding.


ueno:
Thank you!

PR #231, opened 2019-06-02
Port to meson build system

ueno:
This adds support for building with meson in addition to autotools:
https://wiki.gnome.org/Initiatives/GnomeGoals/MesonPorting

It basically works if you do:
```sh
$ meson _build
$ meson test -C _build
```
I am marking it as WIP, until the following tasks are completed:

- [x] `ninja dist`  succeeds
- [x] port to symbol versioning
- [x] `issetugid` check is fixed (mesonbuild/meson#5408)
- [x] possibly rewrite PKCS#11 GNU convention tests
- [x] run tests written in shell-script
- [x] cross-compilation works


p11-kit/meson.build
```diff
@@ -0,0 +1,337 @@
...
+             variables: [
+               'p11_module_configs=@0@'.format(p11_package_config_modules),
+               'p11_module_path=@0@'.format(p11_module_path),
+               # 'p11_trust_paths=@0@'.format(with_trust_paths),
+               'proxy_module=@0@/p11-kit-proxy.so'.format(libdir),
+               'p11_system_config_modules=@0@'.format(p11_package_config_modules)
+             ])
+
+pkcs11_conf_example_conf = configuration_data()
+pkcs11_conf_example_conf.set('p11_system_config_file', p11_system_config_file)
```
inigomartinez: Since meson 0.49 a [dictionary can be used when `set` is always used](http://mesonbuild.com/Reference-manual.html#configuration_data).

p11-kit/meson.build
```diff
@@ -0,0 +1,337 @@
...
+pkcs11_conf_example_conf.set('p11_user_config', p11_user_config)
+pkcs11_conf_example = configure_file(input: 'pkcs11.conf.example.in',
+                                     output: 'pkcs11.conf.example',
```
inigomartinez: Alternatively `@BASENAME@` can be used so it's less error prone in name changes.

common/meson.build
```diff
@@ -0,0 +1,89 @@
...
+  'lexer.c', 'lexer.h',
+  'message.c', 'message.h',
+  'path.c', 'path.h',
+  'pkcs11.h', 'pkcs11x.h', 'pkcs11i.h',
+  'runtime.c', 'runtime.h',
+  'url.c', 'url.h',
+  'init.h'
+]
+
+libp11_common = static_library('libp11-common.a', libp11_common_sources,
```
inigomartinez: This static library is used in `p11-kit/meson.build` file. Why not create an internal dependency for it?

common/meson.build
```diff
@@ -0,0 +1,89 @@
+install_headers('pkcs11.h', 'pkcs11x.h', subdir: 'p11-kit-1/p11-kit')
+
+libp11_common_sources = [
+  'argv.c', 'argv.h',
```
inigomartinez: It is not necessary to list headers in source file list.

common/meson.build
```diff
@@ -0,0 +1,89 @@
...
+]
+
+libp11_common = static_library('libp11-common.a', libp11_common_sources,
+                               include_directories: [configinc, commoninc])
+
+libp11_library_sources = [
+  'library.c', 'library.h'
+]
+
+libp11_library = static_library('libp11-library.a', libp11_library_sources,
```
inigomartinez: Similar to `libp11_common`, this static library is used in `p11-kit/meson.build` file. Why not create an internal dependency for it?

common/meson.build
```diff
@@ -0,0 +1,89 @@
...
+  'test-url',
+  'test-path',
+  'test-lexer',
+  'test-message',
+  'test-argv',
+  'test-runtime'
+]
+
+foreach name : common_tests
+  t = executable(name, '@0@.c'.format(name),
```
inigomartinez: Instead of using string formatting what about using string concatenation here?

```suggestion
  t = executable(name, name + '.c',
```

common/meson.build
```diff
@@ -0,0 +1,89 @@
...
+  'message.c', 'message.h',
+  'path.c', 'path.h',
+  'pkcs11.h', 'pkcs11x.h', 'pkcs11i.h',
+  'runtime.c', 'runtime.h',
+  'url.c', 'url.h',
+  'init.h'
+]
+
+libp11_common = static_library('libp11-common.a', libp11_common_sources,
+                               include_directories: [configinc, commoninc])
```
inigomartinez: meson already includes de current source directory, so there is no need to also include `commoninc`.

common/meson.build
inigomartinez: There is no need to include `library.h` because the compiler will found it, so only `library.c` is needed. The fact that it's only one file, why don't you include directly in the `static_library` call? It will avoid the creation/memory of `libp11_library_sources` variable.

```suggestion
libp11_library = static_library('libp11-library.a', 'library.c',
```

common/meson.build
```diff
@@ -0,0 +1,89 @@
...
+libp11_library = static_library('libp11-library.a', libp11_library_sources,
+                                include_directories: [configinc, commoninc])
+
+libp11_test_sources = [
+  'mock.c', 'mock.h',
+  'test.c', 'test.h'
+]
+
+libp11_test = static_library('libp11-test.a', libp11_test_sources,
+                             include_directories: [configinc, commoninc])
```
inigomartinez: There is no need to include `commoninc`.

common/meson.build
```diff
@@ -0,0 +1,89 @@
...
+libp11_tool_sources = [
+  'tool.c', 'tool.h'
+]
+
+if host_system != 'windows'
+  libp11_tool_sources += ['unix-peer.c', 'unix-peer.h']
+endif
+
+libp11_tool = static_library('libp11-tool.a', libp11_tool_sources,
+                             include_directories: [configinc, commoninc])
```
inigomartinez: There is no need to include `commoninc`.

common/meson.build
```diff
@@ -0,0 +1,89 @@
...
+  'test-lexer',
+  'test-message',
+  'test-argv',
+  'test-runtime'
+]
+
+foreach name : common_tests
+  t = executable(name, '@0@.c'.format(name),
+                 c_args: tests_c_args,
+                 include_directories: [configinc, commoninc],
```
inigomartinez: There is no need to include `commoninc`.

common/meson.build
```diff
@@ -0,0 +1,89 @@
...
+  test(name, t)
+endforeach
+
+common_progs = [
+  'frob-getauxval',
+  'frob-getenv'
+]
+
+foreach name : common_progs
+  executable(name, '@0@.c'.format(name),
```
inigomartinez: Instead of using string formatting what about using string concatenation here?

```suggestion
  executable(name, name + '.c',
```

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
+if get_option('doc')
```
inigomartinez: Why don't you move this condition to the main `meson.build` file so it avoids entering and parsing this file in case the option is `false`?

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
...
+    'rpc-message.h',
+    'util.h',
+    'virtual.h',
+    'virtual-fixed.h',
+    'array.h',
+    'compat.h'
+  ]
+
+  content_files = [
+    'p11-kit-config.xml', 'p11-kit-sharing.xml',
```
inigomartinez: As as suggestion, just to be consistent, wouldn't be better to write one file for each line?

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
...
+    'p11-kit-trust.xml',
+    'p11-kit-remoting.xml',
+    'p11-kit.xml',
+    'pkcs11.conf.xml',
+    'trust.xml',
+    'annotation-glossary.xml'
+  ]
+
+  version_conf = configuration_data()
+  version_conf.set('VERSION', meson.project_version())
```
inigomartinez: Since meson 0.49 you can use dicts for `configuration_data` objects only using `set`. So this would just be:

```suggestion
  version_conf = {'VERSION': meson.project_version()}
```

And being only one line, why don't just write it in the `configure_file` call?

```suggestion
  configure_file(
    input : 'version.xml.in',
    output : 'version.xml',
    configuration : {'VERSION': meson.project_version()})
```

Finally, you can also use `@BASENAME@` as output value, that removes the last extension:


```suggestion
  configure_file(
    input : 'version.xml.in',
    output : '@BASENAME@',
    configuration : {'VERSION': meson.project_version()})
```

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
...
+  configure_file(
+    input : 'version.xml.in',
+    output : 'version.xml',
+    configuration : version_conf)
+
+  userdir_conf = configuration_data()
+  userdir_conf.set('p11_user_config', p11_user_config)
```
inigomartinez: Similar to the `version_conf` example.

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
...
+  configure_file(
+    input : 'userdir.xml.in',
+    output : 'userdir.xml',
+    configuration : userdir_conf)
+
+  sysdir_conf = configuration_data()
+  sysdir_conf.set('p11_system_config', p11_system_config)
```
inigomartinez: Similar to the `version_conf` example.

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
...
+  configure_file(
+    input : 'sysdir.xml.in',
+    output : 'sysdir.xml',
+    configuration : sysdir_conf)
+
+  gnome.gtkdoc('p11-kit',
```
inigomartinez: When comparing the `autotools` generated documentation and `meson` generated documentation, reference urls are relative instead of absolutes.

This can be fixed by using the `fixxref_args` and setting `--extra-dir` parameter. Setting `--html-dir` is also a good idea.

```suggestion
  glib_prefix = dependency('glib-2.0').get_pkgconfig_variable('prefix')

  fixxref_args = [
    '--html-dir=' + (prefix / gnome.gtkdoc_html_dir(meson.project_name())),
    '--extra-dir=' + (glib_prefix / gnome.gtkdoc_html_dir('glib')),
  ]

  gnome.gtkdoc(
    ...
    fixxref_args: fixxref_args,
    ...
```

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
...
+               main_xml : 'p11-kit-docs.xml',
+               namespace : 'p11_kit',
+               src_dir : 'p11-kit',
+               dependencies : libffi_deps + libdl_deps,
+               scan_args : [
+                 '--ignore-headers=' + ' '.join(ignore_headers),
+                 '--rebuild-types',
+                 '--deprecated-guards="P11_KIT_DISABLE_DEPRECATED"'
+               ],
+               scanobjs_args : [
```
inigomartinez: There is no need to use an array if only one parameter is used.

```suggestion
               scanobjs_args : '--version',
```

trust/meson.build
```diff
@@ -0,0 +1,165 @@
...
+
+asn_sources = [
+  'basic.asn',
+  'pkix.asn',
+  'openssl.asn'
+]
+
+asn_h_deps = []
+foreach name : asn_sources
+  asn_h = custom_target('generate @0@.h'.format(name),
```
inigomartinez: I've never used it properly but what about using [generator](http://mesonbuild.com/Reference-manual.html#generator) here?

trust/meson.build
```diff
@@ -0,0 +1,165 @@
...
+                        output: '@0@.h'.format(name),
+                        input: name,
+                        command: [
+                          asn1Parser, '-o', '@OUTPUT@', '@INPUT@'
+                        ])
+  asn_h_dep = declare_dependency(sources: asn_h)
```
inigomartinez: I wouldn't use an internal dependency just for source files. My reasoning behind internal dependencies, in a similar way to external dependencies, the usually hold linking parameters (libraries), header directories (include directories), compiler flags and sometimes also sources, so I use them just to hold generated libraries among with their include directories and external dependencies and also generated source files. When grouping source files I just use the `files` function or just an array.

trust/meson.build
```diff
@@ -0,0 +1,165 @@
...
+              sources: libtrust_sources + ['module-init.c'],
+              include_directories: [configinc, commoninc],
+	      c_args: p11_kit_trust_c_args,
+              dependencies: libtasn1_deps,
+              link_with: [libtrust_data, libp11_library, libp11_common],
+              install: true,
+              install_dir: p11_module_path)
+
+libtrust_testable_c_args = [
+  '-DP11_DEFAULT_TRUST_PREFIX="@0@\"'.format(join_paths(meson.current_build_dir(), 'default')),
```
inigomartinez: There is an extra ' \' character?

```suggestion
  '-DP11_DEFAULT_TRUST_PREFIX="@0@"'.format(join_paths(meson.current_build_dir(), 'default')),
```

Also remember that since meson 0.49, you can just use '/' to join directories avoiding the use of `join_paths`.

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
...
+    'p11-kit.xml',
+    'pkcs11.conf.xml',
+    'trust.xml',
+    'annotation-glossary.xml'
+  ]
+
+  version_conf = configuration_data()
+  version_conf.set('VERSION', meson.project_version())
+  configure_file(
+    input : 'version.xml.in',
```
inigomartinez: In other `meson.build` files there wasn't an space before `:`. It would be nice to make this consistent,.

My personal preference is to avoid spaces before the `parameter`/`value` separator.

p11-kit/meson.build
```diff
@@ -0,0 +1,337 @@
...
+                            install: true,
+                            version: library_version,
+                            include_directories: [configinc, commoninc],
+                            dependencies: libffi_deps + libdl_deps,
+                            link_args: libp11_kit_ldflags,
+                            link_depends: libp11_kit_symbol_map,
+                            link_with: [libp11_kit_internal])
+
+meson.add_install_script(
+  'meson_post_install.sh',
```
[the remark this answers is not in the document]
ueno: I'd keep the script in `p11-kit/p11-kit` subdir, because the script is specific to that subdir.

p11-kit/meson.build
```diff
@@ -0,0 +1,337 @@
...
+
+test('test-messages.sh',
+     find_program('test-messages.sh'),
+     env: p11_kit_tests_env)
+
+test('test-server.sh',
+     find_program('test-server.sh'),
+     env: p11_kit_tests_env)
+
+shared_module('mock-one',
```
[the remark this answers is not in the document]
ueno: Nice, I've rewritten with that loop.

meson.build
```diff
@@ -0,0 +1,392 @@
...
+else
+  conf.set('NDEBUG', 1)
+endif
+
+if get_option('strict')
+  conf.set('WITH_STRICT', 1)
+endif
+
+prefix = get_option('prefix')
+datadir = join_paths(prefix, get_option('datadir'))
```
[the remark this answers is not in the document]
ueno: Yes, using relative paths throughout the code sounds reasonable.  I've rewritten those.

meson.build
```diff
@@ -0,0 +1,392 @@
...
+  endif
+
+  # These are things we can work around
+  headers = [
+    'sys/resource.h',
+    'ucred.h'
+  ]
+
+  foreach h : headers
+    if cc.has_header(h)
```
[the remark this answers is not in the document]
ueno: Hmm, this seems not that simple, because the C code assumes the `HAVE_FOO_H` is undefined (not set to 0).

meson.build
```diff
@@ -0,0 +1,392 @@
...
+    'strndup',
+    'strnstr',
+    'vasprintf',
+  ]
+
+  foreach f : functions
+    if cc.has_function(f)
+      define = 'HAVE_' + f.underscorify().to_upper()
+      conf.set(define, 1)
+      set_variable('have_func_' + f, true)
```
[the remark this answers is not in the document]
ueno: Removed.

meson.build
```diff
@@ -0,0 +1,392 @@
...
+endif
+
+# --------------------------------------------------------------------
+# libtasn1 support
+
+with_libtasn1 = false
+libtasn1_deps = []
+libtasn1_opt = get_option('libtasn1')
+if not libtasn1_opt.disabled()
+  libtasn1 = dependency('libtasn1', version: '>= 2.3')
```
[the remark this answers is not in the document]
ueno: Nice tip, this makes the logic much simpler.

meson.build
```diff
@@ -0,0 +1,392 @@
...
+  if libfreebl3.found() and cc.has_function('NSSLOW_Init',
+                                            dependencies: libfreebl3)
+    conf.set('WITH_FREEBL', 1)
+  else
+    error('could not find the freebl3 library')
+  endif
+endif
+
+# --------------------------------------------------------------------
+# Trust Module
```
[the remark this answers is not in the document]
ueno: Makes sense; merged those checks.

meson_options.txt
```diff
@@ -0,0 +1,51 @@
+option('hash-impl', type : 'combo',
```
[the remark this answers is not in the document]
ueno: Good point, converted to use `_`.

p11-kit/meson.build
```diff
@@ -0,0 +1,337 @@
...
+  'proxy.c',
+  'proxy.h',
+  'proxy-init.c',
+  'rpc-server.c'
+]
+
+libp11_kit_symbol_map = join_paths(meson.current_source_dir(), 'libp11-kit.map')
+libp11_kit_ldflags = cc.get_supported_link_arguments([
+  '-Wl,--version-script,' + libp11_kit_symbol_map,
+  '-Wl,-no-undefined'
```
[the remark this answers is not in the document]
ueno: OK, removed it.

common/meson.build
ueno: Good idea, it eliminated the need of listing `include_directories` in the dependent targets.

common/meson.build
ueno: Removed.

trust/meson.build
ueno: Good catch.

p11-kit/meson.build
```diff
@@ -0,0 +1,337 @@
...
+]
+
+virtual_fixed_generated_h = custom_target('generate virtual-fixed-generated.h',
+                                          output: 'virtual-fixed-generated.h',
+                                          command: [
+                                            '@0@/gen-virtual-fixed.sh'.format(meson.current_source_dir()),
+                                            '@OUTPUT@', '64'
+                                          ])
+
+virtual_fixed_dep = declare_dependency(sources: virtual_fixed_generated_h)
```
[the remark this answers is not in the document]
ueno: I had been struggling with this.  The situation is that there is a dependency between a .c source file and a generated .h file (i.e., the .c file includes the generated .h file).  Without an explicit dependency, meson doesn't seem to handle it well; it compiles the .c file before the .h file is generated.

ueno:
Thank you very much for the thorough review and great suggestions!

meson.build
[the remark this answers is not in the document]
inigomartinez: ok! I've rechecked it, my fault. In this case `set` method should be used instead of `set10`, so meson doesn't define it at all, which is perfect for `ifdef`.

p11-kit/meson.build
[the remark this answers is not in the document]
ueno: Sorry, it turned out to be my confusion; I had mistakenly put `sources:` in the `library` target. If I change it to a plain argument, it works.

p11-kit/meson.build
ueno: Nice tip, rewritten with it.

doc/manual/meson.build
ueno: Nice tip, rewritten with it.

doc/manual/meson.build
ueno: Thank you, incorporated this.

trust/meson.build
ueno: Indeed, switched to listing it in sources.

trust/meson.build
ueno: Yes, generator fits well in this use-case.

inigomartinez:
Both [sysdir.xml.in](https://github.com/p11-glue/p11-kit/pull/231/files?file-filters%5B%5D=.am&file-filters%5B%5D=.build&file-filters%5B%5D=.c&file-filters%5B%5D=.in&file-filters%5B%5D=.map&file-filters%5B%5D=.sh&file-filters%5B%5D=.txt&file-filters%5B%5D=No+extension&owned-by%5B%5D=#diff-6d687cd02f22ad68f0cfcee32026fae5) and [userdir.xml.in](https://github.com/p11-glue/p11-kit/pull/231/files?file-filters%5B%5D=.am&file-filters%5B%5D=.build&file-filters%5B%5D=.c&file-filters%5B%5D=.in&file-filters%5B%5D=.map&file-filters%5B%5D=.sh&file-filters%5B%5D=.txt&file-filters%5B%5D=No+extension&owned-by%5B%5D=#diff-30761a444d2282b18855a00b7f326697) are used to store just a path. Why not use just one file and replace it with the proper path?

This can also be done for `version.xml` file.

```suggestion
echo '@VALUE@' > content.xml.in
```
```suggestion
in_files = [
  {'name': 'sysdir.xml', 'value': p11_system_config},
  {'name': 'userdir.xml', 'value': p11_user_config},
  {'name': 'version.xml', 'value': meson.project_version()},
]

foreach in_file: in_files
  content_files += configure_file(
    input: 'content.xml.in',
    output: in_file['name'],
    configuration: {'VALUE': in_file['value']})
endforeach
```

This way you will reduce 3 template files to just one. (BTW: Although not necessary because they are already build in the proper directory but these generated files are also content files, that's why the `content_files +=`).

doc/manual/meson.build
```diff
@@ -0,0 +1,75 @@
...
+  'p11-kit-proxy.xml',
+  'p11-kit-trust.xml',
+  'p11-kit-remoting.xml',
+  'p11-kit.xml',
+  'pkcs11.conf.xml',
+  'trust.xml',
+  'annotation-glossary.xml'
+]
+
+configure_file(
```
inigomartinez: Both [sysdir.xml.in](https://github.com/p11-glue/p11-kit/pull/231/files?file-filters%5B%5D=.am&file-filters%5B%5D=.build&file-filters%5B%5D=.c&file-filters%5B%5D=.in&file-filters%5B%5D=.map&file-filters%5B%5D=.sh&file-filters%5B%5D=.txt&file-filters%5B%5D=No+extension&owned-by%5B%5D=#diff-6d687cd02f22ad68f0cfcee32026fae5) and [userdir.xml.in](https://github.com/p11-glue/p11-kit/pull/231/files?file-filters%5B%5D=.am&file-filters%5B%5D=.build&file-filters%5B%5D=.c&file-filters%5B%5D=.in&file-filters%5B%5D=.map&file-filters%5B%5D=.sh&file-filters%5B%5D=.txt&file-filters%5B%5D=No+extension&owned-by%5B%5D=#diff-30761a444d2282b18855a00b7f326697) are used to store just a path. Why not use just one file and replace it with the proper path?

This can also be done for `version.xml` file.

```
echo '@VALUE@' > content.xml.in
```
```suggestion
in_files = [
  {'name': 'sysdir.xml', 'value': p11_system_config},
  {'name': 'userdir.xml', 'value': p11_user_config},
  {'name': 'version.xml', 'value': meson.project_version()},
]

foreach in_file: in_files
  content_files += configure_file(
    input: 'content.xml.in',
    output: in_file['name'],
    configuration: {'VALUE': in_file['value']})
endforeach
```

This way you will reduce 3 template files to just one. (BTW: Although not necessary because they are already build in the proper directory but these generated files are also content files, that's why the `content_files +=`).

meson.build
```diff
@@ -0,0 +1,355 @@
...
+  ]
+
+  foreach f : functions
+    if cc.has_function(f)
+      conf.set('HAVE_' + f.underscorify().to_upper(), 1)
+    endif
+  endforeach
+
+  if cc.has_header_symbol('stdio.h', 'asprintf', prefix: '#define _GNU_SOURCE')
+    conf.set('HAVE_DECL_ASPRINTF', 1)
```
inigomartinez: Remember that you can use `set10` here:

```suggestion
    conf.set10('HAVE_DECL_ASPRINTF', cc.has_header_symbol('stdio.h', 'asprintf', prefix: '#define _GNU_SOURCE'))
```

meson.build
```diff
@@ -0,0 +1,355 @@
...
+  else
+    conf.set('HAVE_DECL_ASPRINTF', 0)
+  endif
+
+  if cc.has_header_symbol('stdio.h', 'vasprintf', prefix: '#define _GNU_SOURCE')
+    conf.set('HAVE_DECL_VASPRINTF', 1)
```
inigomartinez: You can also use `set10` here:

```suggestion
    conf.set10('HAVE_DECL_VASPRINTF', cc.has_header_symbol('stdio.h', 'vasprintf', prefix: '#define _GNU_SOURCE'))
```

meson.build
```diff
@@ -0,0 +1,355 @@
...
+  else
+    conf.set('HAVE_DECL_VASPRINTF', 0)
+  endif
+
+  if cc.has_header_symbol('stdlib.h', 'reallocarray')
+    conf.set('HAVE_DECL_REALLOCARRAY', 1)
```
inigomartinez: You can also use `set10` here:

```suggestion
    conf.set10('HAVE_DECL_REALLOCARRAY', cc.has_header_symbol('stdlib.h', 'reallocarray'))
```

p11-kit/meson.build
```diff
@@ -0,0 +1,290 @@
...
+libp11_kit_internal_sources = [
+  'conf.c',
+  'log.c',
+  'filter.c',
+  'rpc-transport.c',
+  'rpc-message.c',
+  'rpc-client.c'
+]
+
+virtual_fixed_generated_h = custom_target('generate virtual-fixed-generated.h',
```
inigomartinez: You can add directly the result of the `custom_target` to the sources so you avoid creating an extra variable:

```suggestion
libp11_kit_internal_sources += custom_target('generate virtual-fixed-generated.h',
```

doc/manual/meson.build
ueno: Thank you, but for me it looks a bit over-engineered.

ueno:
I am going to merge this now that most of the functionality is working (except coverage target - coveralls doesn't seem to like out-of-tree builds). Thank you again @inigomartinez for the review and suggestions!

PR #237, opened 2019-07-01
common: Fix uClibc-ng compilation

neheb:
program_invocation_short_name is const under uClibc-ng.

```
common/compat.c:100:14: error: conflicting types for 'program_invocation_short_name'
 extern char *program_invocation_short_name;
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from common/compat.c:52:
/opt/buildbot/slaves/lede-slave-tah/arc_arc700/build/sdk/staging_dir/toolchain-arc_arc700_gcc-8.3.0_uClibc/include/errno.h:54:46: note: previous declaration of 'program_invocation_short_name' was here
 extern const char *program_invocation_name, *program_invocation_short_name;
                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Makefile:3501: recipe for target 'common/compat.lo' failed
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -135,9 +135,9 @@ if test "$os_unix" = "yes"; then
 		[AC_MSG_ERROR([could not find required gmtime_r() function])])
 
 	# Check if these are declared and/or available to link against
-	AC_CHECK_DECLS([program_invocation_short_name])
+	AC_CHECK_DECLS([program_invocation_short_name]), [], [], [#include <error.h>])
 	AC_MSG_CHECKING([whether program_invocation_short_name is available])
-	AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <argp.h>]],
+	AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <errno.h>]],
 	                                [[program_invocation_short_name = "test";]])],
 	               [AC_DEFINE([HAVE_PROGRAM_INVOCATION_SHORT_NAME], [1],
 	                          [Whether program_invocation_short_name available])


common/compat.c
```diff
@@ -93,8 +93,12 @@
 #include <unistd.h>
 
 #if defined (HAVE_PROGRAM_INVOCATION_SHORT_NAME) && !HAVE_DECL_PROGRAM_INVOCATION_SHORT_NAME
+#ifdef __UCLIBC__
+extern const char *program_invocation_short_name;
```
nmav: The fact that there is a conflicting definition, means that HAVE_DECL_PROGRAM_INVOCATION_SHORT_NAME doesn't work well. configure.ac seems to be checking for it in argp.h, while the manpage in fedora documents it in `errno.h`

neheb: I noticed that while I was trying to fix this. Both musl and uClibc-ng specify program_invocation_short_name in errno.h  . Changing it there did not help.

neheb:
Updated the patch to handle __progname too. Some configurations of uClibc-ng do not define program_invocation_short_name

common/compat.c
nmav: Wouldn't it need defining _GNU_SOURCE prior to including the file?

neheb: Looks like yes.

ueno: Would it help if you change the configure check to:
```sh
# Check if these are declared and/or available to link against
AC_CHECK_DECLS([program_invocation_short_name]), [], [], [#include <error.h>])
```

neheb: I assume you meant errno.h . Added.

ueno: No, actually.  The condition `#if defined (HAVE_PROGRAM_INVOCATION_SHORT_NAME) && !HAVE_DECL_PROGRAM_INVOCATION_SHORT_NAME` means that the symbol is found in the library but the declaration in a header file is missing. If uClibc has a declaration, p11-kit should use it rather than redefining it.

The `AC_LINK_IFELSE` check is for library symbol, so it is not relevant here. Look for `AC_CHECK_DECLS` a few lines above.

neheb: I think this is what you meant.

ueno: Thanks for the update, but do we still need the changes in common/compat.c?

neheb: Looks like no. Will update.

ueno:
Thank you!

Issue #244, opened 2019-09-10
CK_DATE - memory allocation error


vjardin:
Hi,

When I have a CK_DATE, then we get this memory failure that is recorded here https://github.com/p11-glue/p11-kit/blob/master/p11-kit/rpc-message.c#L988

For this issue, value_length is 0 byte which of course is not a validé one. How/when should value_length be set ?

Thank you,

ueno:
It shall be set by the caller of any PKCS#11 function that takes a CK_DATE attribute (typically a part of attribute array, e.g., `C_CreateObject`). How did you reproduce it?

nicesoft-vj:
Using pkcs-dump dump

vjardin:
Using pkcs-dump dump command.

ueno:
Do you mean [this](https://github.com/alonbl/pkcs11-dump)?
Note that the code path is only taken if you set up a module through `p11-kit remote` or `p11-kit server`. Do you have such configuration?

vjardin:
Yes, correct for this pkcs-dump repo. It is a good stress test.

I use p11-kit remote (over ssh).

ueno:
Which PKCS#11 module do you actually remote (e.g., opensc-pkcs11.so, something proprietary)?

vjardin:
I use a proprietary one ; I do not have an opensc compatible card to try with.

From the code I do not understand why a wrong value can lead to such memory issue. Moreover, according to oasis pkcs11 specification, I remember that a request with a value 0 means "get/return the size of the memory", doesn't it ?

vjardin:
For details, see some traces I added from the pkcs11-dump/Dump() function:
`Dump:1149 C_GetAttributeValue(17, 1, {type=272, p=0x7fffc3d71af0, ulValueLen=10240}, 1)`

then, we get these traces:

> (p11-kit:2638) write_at: ok: wrote block of 33
> (p11-kit:2638) read_at: ok: read block of 12
> (p11-kit:2638) read_at: ok: read block of 6
> (p11-kit:2638) read_at: ok: read block of 40
> (p11-kit:2638) rpc_C_GetAttributeValue: GetAttributeValue: enter
> proto_read_attribute_buffer:276 ulValueLen=10240/value
> p11_rpc_buffer_add_attribute:1066
> p11_rpc_buffer_add_attribute:1068
> p11_rpc_buffer_add_attribute:1070 value_type=4 attr->ulValueLen=0
[...]

PR #246, opened 2019-09-10
Fix CK_DATE when date is empty

vjardin:
Fix the issue #244 : one should return an empty buffer whenever
its length is 0.


vjardin:
The travis failure is not related to this pull request. Should it work ?

ueno:
Yes, the CI failures should be fixed with #249. Would you like to rebase against the master?

vjardin:
Yes, I'll do it asap

vjardin:
Hi @ueno ,
Please, can we consider this pull request despite the CI failures which are not related ? I do not have any means to fix the CI.

thank you,

ueno:
Looks like the error on the macOS CI is actually a test failure:
```
FAIL: test-rpc 9 /rpc/byte-array-null
# assertion failed (NULL == (void*)val): (0x00000000 == 0x7f9c494028cb)
# test-rpc: assertion failed (NULL == (void*)val): (0x00000000 == 0x7f9c494028cb)
# in test_byte_array_null() at test-rpc.c:297
# test-rpc: in test_byte_array_null() at test-rpc.c:297
```
Maybe, we should somehow distinguish the cases where the value is empty or zero-length. Let me think about it a bit.

vjardin:
Hi @ueno 

What do you think we should change in order to support both states : empty vs zero length ?  I am asking because I do not understand what's wrong.

ueno:
Sorry @vjardin, the problem is that we also need to account for decoding the empty value. I've submitted #253 to do that.

vjardin:
let's close this pull request, A better fix has been provided by @ueno with the pull request #253

PR #247, opened 2019-09-11
autotools: adjust feature parity after meson

ueno:
This adds symbol versioning and also include meson files in autotools distribution.


vjardin:
LGTM

ueno:
Thanks for checking!

PR #251, opened 2019-09-17
Upgrade compilation for Microsoft C v19.16.27026.1

vjardin:
The Windows support seems to be very old. Many features had been obsoleted. Let's start refreshing. It is still WIP, but I prefer to start pushing it so it'll ease the following with non Windows code.

### DIFF
--- a/common/compat.h
+++ b/common/compat.h
@@ -112,6 +112,11 @@ void         p11_dl_close       (void * dl);
 
 #include <io.h>
 
+#ifndef __GNUC__
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#endif
+
 /* Oh ... my ... god */
 #undef CreateMutex
 
--- a/common/compat.h
+++ b/common/compat.h
@@ -166,6 +166,8 @@ void        p11_mmap_close  (p11_mmap *map);
 
 #define strdup(s) _strdup(s)
 
+#define getpid() _getpid()
+
 #endif /* OS_WIN32 */
 
 /* ----------------------------------------------------------------------------
--- a/common/compat.h
+++ b/common/compat.h
@@ -172,6 +172,8 @@ void        p11_mmap_close  (p11_mmap *map);
 #define setenv(n, v, z) _putenv_s(n, v)
 #endif /* HAVE_SETENV */
 
+#define strdup(s) _strdup(s)
+
 #endif /* OS_WIN32 */
 
 /* ----------------------------------------------------------------------------
--- a/common/compat.h
+++ b/common/compat.h
@@ -166,7 +166,10 @@ void        p11_mmap_close  (p11_mmap *map);
 
 #define strdup(s) _strdup(s)
 
+#ifndef __GNUC__
 #define getpid() _getpid()
+typedef int pid_t;
+#endif
 
 #endif /* OS_WIN32 */
 
--- a/common/compat.h
+++ b/common/compat.h
@@ -163,6 +163,7 @@ void        p11_mmap_close  (p11_mmap *map);
 #ifndef HAVE_SETENV
 #define setenv(n, v, z) _putenv_s(n, v)
 #endif /* HAVE_SETENV */
+#define getenv2(name) secure_getenv(name)
 
 #define strdup(s) _strdup(s)
 
@@ -179,6 +180,8 @@ typedef int pid_t;
 #define write(fd, buf, len) _write(fd, buf, (unsigned int)(len))
 
 #else
+/* Microsoft stdlib.h has deprecated getenv(), let's assumes getenv2() */
+#define getenv2(name) getenv(name)
 #endif /* OS_WIN32 */
 
 /* ----------------------------------------------------------------------------
--- a/common/compat.h
+++ b/common/compat.h
@@ -174,6 +174,11 @@ typedef int pid_t;
 #define STDIN_FILENO 0
 #define STDOUT_FILENO 1
 
+/* microsoft Windows uses ISO definitions, not POSIX anymore */
+#define read(fd, buf, len) _read(fd, buf, (unsigned int)(len))
+#define write(fd, buf, len) _write(fd, buf, (unsigned int)(len))
+
+#else
 #endif /* OS_WIN32 */
 
 /* ----------------------------------------------------------------------------
--- a/common/compat.h
+++ b/common/compat.h
@@ -171,6 +171,9 @@ void        p11_mmap_close  (p11_mmap *map);
 typedef int pid_t;
 #endif
 
+#define STDIN_FILENO 0
+#define STDOUT_FILENO 1
+
 #endif /* OS_WIN32 */
 
 /* ----------------------------------------------------------------------------
--- a/common/compat.h
+++ b/common/compat.h
@@ -99,14 +99,6 @@ void         p11_dl_close       (void * dl);
 
 #ifdef OS_WIN32
 
-#ifndef _WIN32_WINNT
-#define _WIN32_WINNT 0x500
-#endif
-
-#ifndef _WIN32_IE
-#define _WIN32_IE 0x500
-#endif
-
 #define WIN32_LEAN_AND_MEAN 1
 #include <windows.h>
 
--- a/common/lexer.c
+++ b/common/lexer.c
@@ -46,6 +46,7 @@
 #include <assert.h>
 #include <ctype.h>
 #include <errno.h>
+#include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -121,7 +122,7 @@ p11_lexer_next (p11_lexer *lexer,
 				lexer->tok_type = TOK_PEM;
 				lexer->tok.pem.begin = lexer->at;
 				lexer->tok.pem.length = end - lexer->at;
-				assert (end - lexer->at <= lexer->remaining);
+				assert (end - lexer->at <= (ptrdiff_t)lexer->remaining);
 				lexer->remaining -= (end - lexer->at);
 				lexer->at = end;
 				return true;
@@ -140,7 +141,7 @@ p11_lexer_next (p11_lexer *lexer,
 			lexer->remaining = 0;
 			lexer->at = end;
 		} else {
-			assert ((end - lexer->at) + 1 <= lexer->remaining);
+			assert ((end - lexer->at) + 1 <= (ptrdiff_t)lexer->remaining);
 			lexer->remaining -= (end - lexer->at) + 1;
 			lexer->at = end + 1;
 		}
--- a/common/path.c
+++ b/common/path.c
@@ -38,6 +38,11 @@
 
 #include "config.h"
 
+#ifdef OS_WIN32
+#include <winsock2.h>
+#include <shlobj.h>
+#endif
+
 #include "buffer.h"
 #include "debug.h"
 #include "message.h"
@@ -53,10 +58,6 @@
 #ifdef OS_UNIX
 #include <pwd.h>
 #include <unistd.h>
-#endif
-
-#ifdef OS_WIN32
-#include <shlobj.h>
 #endif
 
 
@@ -122,12 +123,12 @@ expand_homedir (const char *remainder)
 	if (remainder != NULL &&
 	    strncmp (remainder, ".config", 7) == 0 &&
 	    is_path_component_or_null (remainder[7])) {
-		env = getenv ("XDG_CONFIG_HOME");
+		env = getenv2 ("XDG_CONFIG_HOME");
 		if (env && env[0])
 			return p11_path_build (env, remainder + 8, NULL);
 	}
 
-	env = getenv ("HOME");
+	env = getenv2 ("HOME");
 	if (env && env[0]) {
 		return p11_path_build (env, remainder, NULL);
 
@@ -303,7 +304,7 @@ bool
 p11_path_prefix (const char *string,
                  const char *prefix)
 {
-	int a, b;
+	size_t a, b;
 
 	return_val_if_fail (string != NULL, false);
 	return_val_if_fail (prefix != NULL, false);
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -119,7 +119,7 @@ p11_rpc_message_prep (p11_rpc_message *msg,
                       int call_id,
                       p11_rpc_message_type type)
 {
-	int len;
+	size_t len;
 
 	assert (type != 0);
 	assert (call_id >= P11_RPC_CALL_ERROR);
@@ -210,7 +210,7 @@ bool
 p11_rpc_message_verify_part (p11_rpc_message *msg,
                              const char* part)
 {
-	int len;
+	size_t len;
 	bool ok;
 
 	if (!msg->sigverify)
@@ -646,7 +646,7 @@ p11_rpc_buffer_add_uint32 (p11_buffer *buffer,
 {
 	size_t offset = buffer->len;
 	if (!p11_buffer_append (buffer, 4))
-		return_val_if_reached ();
+		return_if_reached ();
 	p11_rpc_buffer_set_uint32 (buffer, offset, value);
 }
 
@@ -715,10 +715,11 @@ p11_rpc_buffer_add_byte_array (p11_buffer *buffer,
 		p11_rpc_buffer_add_uint32 (buffer, 0xffffffff);
 		return;
 	} else if (length >= 0x7fffffff) {
+	  /* Check if length can be converted to uint32_t. */
 		p11_buffer_fail (buffer);
 		return;
 	}
-	p11_rpc_buffer_add_uint32 (buffer, length);
+	p11_rpc_buffer_add_uint32 (buffer, (uint32_t)length);
 	p11_buffer_add (buffer, data, length);
 }
 
@@ -939,7 +940,7 @@ p11_rpc_buffer_add_attribute_array_value (p11_buffer *buffer,
 	}
 
 	/* Write the number of items */
-	p11_rpc_buffer_add_uint32 (buffer, count);
+	p11_rpc_buffer_add_uint32 (buffer, (uint32_t)count);
 
 	/* Actually write the attributes.  */
 	for (i = 0; i < count; i++) {
@@ -964,7 +965,7 @@ p11_rpc_buffer_add_mechanism_type_array_value (p11_buffer *buffer,
 	}
 
 	/* Write the number of items */
-	p11_rpc_buffer_add_uint32 (buffer, count);
+	p11_rpc_buffer_add_uint32 (buffer, (uint32_t)count);
 
 	for (i = 0; i < count; i++) {
 		if (mechs[i] > UINT64_MAX) {
@@ -1082,7 +1083,7 @@ p11_rpc_buffer_get_ulong_value (p11_buffer *buffer,
 		return false;
 
 	if (value) {
-		CK_ULONG ulong_value = val;
+		CK_ULONG ulong_value = (CK_ULONG)val;
 		memcpy (value, &ulong_value, sizeof (CK_ULONG));
 	}
 
@@ -1199,7 +1200,7 @@ p11_rpc_buffer_get_byte_array_value (p11_buffer *buffer,
 		memcpy (value, val, len);
 
 	if (value_length)
-		*value_length = len;
+		*value_length = (CK_ULONG)len;
 
 	return true;
 }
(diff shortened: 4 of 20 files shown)


vjardin:
It is failing because there is a confusion between mingw (/usr/x86_64-w64-mingw32/sys-root/mingw/bin) and cl.exe/Visual Studio; both leverage OS_WIN32. I'll check deeper how both can co-exist. If any tips, they'll be welcomed.

vjardin:
I am only using Ubuntu and not any fc30 ;  it seems I cannot set the same build environment  (Windows/wine/mingw) with a Ubuntu. I tried but they are too many gaps Ubuntu vs fc30 for such Windows emulation. I can only use windows native compilation or Ubuntu. Should someone provide me with a fc30, I can try to connect to in order to fix it.

vjardin:
let's record some chats with @ueno :

> - remove this `if` in meson.build:
>   https://github.com/p11-glue/p11-kit/blob/master/p11-kit/meson.build#L90
> - on a decent GNU/Linux distribution, do:
>   $ meson _build --cross-file build/cross_file_mingw64.txt
>   $ ninja -C _build

vjardin:
It seems it was run while some fc servers were down:

> Fedora 30 - x86_64 - Updates                     12 kB/s | 3.4 kB     00:00    
> Failed to download metadata for repo 'updates'
> Error: Failed to download metadata for repo 'updates'
> 

how to re-run it ?

I tested this compilation on my fc30 VM, it worked fine.

vjardin:
The last commit 7f2cfc1 is just a trick to kick travis once again.

vjardin:
Thanks M. Travis, "all passed" 💃 
@ueno , please, can we proceed with this log pull request ? Then I'll have some cosmetic ones that will be generic Linux/Windows. But, in order to ease any upcoming rebase, I'd prefer to get it pushed first.

vjardin:
@ueno : please, how does travis run in order to have the proper rights for the coverage tests ?

From the logs, the major issues related to the coverage is due to:

>  /usr/bin/install -c -m 644 p11-kit/p11-kit-server.socket p11-kit/p11-kit-server.service '/usr/lib/systemd/user'
> /usr/bin/install: cannot create regular file '/usr/lib/systemd/user/p11-kit-server.socket': Permission denied
> /usr/bin/install: cannot create regular file '/usr/lib/systemd/user/p11-kit-server.service': Permission denied
> make[3]: *** [Makefile:4684: install-systemduserunitDATA] Error 1
> make[2]: *** [Makefile:5574: install-am] Error 2
> make[1]: *** [Makefile:4732: install-recursive] Error 1
> make: *** [Makefile:5568: install] Error 2
> make[3]: Leaving directory '/coverage'

could be it related to some lacks for right for my account ?

thank you,

vjardin:
Please, how to proceed with this serie ?

thank you,

common/lexer.c
```diff
@@ -121,7 +121,7 @@ p11_lexer_next (p11_lexer *lexer,
 				lexer->tok_type = TOK_PEM;
 				lexer->tok.pem.begin = lexer->at;
 				lexer->tok.pem.length = end - lexer->at;
-				assert (end - lexer->at <= lexer->remaining);
+				assert (end - lexer->at <= (int)lexer->remaining);
```
ueno: perhaps `ptrdiff_t` might be more appropriate? not sure if it is available in MSVC though.

common/path.c
```diff
@@ -122,12 +123,12 @@ expand_homedir (const char *remainder)
 	if (remainder != NULL &&
 	    strncmp (remainder, ".config", 7) == 0 &&
 	    is_path_component_or_null (remainder[7])) {
-		env = getenv ("XDG_CONFIG_HOME");
+		env = secure_getenv ("XDG_CONFIG_HOME");
```
ueno: I vaguely remember that we deliberately use `getenv` instead of `secure_getenv` here, because this code is only used by p11-kit-client.so, and it can be loaded into a setuid'ed binary.

common/compat.h
```diff
@@ -170,6 +170,14 @@ typedef int pid_t;
...
 #define STDOUT_FILENO 1
 
+/* microsoft Windows uses ISO definitions, not POSIX anymore */
+#define read(fd, buf, len) _read(fd, buf, (unsigned int)(len))
+#define write(fd, buf, len) _write(fd, buf, (unsigned int)(len))
+
+#else
+/* all OS except Windows assumes that closesocket() is close() */
+
+#define closesocket(s) close(s)
```
ueno: hmm, I would like to keep the code aligned to POSIX, not to the winsock API.

p11-kit/rpc-transport.c
```diff
@@ -439,7 +440,7 @@ rpc_socket_read (rpc_socket *sock,
 			/* We ignore the options, so read into the same as buffer */
 			if (!p11_buffer_reset (buffer, sock->read_olen) ||
 			    !p11_buffer_reset (buffer, sock->read_dlen)) {
-				warn_if_reached ();
+				warn_if_reached (;);
```
ueno: I think a better fix would be remove argument (`v`) from the `warn_if_reached` definition in common/debug.h.

common/path.c
```diff
@@ -109,10 +109,12 @@ expand_homedir (const char *remainder)
 {
 	const char *env;
 
+#ifdef OS_UNIX
```
ueno: Rather than adding `#ifdef`, I would suggest adding a stub `getauxval` in common/compat.c.

p11-kit/rpc-message.c
```diff
@@ -646,7 +646,7 @@ p11_rpc_buffer_add_uint32 (p11_buffer *buffer,
 {
 	size_t offset = buffer->len;
 	if (!p11_buffer_append (buffer, 4))
-		return_val_if_reached ();
+		return_val_if_reached (;);
```
ueno: not this patch's fault, but here we should use `return_if_reached` instead.

vjardin:
Thanks for the review.  I'll provide a v2. How do you prefer me to send the updates : should I add just some few more patches in order to track the updates related to your comments (my preferred option) or do you prefer a 'repost' of this serie ?

common/compat.h
vjardin: The issue is to wrap the calls to close on a filedescriptor vs a socket descriptor. How to support it while keeping the same name ?  The only option I know is to use a name that does not conflict with the other names.

What's about closenet () that would be close () on Unix and closesocket  () on Windows ? 

https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-closesocket

ueno:
I would recommend working on a topic branch, doing `git rebase -i` and force pushing as necessary:
https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project

common/compat.h
ueno: My question is rather: would it be really possible to add a compatibility at such level? I even doubt that we can use the same function prototype of `closesocket` for both Unix and Windows, because `closesocket` takes an unsigned integer, while `close` takes a signed integer:
https://docs.microsoft.com/en-us/windows/win32/winsock/socket-data-type-2

If it is not too much burden, I would probably separate the code path completely with `#ifdef`.

vjardin: I believe I could design using,

SOCKET fh = socket( AF_INET, SOCK_STREAM, 0 );
int fd = _open_osfhandle(fh);

vjardin:
This changeset helps to have a clean cross compilation environment for mingw32, native Linux and native Windows cl.exe.

The next step will be focused on the socket handles (descriptor) vs the file descriptor ones : both are not the same on a Windows system while they are on a Linux/Unix based system.

vjardin:
Maybe it is late, but I do not understand the coveralls' report ?
  https://travis-ci.org/p11-glue/p11-kit/jobs/599398201

vjardin:
The coverage failure is coming from:

```
trust/test-edk2.c: In function ‘setup_objects’:
trust/test-edk2.c:159:42: error: passing argument 1 of ‘p11_attrs_dup’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]

  159 |   copy = p11_attrs_build (p11_attrs_dup (attrs), &id, NULL);
      |                                          ^~~~~
In file included from trust/test-edk2.c:43:

./common/attrs.h:45:21: note: expected ‘CK_ATTRIBUTE *’ {aka ‘struct _CK_ATTRIBUTE *’} but argument is of type ‘const CK_ATTRIBUTE *’ {aka ‘const struct _CK_ATTRIBUTE *’}

   45 | CK_ATTRIBUTE *      p11_attrs_dup           (CK_ATTRIBUTE *attrs);
      |                     ^~~~~~~~~~~~~

cc1: all warnings being treated as errors
```

vjardin:
@ueno since you started with the code review, is it possible to finalize the integration of this serie ? 

Sorry for pushing this serie, but right after I'd like to clean my git repo (refork it) and then to rebase some upcoming series (solutions for Windows sockets, various bug fixes). The next series will be much smaller ;)

vjardinMisc:
@ueno : sorry to ask again, but please, can you tell me if you expect something else from this serie ? If not, could you merge it so I can rebase properly the other ongoing pull requests ?

Since I am new with this p11-kit project, feel free to provide any guidance.

thank you,

vjardinMisc:
> ```
> trust/test-edk2.c: In function ‘setup_objects’:
> trust/test-edk2.c:159:42: error: passing argument 1 of ‘p11_attrs_dup’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
> [...]
> cc1: all warnings being treated as errors
> ```

all warnings are fixed, I have aligned native windows with Linux gcc.

Thanks @ueno for helping me to get the test running.

vjardin:
Please, since all the checks are OK, how to get it merged ?

vjardin:
Please, any news ?

vjardin:
Please, again, any news ?
Fluid reviews are required in order to build a proper development stream.

vjardin:
my big fingers landed on the "close button" while I was asking for an update. Sorry. Let's reopen it.

Nevertheless, what's the issue with this serie ?

vjardin:
daily check ? Please,

vjardinMisc:
Please, is any maintainer still online ? How to move forward ?

ueno:
Nagging like this wouldn't move things forward in open source community, where people are voluntarily working on the projects with different priorities from yours. I'm afraid to say, this particular patch set is very low priority to me and I will not merge it any time soon.  The reasons are:
1. There is a standard way of producing Windows binaries: cross compiling with free software tools. We don't have any problem with it as far as I know.
2. MSVC is not the best compiler conforming to the C standards, see: https://en.wikipedia.org/wiki/C99#Implementations That said, most of  the warnings you are fixing in this PR have little sense.

I'm going to close this for now.

vjardinMisc:
Regarding your item 1. : having many compilers to validate a same source code is a way to be open and to improve the quality. Cross compilers and native compilations are both welcomed in any open source project. You should not dictate just one option as a maintainer but you should rather support multiple options.

Regarding your item 2. : MSVC is widely used to compile natively many open-source projects with a high level of quality. Your comment seems to mean Microsoft = Bad.  
I guess that you are aware that many open source projects supports MSVC as a native compiler too (for instance PHP) with a high level of quality standard. 

p11-kit includes (and included) some code to support many OS including the old Windows XP. Adding more surface to the scope of p11-kit is important for the community ;  if you do not like Windows,  it is your personal problem, so your personal feelings should not interfere with the openness of a community code.

Even better MSVC environment helped to fix some bugs which are both valid for Linux and Windows. 

Moreover you spent month before providing such feedbacks. You did not nack't from the day one. Why do you hate windows now and not from the day one ?

The last but not the least: you should never close a topic just because you do not like it. You should be open to the arguments so we can reach a meaningful consensus. 

I frankly support your effort, and I am not offended by your reaction, I'll keep posting ways to improve the p11-kit including its openness unless there is a community agreement (not just your statements) that Windows MSVC is bad.

ueno:
> Regarding your item 1. : having many compilers to validate a same source code is a way to be open and to improve the quality. Cross compilers and native compilations are both welcomed in any open source project. You should not dictate just one option as a maintainer but you should rather support multiple options.

That's your expectation. It's not uncommon that free software projects only support cross compiling toolchain for producing Windows binaries. That might be true that testing with as many compilers improves the quality, but it is only the case when some of the compilers could detect unique defects that couldn't be found by other compilers. On the other hand, adding more options for compiling always increases maintenance burden.

> Regarding your item 2. : MSVC is widely used to compile natively many open-source projects with a high level of quality. Your comment seems to mean Microsoft = Bad.

Your interpretation is ridiculous. I just provided the fact that MSVC's C99 support was behind other free software alternatives.

> I guess that you are aware that many open source projects supports MSVC as a native compiler too (for instance PHP) with a high level of quality standard.

While I know several projects that support both mingw and MSVC compilation, I don't see direct relationship between the fact and the quality standard.

> p11-kit includes (and included) some code to support many OS including the old Windows XP. Adding more surface to the scope of p11-kit is important for the community ; if you do not like Windows, it is your personal problem, so your personal feelings should not interfere with the openness of a community code.

See the above explanation about maintenance cost. Also who do you mean with "the community"? I haven't seen anyone except you requesting this.

> Even better MSVC environment helped to fix some bugs which are both valid for Linux and Windows.

Citation needed.

> Moreover you spent month before providing such feedbacks. You did not nack't from the day one. Why do you hate windows now and not from the day one ?

As this is really low priority for the maintainer, a month delay is not that long in my opinion.   It is not surprising that such items take years to land, even in an actively developed project:
https://bugzilla.gnome.org/show_bug.cgi?id=720380

> The last but not the least: you should never close a topic just because you do not like it. You should be open to the arguments so we can reach a meaningful consensus.

Consensus with whom?

vjardinMisc:
> Consensus with whom?

At least both of us. I need arguments instead of statements that MSVC is not good for p11-kit. I can accept when it is properly argued. Currently, I just see a closed door.

ueno:
> > Consensus with whom?
> 
> At least both of us. I need arguments instead of statements that MSVC is not good for p11-kit. I can accept when it is properly argued. Currently, I just see a closed door.

I've already explained the rationale multiple times directly or indirectly.

From the maintainer's perspective, the pros of officially supporting MSVC are:
- It adds a better coverage of exotic compilers and might improve the code quality.
- Some people might feel convenient if they produce Windows binary with their familiar tools.

On the other hand, the cons are:
- MSVC is historically not compliant with the latest C standards.
- The additional MSVC specific code doesn't look POSIX compliant, which p11-kit is based on. That significantly complicates the code base unless the compatibility layer is carefully designed.
- It creates [a distribution restriction](https://git.savannah.gnu.org/cgit/gettext.git/tree/INSTALL.windows#n60) that may confuse users.
- Overall, it increases maintenance/support cost.

That being said, if you organize this PR to the following, chances are high: (1) changes that fix C compliance, (2) changes that fix MSVC specific C compiler issues, and (3) changes that fix Windows specific runtime issues, though (2) and (3) would still need scrutiny.

ueno:
@gvanem would it be possible to take a look, by any chance? (sorry for poking, but you are the only expert in this area I know of :-)

vjardinMisc:
Thanks both of you. Of course, I am fully open to break down or update this patch based on feedbacks.

vjardinMisc:
@ueno, @gvanem : If you are still online during the seasonal, please, any comments ? I have some times to review the patches during the end of year break.

@ueno : what's about re-opening this pull request so it remains visible for comments ?

ueno:
I spent some time to go through the changes again. It was hard as the commits are not well organized: some commits include unrelated to changes, and the other commits fix mistakes in a different commit of yours, with too brief/irrelevant commit summary. We are not editing wiki pages but developing a software library that needs to evolve. Please learn yourself a bit of basics of distributed development used these days.

Anyway, the below are the reviews; I don't inline the comment on each commit, because many commits are convoluted each other as I mentioned.

fd4d92eef5c5f1ea6ec6a6e39c3f65d4fc38e4de
For Unix compilers other than gcc and clang (e.g., xlc, SolarisStudio), you could use `AC_TYPE_SSIZE_T` instead of hard-coding `#ifndef __GNUC__`.

1886342fc3cd17b662d322d6716d2afa3bc8605c
This is a no-no. <dirent.h> is required by opendir by the standard:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/opendir.html

2c3ef70ecfbbf9819b8505f7f9291c84f9402d0d
The compiler warnings can be turned off by compiler flags, without modifying the code:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strdup-wcsdup?view=vs-2019

If you do want to modify the code, I would suggest checking
strdup in configure and meson similarly to setenv.

f58dc393402f9d29ef29856a6ad91c79867d150b
This need more elaboration at least in the commit log.  As `__argv` is explained as a Microsoft extension, whether it is defined in ucrt/stdlib.h is an implementation detail:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/argc-argv-wargv?view=vs-2019

I would suggest just ignoring the warning.

41bd6d30d98967994b4466e325f780794834ca12
No, sorry; removing `const` to avoid compilation warnings with MSVC is a not acceptable.

1bd428d480b5d216b034a8c0ba62e394ea626e67
This leaks memory as you use `_dupenv_s` which allocates memory
for the value of envvar.  Also coding style and indentation are
messed up.  Please see:
https://github.com/p11-glue/p11-kit/blob/master/CONTRIBUTING.md#coding-style

The commit summary "fix support" is too brief. You could look at:
https://chris.beams.io/posts/git-commit/ for better commit messages.

52fac1adefa0693ae91c466c1e2543fe3a347c0d
If we target only Windows 7 or later, you might want to assign the value to _WIN32_WINNT (and WINVER) instead of removing it:
https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=vs-2019

I would suggest making the Windows XP deprecation change as a separate commit: currently it spans across this commit and the previous commit with unrelated changes.

c3b928d0d59a7eb894024bcef496032f68209999
This contains 3 distinct changes:
1. Avoid including <unistd.h> on Windows
2. Define a POSIX alias for `getpid`
3. Use `secure_getenv` instead of `getenv` for `P11_KIT_DEBUG` checks

(1) is a good change, but please don't move #include to the top. For (2), the same argument applies for `_strdup`. (3) is not acceptable; this removes the ability to debug p11-kit programs when the loader is setuid'ed.

Also the commit summary "fix debug support" is too brief, and if we only include (1) and (2), it is totally irrelevant.  I would suggest using something like: "debug: Fix compiler warnings on Windows".

ec54902cbe5d879f9c79b22ba7d0742f488ded22
This is okay. But please be consistent with the commit message.

7110577bcf62d1a6006d41b3547b36283ff59d7e
This also does 2 distinct things:
1. `const` removal; not acceptable
2. Type casting the return value of `strlen` in CK_ATTRIBUTE; this is a good change

The commit summary "Fix compilation" is too brief.

49975c8ca1cf969e0169750a9c1a4628e52cf3f5
This also does 3 distinct things:
1. Remove `v` argument of `warn_if_reached`; this is a good change
2. Move `#include` block to the top; please don't do that
3. Use underscored version of the functions `_read`, `_write`, `_dup2`, `_close`: the same argument applies for `_strdup`

Also the commit message is misleading; these changes have nothing to do with ISO.

f57672f1cea9d9b83dfce5b8897545a321ae5f45
This is actually a fixup of the previous commit 1bd428d480b5d216b034a8c0ba62e394ea626e67; please squash

60cf8173d4136b82143bd9835f2c4afeba131ab4
The change seems to be good, but the commit summary is too brief.

750269cfca4800577536f171901e8f502565b008
You do not need to define a static variable; just do `return 0`.

18c698ad9ffab544bc9bdb00abd4b299487b714b
The change seems to be good, but the commit summary is too brief.

292bc629f3f9e5028a05666c3573c1a56cbda19c
This is a no-no. Please don't introduce new function name `getenv2` while the interface is the same as POSIX.
You could redefine `getenv` alias on Windows: there should be many ways to do that as we do in gnulib.

Also you mix an unrelated change (`int` -> `size_t`).

> @ueno : what's about re-opening this pull request so it remains visible for comments ?

I would suggest opening a new PR with only meaningful changes, which I can't extract from the current series cleanly.

ueno:
I tried to set up CI on AppVeyor and realized that this series is not quite complete, as there are a lot more to be adjusted (e.g., compatibility code for getopt, <dirent.h> functions, STD{IN,OUT}_FILENO, and VLA).

I did some preparatory work, but I doubt that this effort is worthwhile given that we don't need such adjustment with mingw:
https://ci.appveyor.com/project/ueno/p11-kit

PR #252, opened 2019-09-18
build: Fix undefined behavior in left shift

ueno:
Spotted by UBSan.


noloader:
Tested OK with UBsan. Ship it...

ueno:
Thanks for checking!

PR #253, opened 2019-09-25
rpc: Allow empty CK_DATE value

ueno:
Unlike other data types, CK_DATE value may be empty (and that is the default).  Treat it as a valid value and serialize/deserialize accordingly.

Reported by Vincent JARDIN in:
https://github.com/p11-glue/p11-kit/issues/244


vjardin:
LGTM
I'll perform some specific tests with my smartcards too then I'll ack't it.

vjardin:
It works fine. It is a better fix of the issue #244 than the pull request #246 

Thanks for your rework of the fix.

PR #255, opened 2019-10-07
common: add Russian PKCS#11 extensions to pkcs11x.h header

lumag:
Add values defined to support Russian GOST cryptography to pkcs11x.h
header.

Signed-off-by: Dmitry Eremin-Solenikov <dbaryshkov@gmail.com>


ueno:
Thank you. I'm not familiar with GOST, but wondering if it also requires typedefs for mechanism parameters?

lumag:
No, as far as I understand, no additional mechanisms are required.

PR #258, opened 2019-10-21
Add simple bash completion for provided commands

Jakuje:
It already bogged me that the p11-kit did not provide bash completion and I had to search help or man pages for the accurate wording of the sub-commands so I wrote one, also for the trust command.

I tried to follow the recommendation of the [bash_completion](https://github.com/scop/bash-completion) how to integrate the completion to upstream project, but I am really not sure whether it will work through the installation step nor whether there is some more magic needed for meson builds so any comments or suggestions will be welcomed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -49,6 +49,14 @@ if WITH_TRUST_MODULE
 include trust/Makefile.am
 endif
 
+if WITH_BASH_COMPLETION
+bashcompdir = $(datadir)/bash-completion/completions
+dist_bashcomp_DATA = bash-completion/p11-kit
+if WITH_TRUST_MODULE
+dist_bashcomp_DATA += bash-completion/trust
+endif
+endif
+
 SUBDIRS = . doc po
 
 ACLOCAL_AMFLAGS = -I build/m4
--- a/configure.ac
+++ b/configure.ac
@@ -138,7 +138,7 @@ if test "$os_unix" = "yes"; then
 		[AC_MSG_ERROR([could not find required gmtime_r() function])])
 
 	# Check if these are declared and/or available to link against
-	AC_CHECK_DECLS([program_invocation_short_name]), [], [], [#include <errno.h>])
+	AC_CHECK_DECLS([program_invocation_short_name], [], [], [#include <errno.h>])
 	AC_MSG_CHECKING([whether program_invocation_short_name is available])
 	AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <errno.h>]],
 	                                [[program_invocation_short_name = "test";]])],
--- a/configure.ac
+++ b/configure.ac
@@ -571,6 +571,13 @@ AS_IF([test "$with_systemd" != "no"], [
 
 AM_CONDITIONAL(WITH_SYSTEMD, [test "$with_systemd" = "yes"])
 
+# --------------------------------------------------------------------
+# bash completion
+
+PKG_CHECK_VAR([bashcompdir], [bash-completion], [with_bash_completion=yes])
+AM_CONDITIONAL(WITH_BASH_COMPLETION, [test "$with_bash_completion" = "yes"])
+
+
 AC_CONFIG_FILES([Makefile
 	doc/Makefile
 	doc/manual/Makefile
--- a/.travis/autotools/before_install.sh
+++ b/.travis/autotools/before_install.sh
@@ -5,5 +5,5 @@ export CONTAINER=$(docker run -d fedora sleep 1800)
 
 docker exec $CONTAINER dnf -y install 'dnf-command(builddep)'
 docker exec $CONTAINER dnf -y builddep p11-kit
-docker exec $CONTAINER dnf -y install gettext-devel git libtool make opensc openssl valgrind $EXTRA_PKGS
+docker exec $CONTAINER dnf -y install gettext-devel git libtool make opensc openssl valgrind bash-completion $EXTRA_PKGS
 docker exec $CONTAINER useradd user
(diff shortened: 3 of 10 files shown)


Makefile.am
```diff
@@ -49,7 +49,7 @@ if WITH_TRUST_MODULE
 include trust/Makefile.am
 endif
 
-SUBDIRS = . doc po
+SUBDIRS = . doc po bash-completion
```
ueno: We use automake in a non-recursive way if possible:
https://www.gnu.org/software/automake/manual/html_node/Alternative.html#Alternative

configure.ac
```diff
@@ -571,10 +571,28 @@ AS_IF([test "$with_systemd" != "no"], [
 
 AM_CONDITIONAL(WITH_SYSTEMD, [test "$with_systemd" = "yes"])
 
+# --------------------------------------------------------------------
+# bash completion
+
+AC_ARG_WITH(
+	[bashcompdir],
+	[AS_HELP_STRING([--with-completiondir=PATH],
+			[Directory of Bash completion @<:@detect@:>@])])
```
ueno: I doubt that this needs to be so flexible; I would probably use `AC_ARG_ENABLE` and always install the files in a fixed location.

configure.ac
```diff
@@ -571,10 +571,28 @@ AS_IF([test "$with_systemd" != "no"], [
...
+
+AS_IF([test "${bashcompdir}" != "no"], [
+	PKG_CHECK_MODULES([BASH_COMPLETION], [bash-completion >= 2.0], [],
+		[bashcompdir=no])
+
+	PKG_CHECK_VAR(bashcompdir, [bash-completion], [completionsdir], [],
+		[bashcompdir="${sysconfdir}/bash_completion.d"])
```
ueno: `bashcompdir='$(datadir)/bash-completion'`

Makefile.am
Jakuje: Thank you for the suggestion. I will try to rewrite it to include.

configure.ac
Jakuje: Can you clarify here? It is either `/usr/share/bash-completion/completions/` (as from the pkgconfig `pkg-config --variable=completionsdir bash-completion`) or the legacy location under etc (`/etc/bash_completion.d`) -- safe fallback if the pkgconfig does not work or something.

Jakuje:
> Just a comment in case it was not a conscious choice: as implemented at the moment, the completions use functions from bash-completion such as `_init_completion` and `_filedir`. That's fine with me and I don't expect these to go away, but it also means that the completion snippets here cannot be used without bash-completion (i.e. by just sourcing them in a "vanilla" no-bash-completion-enabled bash session).

Is it common to use the bash completion without the bash-completion? I would expect that out of the box, there would not be anything to load them from "your" directory.

scop:
> Is it common to use the bash completion without the bash-completion?
>
Not in my experience.

I would expect that out of the box, there would not be anything to load
> them from "your" directory.
>
Yeah, one would need to do it manually from bashrc etc.

configure.ac
scop: FWIW the intent in our (bash-completion) autotools recipe is exactly what @Jakuje described. In case you want to support bash-completion 1.x users, `/etc/bash_completion.d` is the dir to use, and as it doesn't have pkgconfig support, that's where the fallback will take the file. It'll work with more modern bash-completion as well, in case pkgconfig or bash-completion's `.pc` file are not available.

vjardin:
+1 for such features too. Thanks, it is going to make my life easier.

configure.ac
Jakuje: I wanted for users to have possibility to disable this if they really do not care for bash or specify their own directory as it is in other projects.

configure.ac
ueno: While I understand that, I still don't think we should support the legacy version, as it adds a technical debt. I've checked some other packages and they assume either `.pc` is present or `$(datadir)/bash-completion` is used:
https://github.com/latchset/clevis/blob/master/src/bash/meson.build
https://github.com/flatpak/flatpak/blob/master/Makefile.am#L138

As for the check I would suggest simply doing:
```sh
PKG_CHECK_VAR([bashcompdir], [bash-completion], [with_bash_completion=yes])
AM_CONDITIONAL(WITH_BASH_COMPLETION, [test "$with_bash_completion" = "yes"])
```
and we probably don't want the configure option.

Jakuje: I doubt that we would be installing these with old bash completion. So far, I left it as it was, but I can change it to the datadir as a fallback if you will insist.

bash-completion/Makefile.am
```diff
@@ -0,0 +1,2 @@
+bashcompdir = @bashcompdir@
```
ueno: This line should be unnecessary (automake automatically produces it).

Also, given the fact that this file will only contain the following line, I would move it to the top-level Makefile.am instead of using `include`.

configure.ac
ueno: I do not insist on the fallback now, but I rather suggest removing the fallback at all.

configure.ac
ueno: I agree with the former, but for the latter case, users can easily adjust with `make install bashcompdir=...`.

ueno:
Anyway, my point is that I would like to minimize the autotools changes for this, as we also maintain (and prefer) meson.

bash-completion/Makefile.am
Jakuje: Fixed.

configure.ac
Jakuje: Ok, makes sense. I tried to update it, so hopefully it will work. As well as the meson target (lets see CI).

Issue #257, opened 2019-10-20
C_GetSlotList() can fail with pkcs11-tool -L and no card


vjardin:
When pkcs11-tool -L (or pkcs11-dump slotinfo) is called, we get the error :
```
(p11-kit:173726) rpc_C_GetSlotList: ret: 336
Fatal: C_GetSlotList-CKR_BUFFER_TOO_SMALL
```
because of:
https://github.com/p11-glue/p11-kit/blob/6b33efaa82848bac7fd73888963a106cf4c03151/p11-kit/rpc-client.c#L871

In fact, count = 0 for the 1st call, then because of 0 slot, count = 0 for the 2nd call.
slot_list = NULL for the 1st call, then slot_slit = 0xA1pointer1A for the 2nd call.
However, for such 2nd call, since count = 0, rpc-server.c gets a slot_list = NULL over the RPC instead of some dummy pointers.
Then, the pkcs11 driver's C_GetSlotList() returns an error (CKR_BUFFER_TOO_SMALL for my case).
The intent of pkcs11-tool or pkcs11-dump for this 2nd call is to get a CKR_OK even if there is not slot however because of the RPC framework, rpc-server.c cannot guess this intent.

(fix to follow)

vjardin:
A possible hack (or fix) could be the following:
```
diff --git a/p11-kit/rpc-client.c b/p11-kit/rpc-client.c
index e202e37..c6d9242 100644
--- a/p11-kit/rpc-client.c
+++ b/p11-kit/rpc-client.c
@@ -864,8 +864,20 @@ rpc_C_GetSlotList (CK_X_FUNCTION_LIST *self,
                    CK_SLOT_ID_PTR slot_list,
                    CK_ULONG_PTR count)
 {
+       CK_SLOT_ID _slot_list;
        return_val_if_fail (count, CKR_ARGUMENTS_BAD);
 
+#if 1
+       /*
+        * XXX: some perform a call with a list, still requesting 0 element
+        * doing so will lead to a CKR_BUFFER_TOO_SMALL !!! with some pkcs11
+        * drivers.
+        */
+       if (slot_list && count && (*count == 0)) {
+               slot_list = &_slot_list; /* we cannot use slot_list */
+               *count = 1;
+       }
+#endif
        BEGIN_CALL_OR (C_GetSlotList, self, (*count = 0, CKR_OK));
                IN_BYTE (token_present);
                IN_ULONG_BUFFER (slot_list, count);

```

[...]

PR #259, opened 2019-10-21
Fix RPC when length-s are 0

vjardinMisc:
Let's add a support for cases when the buffer != NULL but the
length is 0. According to Oasis, buffer = NULL and length = 0
means a query of the length so the subsequent calls with a
buffer != NULL should fill buffer when length is long enough.
If not, according to Oasis, one should get a CKR_BUFFER_TOO_SMALL.

This current fix is for IN_ULONG_BUFFER(), same
should be applied for IN_BYTE_BUFFER() and for IN_ATTRIBUTE_BUFFER().

Fix: issue #257

### DIFF
--- a/p11-kit/rpc-client.c
+++ b/p11-kit/rpc-client.c
@@ -570,7 +570,7 @@ proto_read_sesssion_info (p11_rpc_message *msg,
 #define IN_BYTE_BUFFER(arr, len) \
 	if (len == NULL) \
 		{ _ret = CKR_ARGUMENTS_BAD; goto _cleanup; } \
-	if (!p11_rpc_message_write_byte_buffer (&_msg, arr ? *len : 0)) \
+	if (!p11_rpc_message_write_byte_buffer (&_msg, arr, len)) \
 		{ _ret = CKR_HOST_MEMORY; goto _cleanup; }
 
 #define IN_BYTE_ARRAY(arr, len) \
--- a/p11-kit/rpc-client.c
+++ b/p11-kit/rpc-client.c
@@ -582,7 +582,7 @@ proto_read_sesssion_info (p11_rpc_message *msg,
 #define IN_ULONG_BUFFER(arr, len) \
 	if (len == NULL) \
 		{ _ret = CKR_ARGUMENTS_BAD; goto _cleanup; } \
-	if (!p11_rpc_message_write_ulong_buffer (&_msg, arr ? *len : 0)) \
+	if (!p11_rpc_message_write_ulong_buffer (&_msg, arr, len)) \
 		{ _ret = CKR_HOST_MEMORY; goto _cleanup; }
 
 #define IN_ULONG_ARRAY(arr, len) \
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -372,14 +372,16 @@ p11_rpc_message_write_byte_array (p11_rpc_message *msg,
 
 bool
 p11_rpc_message_write_ulong_buffer (p11_rpc_message *msg,
-                                    CK_ULONG count)
+                                    CK_ULONG_PTR array,
+                                    CK_ULONG_PTR n_array)
 {
 	assert (msg != NULL);
 	assert (msg->output != NULL);
 
 	/* Make sure this is in the right order */
 	assert (!msg->signature || p11_rpc_message_verify_part (msg, "fu"));
-	p11_rpc_buffer_add_uint32 (msg->output, count);
+	p11_rpc_buffer_add_byte (msg->output, array ? 0 : 1);
+	p11_rpc_buffer_add_uint32 (msg->output, *n_array);
 	return !p11_buffer_failed (msg->output);
 }
 
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -336,14 +336,16 @@ p11_rpc_message_write_ulong (p11_rpc_message *msg,
 
 bool
 p11_rpc_message_write_byte_buffer (p11_rpc_message *msg,
-                                   CK_ULONG count)
+                                   CK_BYTE_PTR array,
+                                   CK_ULONG_PTR n_array)
 {
 	assert (msg != NULL);
 	assert (msg->output != NULL);
 
 	/* Make sure this is in the right order */
 	assert (!msg->signature || p11_rpc_message_verify_part (msg, "fy"));
-	p11_rpc_buffer_add_uint32 (msg->output, count);
+	p11_rpc_buffer_add_byte (msg->output, array ? 0 : 1);
+	p11_rpc_buffer_add_uint32 (msg->output, *n_array);
 	return !p11_buffer_failed (msg->output);
 }
 
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -241,14 +241,20 @@ p11_rpc_message_write_attribute_buffer (p11_rpc_message *msg,
 	/* Write the number of items */
 	p11_rpc_buffer_add_uint32 (msg->output, num);
 
+	/* Is arr not null ? */
+	p11_rpc_buffer_add_byte (msg->output, arr ? 0 : 1);
+
 	for (i = 0; i < num; ++i) {
 		attr = &(arr[i]);
 
 		/* The attribute type */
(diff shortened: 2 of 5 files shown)


p11-kit/rpc-client.c
```diff
@@ -582,7 +582,7 @@ proto_read_sesssion_info (p11_rpc_message *msg,
 #define IN_ULONG_BUFFER(arr, len) \
 	if (len == NULL) \
 		{ _ret = CKR_ARGUMENTS_BAD; goto _cleanup; } \
-	if (!p11_rpc_message_write_ulong_buffer (&_msg, arr ? *len : 0)) \
+	if (!p11_rpc_message_write_ulong_buffer (&_msg, arr ? sizeof(*arr): 0, *len)) \
```
ueno: Here the second argument of `p11_rpc_message_write_ulong_buffer` is a flag that indicates whether `arr` is NULL or not, right?

In that case, how about changing the prototype of `p11_rpc_message_write_ulong_buffer` to:
```c
bool
p11_rpc_message_write_ulong_buffer (p11_rpc_message *msg,
                                    CK_ULONG_PTR array,
                                    CK_ULONG_PTR count)
```
and directly passing the arguments?

vjardinMisc: thanks for your review.

I cannot: if I do so, the sizeof(*arr) will get lost. CK_ULONG_PTR does not provide any information about array itself.

The beauty of IN_ULONG_BUFFER() is to be within the context of the callers so we still have the size within this context.

For instance, how to carry the size of CK_SLOT_ID ?

ueno: There seems to be some misunderstanding. First, `sizeof(*arr)` doesn't return the number of elements, but the pointer size, which is always 4 or 8.  Second, I don't think we need to carry the array length in the first place.

Here is a commit I meant:
https://github.com/ueno/p11-kit/commit/33212be41344579aa517e46dda0067c2368c956a

vjardinMisc: I was assuming that the typeof *arr could be anything, so the size could be anything beside 4 or 8 bytes (it could be any kind of structure).

If we assume that *arr is not "anything" then yes, I do agree with you. Then, I agree that it makes the API clearer so we avoid to carry both a length and a count (another length).

Do you want me to update all the serie as you suggest ?

I have seen that you have updated some unit tests too: I could not find how to run these tests. Do you have a pointer/URL/documentation or some notes I could follow in order to check the tests on a Ubuntu environment ?

ueno: > I was assuming that the typeof *arr could be anything, so the size could be anything beside 4 or 8 bytes (it could be any kind of structure).

That's not how C works, I'm afraid :-)

> Do you want me to update all the serie as you suggest ?

Sure.

> I have seen that you have updated some unit tests too: I could not find how to run these tests. Do you have a pointer/URL/documentation or some notes I could follow in order to check the tests on a Ubuntu environment ?

If you use meson, do:
```sh
$ meson _build
$ ninja -C _build
$ ninja -C _build test
```
https://mesonbuild.com/Running-Meson.html#running-tests

If you use autotools, do:
```sh
$ ./configure ...
$ make
$ make check
```
https://www.gnu.org/software/automake/manual/html_node/Tests.html#Tests

You can also look at the logs on travis, which runs all tests.

vjardinMisc: thanks, done based on your advises:

> TOTAL: 759
> PASS:  759

p11-kit/rpc-server.c
```diff
@@ -175,18 +176,22 @@ proto_read_ulong_buffer (p11_rpc_message *msg,
...
 	/* The number of ulongs there's room for on the other end */
 	if (!p11_rpc_buffer_get_uint32 (msg->input, &msg->parsed, &length))
 		return PARSE_ERROR;
 
 	*n_buffer = length;
 	*buffer = NULL;
 
-	/* If set to zero, then they just want the length */
-	if (length == 0)
+	/* If set to zero, then they just want the length for a *buffer = NULL */
```
ueno: This comment looks misleading; how about:
```c
/* If buffer is NULL, the caller just want the length */
```

p11-kit/rpc-server.c
```diff
@@ -175,18 +176,22 @@ proto_read_ulong_buffer (p11_rpc_message *msg,
...
+	if (buffer_is_null)
 		return CKR_OK;
 
-	*buffer = p11_rpc_message_alloc_extra (msg, length * sizeof (CK_ULONG));
+	/* Allocate at least 1 element of buf_len even if length is 0 */
```
ueno: `buf_len` is gone. Also what is the exact reason of doing this? I'm a bit confused because the new test succeeds without this.

vjardin: because we need to allocate at least 1 slot of memory:

- buffer must not be NULL,
- length can be 0


so the underlying OASIS drivers will likely return CKR_BUFFER_TOO_SMALL which means that the lenght MUST be greater than 0 for the next call.

p11-kit/rpc-server.c
vjardinMisc: thanks for the review, fixed.

p11-kit/rpc-server.c
ueno: It doesn't answer my question, sorry. `p11_rpc_message_alloc_extra` actually allocates at least `sizeof(void *)` bytes even if the argument is 0:
https://github.com/p11-glue/p11-kit/blob/master/p11-kit/rpc-message.c#L102

Could you please show me how it can be a problem, e.g., by writing a unit test reproducing your issue?

vjardin: oh ! I did not notice that `p11_rpc_message_alloc_extra()` allocates at least sizeof(void *), so we are all good.

I'll provide an update of this serie  based on your comments for the 3 cases.

By the way, for my information, why does it always allocate extra memory `(msg->output->frealloc) (NULL, sizeof (void *) + length);` ?

p11-kit/rpc-message.c
```diff
@@ -241,14 +241,25 @@ p11_rpc_message_write_attribute_buffer (p11_rpc_message *msg,
...
+		num = 1;
+
 	for (i = 0; i < num; ++i) {
 		attr = &(arr[i]);
 
 		/* The attribute type */
 		p11_rpc_buffer_add_uint32 (msg->output, attr->type);
 
+		/* is the value buffer not NULL ? */
+		p11_rpc_buffer_add_byte (msg->output, attr->pValue ? 1 : 0);
```
ueno: The condition seems to be reverted. Here we add 1 if `attr->pValue` is *non*-NULL, but below you check the value with a variable `is_value_null`.

p11-kit/rpc-server.c
```diff
@@ -251,32 +253,52 @@ proto_read_attribute_buffer (p11_rpc_message *msg,
...
+		return PARSE_ERROR;
+
+	if (!is_attrs) {
+		*result = NULL;
+		*n_result = n_attrs;
+		return CKR_OK;
+	}
+
+	if (n_attrs == 0)
+		size_attrs = 1;
```
ueno: Why this adjustment?

vjardin: Because:
`/* we need at least one (type, value) tuple to be filled */`

I have added this comment, update will come in a minute.

p11-kit/rpc-message.c
vjardin: the code is correct, but the comment is miss-misleading. I have aligned all the *_is_null so the follow the same semantic.

update will come in a minute.

p11-kit/rpc-server.c
ueno: For the same reasoning as ulong buffer, I don't understand why we need it. If there is no attribute in the buffer, we don't need to store any element. If the caller is accessing an element even when the buffer length is 0, it is obviously a programming error.

Unless we come up with a scenario (ideally with a test case) that demonstrates the necessity, I would suggest dropping it for now.

vjardin: You are right. I spent some times in re-reading the Oasis spec today: it cannot happen for any functions that use CK_ATTRIBUTE_PTR.

I'll provide an update that remove this part.

vjardin: Please, note that I have aligned the attrs_is_null too. So all logics should be consistent.

vjardin:
Thank you for the reviews

PR #260, opened 2019-10-21
[Win32+Linux] Fix strings starting with ' or "

vjardinMisc:
For instance, when we do:
set P11_KIT_SERVER_ADDRESS="|C:\Windows\System32\OpenSSH\ssh.exe login@192.168.224.10 p11-kit remote /tmp/libpkcs11_xyz.so"

| is not the 1st character, but " is the first one.

Let's make sure we start from |

### DIFF
--- a/p11-kit/rpc-transport.c
+++ b/p11-kit/rpc-transport.c
@@ -1151,14 +1151,20 @@ p11_rpc_transport_new (p11_virtual *virt,
                        const char *name)
 {
 	p11_rpc_transport *rpc = NULL;
+	int i = 0;
 
 	return_val_if_fail (virt != NULL, NULL);
 	return_val_if_fail (remote != NULL, NULL);
 	return_val_if_fail (name != NULL, NULL);
 
+	/* skip char when the strings start with ' or " */
+	if ((remote[i] == '\'') ||
+	    (remote[i] == '"'))
+		i++;
+
 	/* This is a command we can execute */
-	if (remote[0] == '|') {
-		rpc = rpc_exec_init (remote + 1, name);
+	if (remote[i] == '|') {
+		rpc = rpc_exec_init (remote + i + 1, name);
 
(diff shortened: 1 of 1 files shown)


vjardin:
I agree for Linux but it is needed for Windows. 

Should I ifdef OS_WIN32 it ?

I wanted to avoid too many ifdef.

ueno:
> I agree for Linux but it is needed for Windows.

Is there really no way to set envvars without quotes in Windows? I found:
https://ss64.com/nt/syntax-dequote.html

vjardinMisc:
> Is there really no way to set envvars without quotes in Windows? I found:
> https://ss64.com/nt/syntax-dequote.html

I tried it, but it seems that because of starting with a pipe '|', then it keeps failing,
```
$ type test.bat
::::::::: one line Dequote example ::::::::::::
@Echo Off
Setlocal

:: OK
set P11_KIT_SERVER_ADDRESS="C:\Windows\Sy stem32\OpenSSH\ssh.exe"
CALL :dequote P11_KIT_SERVER_ADDRESS
echo %P11_KIT_SERVER_ADDRESS%

:: NOK
set P11_KIT_SERVER_ADDRESS="|C:\Windows\Sys tem32\OpenSSH\ssh.exe"
CALL :dequote P11_KIT_SERVER_ADDRESS
echo %P11_KIT_SERVER_ADDRESS%

Goto :eof

:DeQuote
for /f "delims=" %%A in ('echo %%%1%%') do set %1=%%~A
Goto :eof
```
```
$ test.bat
C:\Windows\Sy stem32\OpenSSH\ssh.exe
'C:\Windows\Sys' n’est pas reconnu en tant que commande interne
ou externe, un programme exécutable ou un fichier de commandes.
```

ueno:
Sorry, even if this is a problem for you, this change is not acceptable.

PR #261, opened 2019-10-21
[Win32] Sockets are not socket but fd

vjardinMisc:
On Windows, we spawn() to get a child process. We do not support
any Unix neither TCP socket. So, the close() ones are for the
file descriptors and not the socket handles.

This feature makes sense mainly once the serie #251 will be merged.


ueno:
No.

PR #262, opened 2019-10-23
Unbreak list matching in enable-in and disable-in

Jakuje:
The enable-in and disable-in options of the module were failing to match in case of there was a match before with substring of the searched name, for example if the running program was `p11-kit` and the `enable-in` rule looked like `p11-kit-proxy,p11-kit`, it was falsely not matched.

The attached patch is trying to fix that by implementing more robust search in list as well as attaching reproducer to the testsuite.


vjardinMisc:
LGTM

PR #265, opened 2019-10-31
trust: Support CKA_NSS_{SERVER,EMAIL}_DISTRUST_AFTER

ueno:
These new attributes are introduced in:
https://bugzilla.mozilla.org/show_bug.cgi?id=1465613

The value of the attribute can be either false (represented as a single octed "\x00"), or a UTCTime in a restricted form (i.e., "YYMMDDHHMMSSZ").  For future proof, we also support GeneralizedTime in the form "YYYYMMDDHHMMSSZ".

### DIFF
--- a/trust/builder.c
+++ b/trust/builder.c
@@ -339,6 +339,82 @@ type_der_ext (p11_builder *builder,
 	return check_der_struct (builder, "PKIX1.Extension", attr);
 }
 
+static bool
+type_false_or_time (p11_builder *builder,
+		    CK_ATTRIBUTE *attr)
+{
+	struct tm tm;
+	struct tm two;
+	char *value;
+
+	if (sizeof (CK_BBOOL) == attr->ulValueLen &&
+	    *((CK_BBOOL *)attr->pValue) == CK_FALSE)
+		return true;
+
+	value = attr->pValue;
+
+	switch (attr->ulValueLen) {
+	case 13:
+		/* UTCTime restricted by RFC 5280 4.1.2.5.1, i.e., in
+		 * the format "YYMMDDHHMMSSZ" */
+		if (value[attr->ulValueLen - 1] != 'Z')
+			return false;
+
+		tm.tm_year = atoin (value, 2);
+		if (tm.tm_year < 0)
+			return false;
+		if (tm.tm_year >= 50)
+			tm.tm_year += 1900;
+		else if (tm.tm_year >= 0)
+			tm.tm_year += 2000;
+		value += 2;
+
+		break;
+	case 15:
+		/* GeneralizedTime restricted by RFC 5280 4.1.2.5.2,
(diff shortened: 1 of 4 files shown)


vjardin:
LGTM

ueno:
This patch is not correct actually; we need to treat those attributes as of type `CK_BYTE[]`.

trust/builder.c
```diff
@@ -339,6 +339,69 @@ type_der_ext (p11_builder *builder,
 	return check_der_struct (builder, "PKIX1.Extension", attr);
 }
 
+static bool
```
ep69: I suggest to add a comment what the function does.

trust/builder.c
```diff
@@ -339,6 +339,69 @@ type_der_ext (p11_builder *builder,
...
+	value += 2;
+	tm.tm_hour = atoin (value, 2);
+	value += 2;
+	tm.tm_min = atoin (value, 2);
+	value += 2;
+	tm.tm_sec = atoin (value, 2);
+
+	if (tm.tm_mon <= 0 || tm.tm_mday <= 0 ||
+	    tm.tm_hour < 0 || tm.tm_min < 0 || tm.tm_sec < 0)
+		return false;
```
ep69: What about the other extremes? E.g., `tm_hour >= 24`

trust/builder.c
```diff
@@ -339,6 +339,69 @@ type_der_ext (p11_builder *builder,
...
+	struct tm tm;
+	char *value;
+
+	if (sizeof (CK_BBOOL) == attr->ulValueLen &&
+	    *((CK_BBOOL *)attr->pValue) == CK_FALSE)
+		return true;
+
+	value = attr->pValue;
+
+	switch (attr->ulValueLen) {
```
ep69: I would say the cases in the switch still share some logic which could be in one place only, but that is rather a nitpick, this way is probably more readable.

trust/test-builder.c
```diff
@@ -863,6 +863,71 @@ test_invalid_dates (void)
...
+
+	p11_attrs_free (extra);
+	p11_attrs_free (attrs);
+	attrs = NULL;
+
+	input[0].pValue = "\061\071\060\067\060\061\060\060\060\060\060\060\132";
+	input[0].ulValueLen = 13;
+	rv = p11_builder_build (test.builder, test.index, attrs, input, &extra);
+	assert_num_eq (CKR_OK, rv);
+
```
ep69: I suggest adding also a test for `ulValueLen == 15`

trust/builder.c
ueno: I was supposed to add a check using `mktime` below, but completely forgot about it. Good catch, fixed.

trust/test-builder.c
ueno: Good idea, added.

ueno:
Thanks for the review!

PR #269, opened 2019-12-14
If server returns already initialized, don't mark as not initialized

davedoesdev:


davedoesdev:
Server has managed module.
Client 1 RPCs to server but disconnects before calling C_Finalize
So ref count of managed module on server is 1.
Client 2 RPCs to server, gets ALREADY_INITIALIZED because ref count is 1.
rpc-client.c (in this PR) marks module on client as uninitialized
So method like GetSlots return no slots.

I'm having trouble making the test harness use managed modules + RPCs. Any example would be good.

vjardinMisc:
LGTM: I cannot reproduce it, but the logic seems to make sense.

ueno:
It's 2019; we wouldn't mark it LGTM when nobody can reproduce the issue and no test case is provided to ensure that the fix is actually correct (and to prevent future regressions).

See also: https://github.com/p11-glue/p11-kit/blob/master/CONTRIBUTING.md#test-suite

davedoesdev:
I'm happy to write a test case if someone could help me get the test infrastructure to use managed modules and RPCs in the same test. I can't seem to work out how to do that.

ueno:
I guess you could write a small program that emulates the client 1 behavior, load it through test-server.sh and then reconnect with p11tool/p11-kit-client.so?

ueno:
I'm merging this for now; let's add a test afterwards. Thank you for the PR!

PR #270, opened 2020-01-07
rpc: Add vsock transport support

dwmw2:
This adds support for AF_VSOCK as a transport for the remoting protocol.

With this, software tokens can be isolated into a microVM of their own and the communication happens over the vsock from the host/parent.

In addition to the barrier that virtualisation traditionally provides, technologies like AMD SEV and AWS Nitro Enclaves can turn this into something which approximates the security of a true hardware token.

There is no transport-level access control (yet). We could potentially filter on the client CID, and/or restrict it to root-owned ports. Using a PIN on the software token obviously does work.

I also haven't updated the documentation in p11-kit-remoting.xml. Mostly because I don't really follow it as-is. My current test is just a module file containing:
```
remote:vsock:cid=3;port=9999
```

If I were to touch the existing doc, I'd probably start by ripping out the latter half of it. NSS should be using p11-kit-proxy.so by *default* these days, shouldn't it? It shouldn't require anything special.

### DIFF
--- a/common/vsock.c
+++ b/common/vsock.c
@@ -1,0 +1,122 @@
+/*
+ * Copyright © 2020 Amazon.com, Inc. or its affiliates.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the
+ *       following disclaimer.
+ *     * Redistributions in binary form must reproduce the
+ *       above copyright notice, this list of conditions and
+ *       the following disclaimer in the documentation and/or
+ *       other materials provided with the distribution.
+ *     * The names of contributors to this software may not be
+ *       used to endorse or promote products derived from this
+ *       software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * Author: David Woodhouse <dwmw2@infradead.org>
+ */
+
+#include "config.h"
+
+#include "vsock.h"
+
+#include <limits.h>
+#include <string.h>
+#include <stdlib.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#ifdef HAVE_VSOCK
+#include <sys/socket.h>
+#include <linux/vm_sockets.h>
+#include <sys/ioctl.h>
+#endif
+
+/* This generic parsing utility doesn't actually require the
+ * vm_sockets.h header and thus doesn't require conditional
+ * compiliation... except for this one definition. */
+#ifndef VMADDR_CID_ANY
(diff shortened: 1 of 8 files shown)


common/vsock.c
```diff
@@ -0,0 +1,119 @@
...
+			*port = val;
+			port_found = true;
+		} else {
+			return false;
+		}
+
+		target = endptr;
+		if (*target == ';')
+			target++;
+		else if (*target)
```
ueno: Is this `else` clause needed?

common/vsock.c
```diff
@@ -0,0 +1,119 @@
...
+/* This generic parsing utility doesn't actually require the
+ * vm_sockets.h header and thus doesn't require conditional
+ * compiliation... except for this one definition. */
+#ifndef VMADDR_CID_ANY
+#define VMADDR_CID_ANY -1U
+#endif
+
+bool
+p11_vsock_parse_addr (const char *target,
+		      unsigned int *cid,
```
ueno: nit: it might be a bit clearer to spell out `uint32_t` for both `cid` and `port`

common/vsock.c
```diff
@@ -0,0 +1,119 @@
...
+	unsigned long val;
+	char *endptr;
+
+	/* If the caller already set the CID to 'any' then
+	 * don't require it in the given address. */
+	if (*cid == VMADDR_CID_ANY)
+		cid_found = true;
+
+	while (*target) {
+		if (strncmp(target, "cid=", 4) == 0) {
```
ueno: nit: add a space between the function name and the open paren.

configure.ac
```diff
@@ -154,6 +154,21 @@ if test "$os_unix" = "yes"; then
...
+					  .svm_family = AF_VSOCK,
+					  .svm_cid = VMADDR_CID_ANY,
+					  };
+					  (void)&sa;
+					  return 0;]])],
+	               [AC_DEFINE([HAVE_VSOCK], [1],
+	                          [Whether vsock support available])
+	                AC_MSG_RESULT([yes])],
+	               [AC_MSG_RESULT([no])])
+
```
ueno: If possible, please add this check in`meson.build` as well.

p11-kit/rpc-transport.c
```diff
@@ -1145,6 +1146,89 @@ rpc_unix_init (const char *remote,
...
+typedef struct {
+	p11_rpc_transport base;
+	struct sockaddr_vm sa;
+} rpc_vsock;
+
+static CK_RV
+rpc_vsock_connect (p11_rpc_client_vtable *vtable,
+		   void *init_reserved)
+{
+	rpc_unix *run = (rpc_unix *)vtable;
```
ueno: Why not use `rpc_vsock`?

common/vsock.c
```diff
@@ -0,0 +1,119 @@
...
+		      unsigned int *port)
+{
+	bool cid_found = false;
+	bool port_found = false;
+	unsigned long val;
+	char *endptr;
+
+	/* If the caller already set the CID to 'any' then
+	 * don't require it in the given address. */
+	if (*cid == VMADDR_CID_ANY)
```
ueno: Instead of using the inout argument, how about making this function take either "cid=xxx;port=xxx" or "port=xxx" and set *cid to VMADDR_CID_ANY if `cid=` part is omitted?

common/vsock.c
dwmw2: I used 'unsigned int' because those are what's in the sockaddr_vm.

common/vsock.c
dwmw2: The client *cannot* use VMADDR_CID_ANY. But I suppose I could make the client code check for that default value, and bail out for itself.

p11-kit/rpc-transport.c
dwmw2: Because it escaped my search/replace when I cut and pasted the unix code and modified it. Will fix.

common/vsock.c
dwmw2: Yes, because it catches the `vsock:cid=500port=500` case without the semicolon.

The acceptable things to follow a valid parameter are either a semicolon, or the end of the string.

common/vsock.c
ueno: I see, just stumbled because the `vsock` man page explicitly says those are 32-bit.

configure.ac
dwmw2: Done.

p11-kit/rpc-transport.c
dwmw2: Fixed.

common/vsock.c
dwmw2: Done.

common/vsock.c
dwmw2: Done.

ueno:
The CI failure seems to be unrelated to this PR (maybe because of cppcheck package update); let me fix them.

dwmw2:
Yeah, I was just looking at that.

```
p11-kit/proxy.c:288,warning,nullPointer,Possible null pointer dereference: mappings
```

Looks like a broken use of `realloc()` which would leak the original `py->mappings` on failure if it didn't crash first. Realloc considered harmful.

dwmw2:
No wait, it *is* checking for NULL. So it only leaks. Looks like a cpp false positive and a memory leak. Or cppcheck is reporting the memory leak in a suboptimal way?

ueno:
I think that's a false-positive: the same warning is emitted even with an explicit assert on the `mappings` and `n_mappings` values:
https://paste.centos.org/view/d1382997#L8
Since it's a warning, I guess we can ignore it. Other errors seem to be fixed with the cppcheck 1.90 package.

dwmw2:
The failure to complain about the wrong use of realloc(), on the other hand, is a false negative.

Arguably even GCC ought to complain about `x = realloc(x, …)`

ueno:
Indeed. I've filed #270 for that.

PR #271, opened 2020-01-08
build: Fix realloc usage

ueno:
As realloc() doesn't touch the original memory block, we need to use a local variable to avoid potential memory leak in failure cases.

Pointed by David Woodhouse in #270.

### DIFF
--- a/p11-kit/filter.c
+++ b/p11-kit/filter.c
@@ -96,11 +96,12 @@ static bool
 filter_add_slot (FilterData *filter, CK_SLOT_ID slot, const CK_TOKEN_INFO *token)
 {
 	if (filter->n_slots >= filter->max_slots) {
+		FilterSlot *slots;
 		filter->max_slots = filter->max_slots * 2 + 1;
-		filter->slots = realloc (filter->slots,
-					 filter->max_slots * sizeof (FilterSlot));
-		if (filter->slots == NULL)
-			return false;
+		slots = realloc (filter->slots,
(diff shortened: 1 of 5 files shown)


p11-kit/filter.c
```diff
@@ -96,11 +96,12 @@ static bool
...
 {
 	if (filter->n_slots >= filter->max_slots) {
+		FilterSlot *slots;
 		filter->max_slots = filter->max_slots * 2 + 1;
-		filter->slots = realloc (filter->slots,
-					 filter->max_slots * sizeof (FilterSlot));
-		if (filter->slots == NULL)
-			return false;
+		slots = realloc (filter->slots,
+				 filter->max_slots * sizeof (FilterSlot));
```
Jakuje: This line still mixes tabs and spaces for indentation. Not sure what is the proper coding style here though

ueno: Our coding style is a mixture of the Linux kernel coding style and the GNOME coding style: the function arguments are aligned to the left parenthesis, while the basic indentation is a hard tab. Maybe we should document it clearly in somewhere.

ueno:
Thanks for the review!

p11-kit/filter.c
Jakuje: This aligns nicely in the diff in email, but for some reason does not in this view from github (it renders for some reason tabs in 67 spaces). To fix this for good everywhere, the two tabs would have to be rewritten to spaces, but it is no big deal.

PR #279, opened 2020-02-07
rpc: Add missing null checks in attribute value serializers

ueno:
To get the length of an attribute value, the client calls `C_GetAttributeValue` with `attr->pValue` set to `NULL` and `attr->ulValueLen` set to -1.  In the RPC level, this is already indicated with a
'validity' flag, but there were a couple of places that assume `attr->pValue` is non-NULL.

This was uncovered by:
https://bugzilla.redhat.com/show_bug.cgi?id=1766340

### DIFF
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -953,9 +953,7 @@ p11_rpc_buffer_add_mechanism_type_array_value (p11_buffer *buffer,
 					       const void *value,
 					       CK_ULONG value_length)
 {
-	const CK_MECHANISM_TYPE *mechs = value;
 	size_t count = value_length / sizeof (CK_MECHANISM_TYPE);
-	size_t i;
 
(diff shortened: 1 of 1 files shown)


LorbusChris:
This would fix a long standing issue =) friendly ping

ueno:
Sorry, I realized that this alone doesn't fix [the mentioned RHBZ bug](https://bugzilla.redhat.com/show_bug.cgi?id=1766340), which may require support in Firefox side. Yet the change itself should be correct.

ueno:
Thanks for the review (merged after rewording the commit message).

dustymabe:
Hey @ueno @t8m - Is there anything tracking the remaining work mentioned in https://github.com/p11-glue/p11-kit/pull/279#issuecomment-597083379 ?

xref:  https://bugzilla.redhat.com/show_bug.cgi?id=1766340#c11

Issue #266, opened 2019-11-14
httpd fails to get signatures when using p11-kit-proxy + engine_pkcs11 and new devices are added


ansasaki:
The apache server (httpd) fails to get signatures from the PKCS#11 module when new devices are added.

It seems to be caused by a mismatch of the device mapped to the session handle obtained before a device is added and the device mapped by p11-kit-proxy to the handle after a device is added.

In a more specific scenario, if httpd is configured to use keys from a SoftHSM token and a smart card reader (using the openSC module) is connected before starting httpd, it doesn't work: It successfully starts, but fails to get signatures from the SoftHSM. This seems to be related to different list of slots/modules seen by the processes before and after forking and changing the user.

A minimal reproducer is available in: https://github.com/ansasaki/toolbox/blob/master/p11-kit-reproducer/fork-change-slot.c which is a slightly modified version of https://github.com/OpenSC/libp11/blob/master/tests/fork-change-slot.c

The reproducer requires OpenSSL, SoftHSM (from https://github.com/opendnssec/SoftHSMv2), openSC, a smart card reader, a smart card,  and engine_pkcs11 (from https://github.com/OpenSC/libp11). The steps to reproduce are:

1. Compile the reproducer:
```
$ gcc fork-change-slot.c -o fork_change_slot -lcrypto
```

2. Create a SoftHSM token and populate with keys
  - download https://github.com/ansasaki/toolbox/blob/master/setup-softhsm-rsa.sh
  - run:
```
$ source setup-softhsm-rsa.sh
```

3. Run the reproducer passing the PKCS#11 URI for the key in the softhsm and the p11-kit-proxy module (note: the pid of the child process will be printed; it will be necessary for the step 5)
```
$ fork_change_slot "pkcs11:token=softhsm;object=test;type=private?pin-value=1234" &
```

4. Connect the smart card reader with the smart card (make sure ``pcscd`` is running)

5. Send the ``SIGUSR1`` signal to the child process:
```
$ kill -USR1 <pid>
```
**Actual Results:**
```
pid <pid>: EVP_SignFinal generated errors:
140099411543040:error:8207A060:PKCS#11 module:pkcs11_private_encrypt:Key handle invalid:p11_rsa.c:116:
child <pid> exited with status 1
```

**Expected Results**
```
pid <pid1>: 256-byte signature created
pid <pid1>: Signature matched
child <pid1> exited with status 0
pid <pid2> 256-byte signature created
pid <pid2>: Signature matched
```

PR #281, opened 2020-02-13
WIP: proxy: Add option to use deterministic slot IDs

ansasaki:
The added option makes the p11-kit-proxy module to provide slot IDs
derived from the information retrieved from the devices.  This way,
devices will get a deterministic ID regardless of the order the devices are found.

Fixes:  #266

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,10 @@ matrix:
       sudo: required
       services: docker
       env: PROFILE=cppcheck SRCDIR=/srcdir BUILDDIR=/builddir EXTRA_PKGS="cppcheck"
+    - os: linux
+      sudo: required
+      services: docker
+      env: MESON_BUILD_OPTS="-Ddeterministic_slots=true" SRCDIR=/srcdir BUILDDIR=/builddir
     - os: osx
 
 language: c
--- a/configure.ac
+++ b/configure.ac
@@ -538,6 +538,24 @@ AC_SUBST(LCOV)
 AC_SUBST(GCOV)
 AC_SUBST(GENHTML)
 
+# --------------------------------------------------------------------
+# Deterministic slots
+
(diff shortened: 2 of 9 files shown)


ansasaki:
@ueno Pushed with meson option and a new build in Travis. But I couldn't confirm it worked as expected (I can't see the resulting config file). Probably I should add a specific test case.

ueno:
I'm wondering if it might make more sense to provide this option at runtime (i.e., a new configuration option in `/etc/pkcs11/pkcs11.conf`). What do you think?

ansasaki:
> I'm wondering if it might make more sense to provide this option at runtime (i.e., a new configuration option in `/etc/pkcs11/pkcs11.conf`). What do you think?

If possible, I think it would be better. I expect the configuration to be read once and keep the same behaviour even if the configuration file option is changed during execution.

One thing that is missing in my changes is to handle collisions. Probably I'd have to change the mapping data structure to be a hashmap instead of a simple array. What do you think?

Meanwhile, I'm trying to figure out if I can also change the engine to handle situation like the one that started this whole effort: make it to detect if the handle it holds is still valid.

ansasaki:
I found a bug which is probably the cause of the issue (#266):

When the proxy is assigning the slot IDs to the list, it reuses the IDs of slots previously found. The problem is that new slots can be found before old slots, and end up receiving the ID which were previously assigned to other slot. The resulting list would have duplicate IDs.

The solution is to first set all reused IDs and then assign the new IDs for new found slots. Other good thing to do is to make sure IDs assigned to new found slots were never used before. This is possible by making the IDs monotonically crescent.

This proposed changes (to use deterministic slots) was trying to solve the issue by the wrong approach and adding more complexity than necessary.

I will close this PR in favour of #282

PR #282, opened 2020-02-18
proxy: Do not assign duplicate slot IDs

### LINKED ISSUE #266: httpd fails to get signatures when using p11-kit-proxy + engine_pkcs11 and new devices are added

ansasaki:
This PR supersedes #281  

Previously, when assigning the slot IDs to the mapping, it could happen that new slots were listed by the module before slots previously seen. In this case duplicate slot IDs would be assigned to different slots.

This change makes all the reused slot IDs to be assigned before the new slot IDs. It also tries to make the slot IDs monotonically crescent.

A reproducer test case was added.

Fixes #266

### DIFF
--- a/p11-kit/proxy.c
+++ b/p11-kit/proxy.c
@@ -83,6 +83,7 @@ typedef struct {
 	p11_dict *sessions;
 	CK_FUNCTION_LIST **inited;
 	unsigned int forkid;
+	CK_SLOT_ID last_id;
 } Proxy;
 
 typedef struct _State {
@@ -113,22 +114,20 @@ map_slot_unlocked (Proxy *px,
                    CK_SLOT_ID slot,
                    Mapping *mapping)
 {
+	unsigned int i;
+
 	assert (px != NULL);
 	assert (mapping != NULL);
 
-	if (slot < MAPPING_OFFSET)
-		return CKR_SLOT_ID_INVALID;
-	slot -= MAPPING_OFFSET;
-
-	if (slot > px->n_mappings) {
-		return CKR_SLOT_ID_INVALID;
-	} else if (px->n_mappings == 0) {
-		return CKR_SLOT_ID_INVALID;
-	} else {
-		assert (px->mappings);
-		memcpy (mapping, &px->mappings[slot], sizeof (Mapping));
-		return CKR_OK;
+	for (i = 0; i < px->n_mappings; i++) {
+		assert (px->mappings != NULL);
+		if (px->mappings[i].wrap_slot == slot) {
+			memcpy (mapping, &px->mappings[i], sizeof(Mapping));
+			return CKR_OK;
+		}
 	}
+
+	return CKR_SLOT_ID_INVALID;
 }
 
 static CK_RV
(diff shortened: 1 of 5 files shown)


p11-kit/proxy.c
```diff
@@ -279,29 +277,59 @@ proxy_list_slots (Proxy *py, Mapping *mappings, unsigned int n_mappings)
...
+			/* Remember the largest ID assigned */
+			for (j = 0; j < n_mappings; ++j) {
+				if (mappings[j].wrap_slot > max_id)
+					max_id = mappings[j].wrap_slot;
+			}
+
+			/* The ID for the next new mapping will be greater than any
+			 * previously assigned ID */
+			++max_id;
+			if (max_id > MAPPING_OFFSET)
```
ueno: any reason to subtract MAPPING_OFFSET while it is added back below?

p11-kit/proxy.c
```diff
@@ -279,29 +277,59 @@ proxy_list_slots (Proxy *py, Mapping *mappings, unsigned int n_mappings)
...
 				}
-				py->mappings[py->n_mappings].funcs = funcs;
-				py->mappings[py->n_mappings].wrap_slot =
-					(n_mappings == 0 || j == n_mappings) ?
-					py->n_mappings + MAPPING_OFFSET :
-					mappings[j].wrap_slot;
-				py->mappings[py->n_mappings].real_slot = slots[i];
-				++py->n_mappings;
+				if (n_mappings == 0 || j == n_mappings)
+					is_new_mapping[i] = true;
```
ueno: can we simply maintain max_id in the Proxy structure, and increment and assign a new ID here?

ansasaki: It would be better and make things easier. I tried to make it to work without changing the structures.

I'll change this.

p11-kit/proxy.c
ansasaki: The reason is to keep the ID's sequential and not waste ``MAPPING_OFFSET`` ID's. But I agree it wasn't implemented in the best way. I think the best would be to keep the last ID assigned in the proxy structure.

p11-kit/proxy.c
ansasaki: I moved the ``max_id`` to the proxy structure as ``last_id``, which holds the last assigned slot ID. Then we can assign unique ID's and store the used value in ``proxy->last_id``

p11-kit/proxy.c
```diff
@@ -279,29 +278,46 @@ proxy_list_slots (Proxy *py, Mapping *mappings, unsigned int n_mappings)
...
 						break;
+					}
+				}
+				if (n_mappings == 0 || j == n_mappings)
+					is_new_mapping[i] = true;
+			}
+
+			/* And now add a mapping for each new slot */
+			for (i = 0; i < count; ++i) {
+				if (is_new_mapping[i]) {
```
ueno: nit: I suspect this loop can be simpler if you record slot IDs instead of boolean flags.

ansasaki: Makes sense, I'll change it.

ansasaki: Changed it to store the ID's.

Issue #289, opened 2020-04-02
test-rpc.c:187: warning: integer constant is too large for 'long' type


noloader:
I'm building GnuTLS on an old PowerMac. One of the distant dependencies is P11-Kit. P11-Kit is being built from the 0.23.19 release tarball.

`make` is showing an interesting warning:

```
p11-kit/test-rpc.c: In function 'test_uint64':
p11-kit/test-rpc.c:187: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c:203: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c: In function 'test_uint64_static':
p11-kit/test-rpc.c:220: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c: In function 'test_ulong_value':
p11-kit/test-rpc.c:398: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c:412: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c:419: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c:425: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c:431: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c:437: warning: integer constant is too large for 'long' type
p11-kit/test-rpc.c:440: warning: integer constant is too large for 'long' type
```

The offending code is:

```
p11_buffer buffer;
uint64_t val = 0xFFFFFFFFFFFFFFFF;
size_t next;
bool ret;
```
I *think* the solution is a suffix: `0xFFFFFFFFFFFFFFFFull`.

You may be able to duplicate this on a 32-bit machine. On a 32-bit machine, you are likely using a ILP32 data model.

ueno:
Thanks for the report; could you check if #290 works for you?

PR #290, opened 2020-04-13
test-rpc: Suppress compiler warning on ILP32 platforms

ueno:
On ILP32 platforms CK_ULONG is 32-bit and 64-bit values are truncated.
This patch determins the usable value from SIZEOF_UNSIGNED_LONG.
Also, for consistency, use UINTxx_MAX for other accessors.

Reported by Jeffrey Walton in:
https://github.com/p11-glue/p11-kit/issues/289


ueno:
Thanks for the review.

Issue #285, opened 2020-03-16
Tests are always built with Meson


diizzyy:
Preferably this would be optional

ueno:
What's the use-case of this? I don't oppose to have it (in particular if you come up with a patch), but I am not willing to actively add support for it.

diizzyy:
Fixed in commit 825fd4e4526589eb73a54ada18127196fb65fde3
Thanks!

PR #291, opened 2020-04-13
meson: Add options to disable NLS and building test programs

### LINKED ISSUE #284: There's no way to disable nls using Meson
It would be nice if Meson supported ```--disable-nls``` (equivalent) as when using Autotools.

ueno:
Fixes #284 and #285.


ueno:
Thanks for the review.

PR #292, opened 2020-04-13
test-token: Check if the path is actually writable

### LINKED ISSUE #288: FAIL: test-token 6 /token/not-writable
Hi Everyone,

I'm building GnuTLS on an old PowerMac. One of the distant dependencies is P11-Kit. P11-Kit is being built from the 0.23.19 release tarball. The issue is also present in 0.23.20 release tarball.

`make check` is showing one failure:

```
$ make check
...
PASS: test-token 1 /token/load
PASS: test-token 2 /token/flags
PASS: test-token 3 /token/path
PASS: test-token 4 /token/label
PASS: test-token 5 /token/slot
FAIL: test-token 6 /token/not-writable
# test-token: assertion failed (!p11_token_is_writable (token))
# test-token: in test_not_writable() at test-token.c:244
PASS: test-token 7 /token/writable-no-exist
PASS: test-token 8 /token/writable-exists
PASS: test-token 9 /token/load-found
PASS: test-token 10 /token/load-already
...
# TOTAL: 761
# PASS:  760
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
```

I have not noticed the failure on other platforms, but I may have missed it. There's a massive wall of text when the script runs so I probably miss a lot of things I should tend to.

Please let me know what else you would like to help resolve the issue.

Also see [Compiling issue on fresh linux from scratch built](https://github.com/p11-glue/p11-kit/issues/216), which appears to be the same issue on LFS.

ueno:
Instead of assuming a normal user cannot write to "/", this creates an
unwritable directly and checks if it is writable; otherwise the test
case is skipped.

Fixes #288.

### DIFF
--- a/trust/test-token.c
+++ b/trust/test-token.c
@@ -40,6 +40,10 @@
 #include <stdio.h>
 #include <string.h>
 
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <fcntl.h>
+
 #include "attrs.h"
 #include "debug.h"
 #include "parser.h"
@@ -58,6 +62,7 @@ struct {
 	p11_index *index;
 	p11_parser *parser;
 	char *directory;
+	char *unwritable;
 } test;
 
 static void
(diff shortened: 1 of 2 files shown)


trust/test-token.c
```diff
@@ -237,26 +267,30 @@ static void
...
-		p11_token_free (token);
-	}
-#endif
+	path = p11_path_build (test.unwritable, "test", NULL);
+	fd = open (path, O_WRONLY | O_CREAT, S_IRUSR | S_IWUSR);
+	free (path);
+	close (fd);
+
+	if (fd >= 0)
+		assert_skip ("cannot perform non-writable test", NULL);
```
t8m, line 279: I suppose it does not matter that you leave the test file created here and that it will be removed in the teardown_temp() call?

ueno: Yes, the test directory is recursively removed.

ueno:
Thanks for the review.

PR #301, opened 2020-06-12
anchor: Exit with non-zero code, if any error occurs

### LINKED ISSUE #300: trust anchor --remove doesn't print any output or change exit code on failure
p11-kit-0.23.20:
```
# trust anchor --remove 'pkcs11:id=%AD%6C%AA%94%60%9C%ED%E4%FF%FA%3E%0A%74%2B%63%03%F7%B6%59%BF;type=cert'
```
Exit code is zero and no error is printed. When run with verbose it prints that the certificate was not removed.

p11-kit-0.23.2:
```
# trust anchor --remove 'pkcs11:id=%AD%6C%AA%94%60%9C%ED%E4%FF%FA%3E%0A%74%2B%63%03%F7%B6%59%BF;type=cert'
p11-kit: couldn't remove read-only certificate
```
Exit code is again zero but an error message is printed.

I think the expected behavior on failure would be to print a message and set an error code.

ueno:
This makes `trust anchor --store` or `trust anchor --remove` exit with non-zero if any error happens.

Also it makes the tools print error messages by default.

Fixes #300.

### DIFF
--- a/common/tool.c
+++ b/common/tool.c
@@ -182,14 +182,14 @@ command_usage (const p11_tool_command *commands)
 static void
 verbose_arg (void)
 {
-	putenv ("P11_KIT_DEBUG=tool");
+	setenv ("P11_KIT_DEBUG", "tool", 0);
 	p11_message_loud ();
 }
 
 static void
 quiet_arg (void)
 {
-	putenv ("P11_KIT_DEBUG=");
+	unsetenv ("P11_KIT_DEBUG");
 	p11_message_quiet ();
 }
 
--- a/common/tool.c
+++ b/common/tool.c
@@ -184,7 +184,6 @@ verbose_arg (void)
 {
 	putenv ("P11_KIT_DEBUG=tool");
 	p11_message_loud ();
-	p11_debug_init ();
 }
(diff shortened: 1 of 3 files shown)


common/tool.c
```diff
@@ -182,17 +182,15 @@ command_usage (const p11_tool_command *commands)
 static void
 verbose_arg (void)
 {
-	putenv ("P11_KIT_DEBUG=tool");
```
nmav, line 185: I wonder why putenv was there in the first place. The only reason I can think of is that putenv works in windows.

ueno:
Thanks for the review.

FTR, the command now produces the following output:
```console
$ trust anchor --remove 'pkcs11:id=%AD%6C%AA%94%60%9C%ED%E4%FF%FA%3E%0A%74%2B%63%03%F7%B6%59%BF;type=cert'
p11-kit: couldn't remove read-only certificate
p11-kit: 1 error while processing
zsh: exit 1     trust anchor --remove
```
It's not perfect as it says "p11-kit" rather than "trust", but anyway, should be better than before.

Issue #303, opened 2020-07-25
Use of calloc(0,x) in p11-kit/modules.c assumes a GNU-compatible libc


twwlogin:
In p11-kit/modules.c for 0.23.20, you have:
```
static CK_SESSION_HANDLE *
managed_steal_sessions_inlock (p11_dict *sessions,
                        bool matching_slot_id,
                        CK_SLOT_ID slot_id,
                        int *count)
{
        CK_SESSION_HANDLE *stolen;
        CK_SESSION_HANDLE *key;
        CK_SLOT_ID *value;
        p11_dictiter iter;
        int at, i;

        assert (sessions != NULL);
        assert (count != NULL);

        stolen = calloc (p11_dict_size (sessions), sizeof (CK_SESSION_HANDLE));
        return_val_if_fail (stolen != NULL, NULL);
```
According to the malloc(3) on Linux:
> calloc()  allocates memory for an array of nmemb elements of size bytes
> each and returns a pointer to the allocated memory.  The memory is  set
> to  zero.  If nmemb or size is 0, then calloc() returns either NULL, or
> a unique pointer value that can later be successfully passed to free().

According to the malloc man page on AIX:
> Upon successful completion, the calloc subroutine returns a pointer to
> the allocated, zero-initialized array. If the size requested is 0, the
> calloc subroutine returns NULL in normal circumstances. However, if the
> program was compiled with the macro _LINUX_SOURCE_COMPAT defined, the
> calloc subroutine returns a valid pointer to a space of size 0.
>
> If the request cannot be satisfied for any reason, the calloc
> subroutine returns NULL.

According to malloc(3C) on Solaris:
> If there is no available memory, malloc(), realloc(), memalign(),  val-
> loc(),  and  calloc()  return  a null pointer. When realloc() is called
> with size > 0 and returns NULL, the block pointed to  by  ptr  is  left
> intact.  If  size,  nelem,  or  elsize is 0, either a null pointer or a
> unique pointer that can be passed to free() is returned.

I compiled a small test program on Solaris 10, 11, HP-UX 11.23, 31 and calloc(0,x) returned a non-NULL value. On AIX, a NULL value was returned except when _LINUX_SOURCE_COMPAT was defined (so, by default, AIX does not have a GNU-compatible libc). Aa workaround is to use _LINUX_SOURCE_COMPAT. However, is that what we really want to do, or should the code be rewritten to not assume a GNU-compatible libc?

twwlogin:
[...]

PR #304, opened 2020-07-25
p11-kit/modules.c: Avoid passing 0 as first argument to calloc().

twwlogin:
Solves issue #303 for AIX.


twwlogin:
I looked at other uses of calloc() in p11-kit/ and trust/ but I think they all pass a non-zero int as the first value.

ueno:
The CI failures look unrelated to this change; let's fix those after merging it.

PR #305, opened 2020-08-01
Avoid using setenv() for MinGW users.

Carrot031:
In MinGW (Minimalist GNU for Windows), unsetenv() is undefined...
I propose to use setenv() instead.

Signed-off-by: Kai Takahashi <www.carrotsoft@gmail.com>


ueno:
Looks good to me, thanks!

PR #307, opened 2020-08-02
common: Get program name based on executable path if possible

### LINKED ISSUE #263: Improve handling of progname
Trying to enable a p11-kit module in chromium, I ended up with the following logs:

> (p11-kit:24673) is_module_enabled_unlocked: disabled module 'opensc-local' running in 'chromium-browser/chromium-browser --type=utility --field-trial-handle=17202589914978909127,12880597771655614828,131072 --lang=en-US --service-sandbox-type=network --service-request-channel-token=327873692212110115 --shared-files=v8_context_snapshot_data:1'

This prevents any sensible matching against the `enable-in` and `disable-in` rules defined in the module files.

As promptly noticed by @ueno chromium does ugly things with the argv[0] and it requires special handling, as in the example of mesa:

https://github.com/mesa3d/mesa/commit/759b94038987bb983398cd4b1d2cb1c8f79817a9

Using something similar inside of the `_p11_get_progname_unlocked` is probably a way to go and it should not be too much of work.

ueno:
Some programs (e.g., Chromium) pack command line arguments into `argv[0]`. Check if it is the case by reading `/proc/self/exe` and extract the program name.

Logic borrowed from:
<https://github.com/mesa3d/mesa/commit/759b94038987bb983398cd4b1d2cb1c8f79817a9>.

Fixes #263.

### DIFF
--- a/common/compat.c
+++ b/common/compat.c
@@ -112,7 +112,36 @@ getprogname (void)
 	if (p != NULL)
 		name = p + 1;
 #elif defined (HAVE_PROGRAM_INVOCATION_SHORT_NAME)
+#ifdef __linux__
+	name = program_invocation_name;
+	assert (name);
+	if (*name == '/') {
+		/*
+		 * Some programs pack command line arguments into argv[0].
+		 * Check if it is the case by reading /proc/self/exe and extract
+		 * the program name.
+		 *
+		 * Logic borrowed from:
+		 * <https://github.com/mesa3d/mesa/commit/759b94038987bb983398cd4b1d2cb1c8f79817a9>.
+		 */
+		static char *buf;
+
+		if (!buf)
+			buf = realpath ("/proc/self/exe", NULL);
+
+		if (buf && strncmp (buf, name, strlen (buf)) == 0)
+			/* Use the executable path if the prefix matches. */
+			name = strrchr (buf, '/') + 1;
+		else
+			/* Otherwise fall back to
+			 * program_invocation_short_name. */
+			name = program_invocation_short_name;
+	} else {
+		name = program_invocation_short_name;
+	}
+#else
 	name = program_invocation_short_name;
+#endif
 #elif defined (HAVE___PROGNAME)
 	name = __progname;
 #else
--- a/common/frob-getprogname.c
+++ b/common/frob-getprogname.c
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2020 Red Hat Inc.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the
+ *       following disclaimer.
+ *     * Redistributions in binary form must reproduce the
+ *       above copyright notice, this list of conditions and
+ *       the following disclaimer in the documentation and/or
+ *       other materials provided with the distribution.
+ *     * The names of contributors to this software may not be
+ *       used to endorse or promote products derived from this
+ *       software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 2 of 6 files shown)


common/compat.c
```diff
@@ -112,7 +112,28 @@ getprogname (void)
...
+		 * Logic borrowed from:
+		 * <https://github.com/mesa3d/mesa/commit/759b94038987bb983398cd4b1d2cb1c8f79817a9>.
+		 */
+		static char *buf;
+
+		if (!buf)
+			buf = realpath ("/proc/self/exe", NULL);
+
+		if (buf && strncmp (buf, name, strlen (buf)) == 0)
+			    name = strrchr (buf, '/');
```
t184256, line 134: off-by-one here

common/compat.c
```diff
@@ -112,7 +112,28 @@ getprogname (void)
...
+		 *
+		 * Logic borrowed from:
+		 * <https://github.com/mesa3d/mesa/commit/759b94038987bb983398cd4b1d2cb1c8f79817a9>.
+		 */
+		static char *buf;
+
+		if (!buf)
+			buf = realpath ("/proc/self/exe", NULL);
+
+		if (buf && strncmp (buf, name, strlen (buf)) == 0)
```
t184256, line 133: This won't work if name isn't a realpath itself, e.g. a relative path or a symlinked path. Is there a need to cover those cases as well?

common/compat.c
```diff
@@ -112,7 +112,28 @@ getprogname (void)
 	if (p != NULL)
 		name = p + 1;
 #elif defined (HAVE_PROGRAM_INVOCATION_SHORT_NAME)
+	const char *p;
 	name = program_invocation_short_name;
+#ifdef __linux__
+	p = strrchr (name ? name : "", '/');
```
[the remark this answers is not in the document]
t184256: On a second thought, why `strrchr`? If it's only effectively triggered on absolute paths, can just the first character be checked?

common/compat.c
```diff
@@ -112,7 +112,28 @@ getprogname (void)
 	if (p != NULL)
 		name = p + 1;
 #elif defined (HAVE_PROGRAM_INVOCATION_SHORT_NAME)
```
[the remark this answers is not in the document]
ueno: The other branches are basically for other OSes, which likely don't have `/proc/self/exe` I suspect. For the meantime I'd skip the effect to Linux only.

common/compat.c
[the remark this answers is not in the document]
ueno: That's a very good point; fixed to check only the first character.

common/compat.c
ueno: Fixed.

common/compat.c
ueno: Indeed; however, it would bring too much complexity in this exceptional path IMO. Let's skip it until we find an actual use-case that require such checks.

common/frob-getprogname.c
```diff
@@ -0,0 +1,124 @@
...
+			if (waitpid (pid, &status, 0) < 0) {
+				perror ("waitpid");
+				exit (EXIT_FAILURE);
+			}
+
+			assert (!WIFSIGNALED (status));
+			assert (WIFEXITED (status));
+			assert (WEXITSTATUS (status) == 0);
+
+			p = strchr (buffer, '\n');
```
t184256, line 109: I'm not sure what exactly are you anticipating here, so that a string would be missing a newline, but I suspect it has a chance of also not being null-terminated in this case.

I propose folding this check into a `strncmp ("frob-getprogname\n", buffer, sizeof(buffer))` or, alternatively, null-terminating the buffer after reading it (and maybe not bothering with `\n` altogether).

common/frob-getprogname.c
```diff
@@ -0,0 +1,124 @@
...
+			close (pfds[1]);
+			while (1) {
+				nread = read (pfds[0], buffer + offset, sizeof(buffer) - offset);
+				if (nread < 0) {
+					perror ("read");
+					exit (EXIT_FAILURE);
+				}
+				if (nread == 0)
+					break;
+				if (sizeof(buffer) - offset < nread) {
```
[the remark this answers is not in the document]
ueno: Indeed, let's remove the check then.

common/frob-getprogname.c
ueno: I replaced `strchr` with `memchr` so it also handles the non null-terminated case.

ueno:
Thank you for the detailed review!

PR #315, opened 2020-08-19
meson: generate and install trust-extract-compat

xry111:
Now meson.build files don't mention trust-extract-compat.  This PR makes meson generate and install it, just like autotools.


trust/meson.build
```diff
@@ -105,6 +105,11 @@ executable('trust',
                           libp11_tool_dep] + libffi_deps + dlopen_deps + libtasn1_deps,
            install: true)
 
+configure_file(input: 'trust-extract-compat.in',
+               output: 'trust-extract-compat',
+               configuration: configuration_data(),
```
ueno, line 110: It's not a problem in your patch, but it looks like `trust-extract-compat.in` doesn't contain any substitution keywords (i.e., @...@). Maybe we can just rename `trust-extract-compat.in` to `trust-extract-compat` in the repo?

ueno (commented):
Good catch, thank you!

trust/meson.build
xry111: I agree.  New version force-pushed.

ueno:
To fix the CI failure, I guess you would need to add `EXTRA_DIST += trust/trust-extract-compat` in `trust/Makefile.am`.

ueno:
Thanks!

PR #318, opened 2020-08-27
unix-peer: replace incorrect include1

neheb:
Fixes musl warning:

warning: #warning redirecting incorrect #include <sys/errno.h> to <errno.h> [-Wcpp]


ueno:
Thank you!

PR #319, opened 2020-08-31
p11_test_copy_setgid: Skip setgid tests on nosuid filesystems

andersk:
Some build environments build packages from a build directory mounted nosuid, causing `p11_test_copy_setgid` to fail.  Skip the test in this case.

### DIFF
--- a/common/test.c
+++ b/common/test.c
@@ -53,6 +53,7 @@
 
 #ifdef OS_UNIX
 #include <sys/stat.h>
+#include <sys/statvfs.h>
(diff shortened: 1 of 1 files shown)


ueno (approved):
Thank you, looks good to me.

Although it might be sensible to have a build-time check on `statvfs`, it seems to be available ubiquitously so let's use it unconditionally:
https://www.gnu.org/software/gnulib/manual/html_node/statvfs.html#statvfs

PR #321, opened 2020-09-06
common: Check for a NULL locale before freeing it

tavianator:
If newlocale() fails, (locale_t) 0 ends up being passed to freelocale(),
resulting in a segfault when the library is unloaded.


ueno (approved):
Thank you; looks good to me.

Issue #322, opened 2020-09-21
meson regression: bashcompdir


jhgit:
One could specify bashcompdir using autoconf-based configure.  Not so with meson configure.  I would guess it's simple, but  don't know meson well enough to have a suggested patch yet.

jhgit:
Here's a patch.  Please review.

    --- meson_options.txt.orig  2020-05-17 17:46:43 UTC
    +++ meson_options.txt
    @@ -42,6 +42,10 @@ option('systemd', type : 'feature',
            value : 'auto',
            description : 'Use systemd socket activation')
     
    +option('bashcompdir', type : 'string',
    +       value : '',
    +       description : 'Override default location for bash completion files')
    +
     option('bash_completion', type : 'feature',
            value : 'auto',
            description : 'Install bash completion files')

    --- bash-completion/meson.build.orig        2020-05-17 17:46:43 UTC
    +++ bash-completion/meson.build
    @@ -1,11 +1,15 @@
    -bashcomp = dependency('bash-completion', required: get_option('bash_completion'))
    -
    -if bashcomp.found()
    -  bashcompdir = bashcomp.get_pkgconfig_variable('completionsdir')
    +bashcompdir = get_option('bashcompdir')
    +if bashcompdir == ''
    +  bashcomp = dependency('bash-completion', required: get_option('bash_completion'))
    +  if bashcomp.found()
    +    bashcompdir = bashcomp.get_pkgconfig_variable('completionsdir')
    +  else
    +    warning('Will not install bash completion due to missing dependencies!')
    +  endif
    +endif
    +if bashcompdir != ''
       install_data('p11-kit', install_dir: bashcompdir)
       if with_trust_module
         install_data('trust', install_dir: bashcompdir)
       endif
    -else
    -  warning('Will not install bash completion due to missing dependencies!')
     endif

ueno:
Would you like to file a regular PR so I can review with inline comments?

jhgit:
You mean PR = Pull Request?  (PR can also mean "problem report" which is what this is).   I will do a pull request.

ueno:
Yes, thanks!

ueno:
Closed with #323.

PR #323, opened 2020-09-23
meson: Allow override of default bashcompdir.

jhgit:
Fixes meson regression (issue #322).  Pass -Dbashcompdir=/xxx to meson.


ueno (approved):
Looks good to me, thanks!

PR #325, opened 2020-10-05
proxy: C_CloseAllSessions: Make sure that calloc args are non-zero

ueno:
This prevents efence warning if either of the calloc arguments is zero.  While it is is safe on glibc systems, POSIX says the behavior is implementation-defined.
    
Reported by Paul Wouters.


letoams:
This patch resolves my issue. Thanks!

ueno:
Thank you for the review.

PR #326, opened 2020-10-22
common: Fix infloop in p11_path_build

ueno:
If `p11_path_build` is called with 2 or more arguments and the non-first argument is an empty string (""), it previously fell into an infloop.

Reported by Karel Srot.

### DIFF
--- a/common/path.c
+++ b/common/path.c
@@ -241,8 +241,10 @@ p11_path_build (const char *path,
 			num--;
 
 		if (at != 0) {
-			if (num == 0)
+			if (num == 0) {
+				path = va_arg (va, const char *);
 				continue;
+			}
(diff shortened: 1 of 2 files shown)


t184256:
It does the job and avoids the infinite loop on encountering `""`, but it didn't feature stripping leading separators on the follow-up path component. I tried addressing separator duplication by moving the said stripping to the top of the loop, here's the result: https://github.com/t184256/p11-kit/commit/d9a9061e7422b10d079d6262ef7cb2ddc86420c9

ueno:
Cool, would you mind filing a new PR superseding this then? :-)

ueno:
Superseded by #327.

PR #327, opened 2020-10-27
Follow-up improvements to p11_path_build

t184256:
This PR builds upon @ueno's #326 that fixes handling of empty string components in `p11_path_build`.
It incorporates the following improvements on top of that:
* components following empty strings get their leading path separators stripped
* multiple leading path separators of the first component are now stripped as well
* `(".../", "", "/...")` situation also won't result in duplicate separators

Note that:
* `is_path_component` has been extracted from `is_path_component_or_null`, hope it's OK
* I haven't actually executed the windows tests

### DIFF
--- a/common/path.c
+++ b/common/path.c
@@ -119,7 +119,7 @@ expand_homedir (const char *remainder)
 		return NULL;
 	}
 
-	while (remainder[0] && is_path_component_or_null (remainder[0]))
+	while (is_path_component (remainder[0]))
 		remainder++;
 	if (remainder[0] == '\0')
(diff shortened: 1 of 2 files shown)


t184256:
d0ca2c6 force-push: make formatting more consistent (add spaces after function names)

ueno (approved):
Thanks; looks good to me, though we might eventually want to reword "path component" as "path separator".

t184256:
@ueno, good catch, this somehow evaded me! I've renamed them to is_path_separator* in ec3ef1c.

PR #328, opened 2020-10-30
WIP: github actions: Initial CI setup

ansasaki:
This is an attempt to evaluate the usefulness of the github actions as a CI alternative.

After various attempts to write the configuration file in a "smarter" format, the conclusion for now is that the github actions lacks features that are required for better usability. For instance, the configuration file does not support reuse of steps from a job in another. Other missing feature is the possibility to share a docker image built in a job to another job.

It is necessary to invest more effort to understand if there are better ways to solve the issues found.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,244 @@
+name: test
+on: [push]
+env:
+  BUILDDIR: "builddir"
+  SRCDIR: "srcdir"
+  COVDIR: "coverage"
+  INSTALLDIR: "installdir"
+  PACKAGES: "gettext-devel git libtool make opensc openssl valgrind meson ninja-build bash-completion"
+jobs:
+  autotools:
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:latest
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,244 @@
...
+  autotools:
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:latest
+    env:
+      BUILD_OPTS: --enable-coverage
+      COVERAGE: yes
+      EXTRA_PKGS: lcov python-pip
+    steps:
+      # Install packages
```
ueno, line 19: I guess we can probably make use of the [GH container registry](https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/about-github-container-registry) to avoid installing packages on each CI run. If you are willing to try that, I can give you access to the project; let me know.

ansasaki: That would be nice to try!

ueno:
Let's merge it as-is and see if we can improve by utilizing container registry.

PR #329, opened 2020-11-06
anchor: Prefer persistent format when storing anchor

ueno:
When a new certificate is stored with "trust anchor --store" from a .p11-kit file, the command treated it as a PEM file, while it should
preserve extra fields in the file.

### DIFF
--- a/trust/test-extract.sh
+++ b/trust/test-extract.sh
@@ -8,7 +8,7 @@ teardown()
 		if [ -d $x ]; then
 			rmdir $x
 		elif [ -f $x ]; then
-			rm $x
+			rm -f $x
 		fi
 	done
 	TD=""
@@ -51,6 +51,7 @@ setup()
 	CERT_1_CN=test_$(dd if=/dev/urandom count=40 bs=1 status=none | base64 | tr -d '+/=')
 	CERT_2_CN=test_$(dd if=/dev/urandom count=40 bs=1 status=none | base64 | tr -d '+/=')
 	CERT_3_CN=test_$(dd if=/dev/urandom count=40 bs=1 status=none | base64 | tr -d '+/=')
+	CERT_4_CN=test_$(dd if=/dev/urandom count=40 bs=1 status=none | base64 | tr -d '+/=')
 
 	# Generate relevant certificates
 	openssl_quiet req -x509 -newkey rsa:512 -keyout /dev/null -days 3 -nodes \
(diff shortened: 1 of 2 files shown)


ueno:
cc @t184256

trust/test-extract.sh
```diff
@@ -59,12 +60,23 @@ setup()
...
+	openssl_quiet req -x509 -newkey rsa:512 -keyout /dev/null -days 3 -nodes \
+		-out cert_4.pem -subj /CN=$CERT_4_CN
 
-	TD="cert_1.pem cert_2.pem cert_3.pem $TD"
+	TD="cert_1.pem cert_2.pem cert_3.pem cert_4.pem $TD"
 
 	mkdir -p $SOURCE_1/anchors
 	cp cert_1.pem $SOURCE_1/anchors/
 
+	cat > cert_4.p11-kit <<EOF
```
t184256, line 71: This reads out-of-order, I'd prefer it moved after the cert1-3 code.

trust/test-extract.sh
```diff
@@ -97,4 +110,11 @@ test_blocklist()
 	assert_not_contains blocklist-test.pem $CERT_3_CN
 }
 
-run test_extract test_blocklist
+test_persist()
+{
+	trust dump --filter "pkcs11:object=$CERT_4_CN" > cert_4.p11-kit.dump
+	TD="cert_4.p11-kit.dump $TD"
+	assert_contains cert_4.p11-kit.dump nss-server-distrust-after:
```
t184256, line 117: I believe this could be improved by checking for actual values as well and not just the presence of the field.

t184256:
Is 0714723 related to the change in question? If yes, how?

ueno:
https://github.com/p11-glue/p11-kit/commit/07147234b2bab874bb0bea2ce794e872bb601a2c is for a CI failure in mingw64 cross build. Since the other one (autotools) is still failing, I'll split it to a separate PR.

t184256:
I can't confirm that the test is being called. I've placed an assertion that should've failed and tried both `ninja make` or `make check`, both passed. Could you please confirm that it is invoked?

ueno:
It's need to be called from the autotools build, something like:
```console
./autogen.sh --without-systemd --without-bash-completion --with-trust-paths=$PWD/ca-trust-source --prefix=$PWD/prefix
make
make install
mkdir ca-trust-store
make installcheck
```
Unfortunately, the autotools CI job is currently failing because of the reasons I haven't investigated yet.

ueno:
> Unfortunately, the autotools CI job is currently failing because of the reasons I haven't investigated yet.

This should be fixed with #330 (the reason was that docker on the base system is not compatible with Fedora 33).

t184256:
Can confirm that the test now passes in CI, and that it fails with `test-extract.sh: nss-server-distrust-after is not preserved` if the change is reverted. Would be ideal if meson builds also triggered this, but that's probably a separate issue.

ueno:
Thanks for checking!

> Would be ideal if meson builds also triggered this, but that's probably a separate issue.

This is a bit difficult because we would need to tweak the paths embedded in the library.

Issue #331, opened 2020-11-16
Cannot build with -Dman=true


t184256:
At ce66cf00b6b207c1d452af23cb062ca0adf57dac:

```
$ meson _build -Dman=true
...
$ ninja -C _build
...
FAILED: doc/manual/pkcs11.conf.5
/usr/bin/xsltproc --nonet --stringparam man.output.quietly 1 --stringparam funcsynopsis.style ansi --stringparam man.th.extra1.suppress 1 --stringparam man.authors.section.enabled 0 --stringparam man.copyright.section.enabled 0 --path /home/asosedki/code/p11-kit-clean/_build/doc/manual -o doc/manual/pkcs11.conf.5 http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl ../doc/manual/pkcs11.conf.xml
warning: failed to load external entity "../doc/manual/sysdir.xml"
../doc/manual/pkcs11.conf.xml:255: parser error : Failure to process entity sysdir
        <literal>&sysdir;/pkcs11.conf</literal> and the user global
                         ^
../doc/manual/pkcs11.conf.xml:255: parser error : Entity 'sysdir' not defined
        <literal>&sysdir;/pkcs11.conf</literal> and the user global
                         ^
warning: failed to load external entity "../doc/manual/userdir.xml"
../doc/manual/pkcs11.conf.xml:256: parser error : Failure to process entity userdir
        configuration file is in <literal>&userdir;/pkcs11.conf</literal> in the
                                                   ^
../doc/manual/pkcs11.conf.xml:256: parser error : Entity 'userdir' not defined
        configuration file is in <literal>&userdir;/pkcs11.conf</literal> in the
                                                   ^
warning: failed to load external entity "../doc/manual/sysdir.xml"
../doc/manual/pkcs11.conf.xml:260: parser error : Failure to process entity sysdir
        <literal>&sysdir;/modules</literal> directory, with one configuration
                         ^
../doc/manual/pkcs11.conf.xml:260: parser error : Entity 'sysdir' not defined
        <literal>&sysdir;/modules</literal> directory, with one configuration
                         ^
warning: failed to load external entity "../doc/manual/userdir.xml"
../doc/manual/pkcs11.conf.xml:261: parser error : Failure to process entity userdir
        file per module. In addition the <literal>&userdir;/modules</literal> directory
                                                           ^
../doc/manual/pkcs11.conf.xml:261: parser error : Entity 'userdir' not defined
[...]

PR #333, opened 2020-11-16
meson: Allow building manpages without gtk-doc

heftig:
Fixes https://github.com/p11-glue/p11-kit/issues/331


ueno (approved):
Nice; thank you!

Issue #332, opened 2020-11-16
meson: doc/manual/pkcs11.conf.5 has problems with userdir/sysdir expansions


t184256:
`doc/manual/pkcs11.conf.5` has `<literal>&sysdir;/pkcs11.conf</literal>` expanded to `etc/pkcs11 /pkcs11\&.conf` (notice the space in the variable expansion)

I see that `doc/manual/Makefile.am` uses `echo -n` to generate the intermediate files with the values, this might be a relevant nuance lost during the transition to meson.

ueno:
Indeed; we'll probably have to use the [`command` keyword](https://mesonbuild.com/Reference-manual.html#configure_file), or rewrite it with a custom target to avoid the trailing newline.

PR #334, opened 2020-11-20
meson: Use custom_target for generating external XML entities

ueno:
configure_file() adds a trailing newline, and thus the resulting
document contains unwanted spaces between the resolved entities and
the following text.  Instead, use custom_target() along with 'echo -n'
to generate the XML entity files.

Fixes: #332


t184256 (approved):
Seems to fix #332 for me.

ueno:
Thanks for the review.

PR #335, opened 2020-11-27
meson: expand ternary operator in function call for compatibility

ueno:
While the minimum version requirement of meson is [0.49](https://github.com/p11-glue/p11-kit/blob/master/meson.build#L3), the current meson.build causes the parser to crash, because of the use of the ternary operator inside a function call:
https://github.com/mesonbuild/meson/issues/5003

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -30,7 +30,11 @@ conf.set('PACKAGE_MAJOR', major_version)
 conf.set('PACKAGE_MINOR', minor_version)
 
 host_system = host_machine.system()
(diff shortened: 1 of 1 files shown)


ueno:
cc @t184256 @Jakuje

t184256 (approved):
Oh, thanks, now I understand what you were leading to in that other discussion.

I can confirm that this patch fixes the Linux build using meson 0.49.2.

ueno:
Thanks for checking.

Jakuje:
Thanks for looking into that!

PR #342, opened 2020-12-12
p11-kit/lists.c: Add stdint.h to fix compilation

diizzyy:
Add stdint.h otherwise compilation fails on FreeBSD 13-CURRENT with "use of undeclared identifier 'SIZE_MAX'"

Signed-off-by: Daniel Engberg <daniel.engberg.lists@pyret.net>


ueno:
Thank you!

PR #352, opened 2021-02-06
Follow-up changes after GitHub Actions migration

ueno:


ueno:
I have no idea about the coverage decrease; perhaps it calculates the line coverage and the branch coverage in (somehow) combined way? In any case, we didn't touch the source code, I'd say it's not a blocker.

Issue #357, opened 2021-03-05
Weird situation around internationalization


ericonr:
I am switching p11-kit to use the meson build system instead of autotools, which changed a few things. First, autotools seems to never want to install the files from `po/`, so the package doesn't ship any of them. Switching to meson was enough for them to be properly generated and installed.

However, I can't figure out how you are *using* the files...

This block, from `messages.c`, is never enabled (`ENABLE_NLS` and `PACKAGE_NAME` aren't defined):

```
#ifdef ENABLE_NLS
#include <libintl.h>
#define _(x) dgettext(PACKAGE_NAME, x)
#else
#define _(x) x
#endif
```

but even if it were, the necessary `textdomain` call to initialize gettext doesn't exist anywhere. If you can guarantee the toolchain supports a constructor, you can put it there, though I'm not sure how it behaves if called before `setlocale`, which also isn't called anywhere here.

For now, at least, it seems like enabling `nls` only generates files which aren't used by any part of the resulting package :/

ueno:
Thank you for pointing that out. I've filed #358 to fix this as well as to make more strings translatable (it might be too much though).

PR #358, opened 2021-03-05
Fix gettext integration with tools

ueno:
Fixes: #357

### DIFF
--- a/p11-kit/p11-kit.c
+++ b/p11-kit/p11-kit.c
@@ -49,6 +49,14 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#ifdef ENABLE_NLS
+#include <libintl.h>
+#define _(x) dgettext(PACKAGE_NAME, x)
+#else
+#define _(x) (x)
+#endif
+#define N_(x) (x)
+
 #include "tool.h"
 
 int       p11_kit_list_modules    (int argc,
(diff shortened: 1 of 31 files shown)


p11-kit/p11-kit.c
```diff
@@ -49,6 +49,14 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#ifdef ENABLE_NLS
+#include <libintl.h>
+#define _(x) dgettext(PACKAGE_NAME, x)
+#else
+#define _(x) (x)
+#endif
+#define N_(x) (x)
```
tomato42, line 58: what's this for?

trust/trust.c
```diff
@@ -54,12 +54,14 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#define N_(x) (x)
```
tomato42, line 57: here too, why `N_`?

tomato42 (commented):
I don't get the use of `N_`...

p11-kit/p11-kit.c
ueno: It's a marker for the xgettext program to extract the translatable strings into POT files, but not replace the text at run time:
https://developer.gnome.org/glib/stable/glib-I18N.html#N-:CAPS

tomato42 (approved):
LGTM

ueno:
Thanks for the review.

PR #360, opened 2021-03-10
packit: fedora renamed master branch to rawhide

TomasTomecek:
\+
```
packit: clean up

* synced_files: that's default so we can drop it
* create-archive is not used for propose_downstream
```

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,7 +1,4 @@
 specfile_path: p11-kit.spec
-synced_files:
-  - p11-kit.spec
-  - .packit.yaml
 upstream_package_name: p11-kit
 downstream_package_name: p11-kit
 
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -8,9 +8,9 @@ downstream_package_name: p11-kit
 # Use only populated spec files and upstream sources.
 actions:
   post-upstream-clone:
-    - wget https://src.fedoraproject.org/rpms/p11-kit/raw/master/f/p11-kit.spec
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -13,9 +10,6 @@ actions:
     - wget https://src.fedoraproject.org/rpms/p11-kit/raw/rawhide/f/trust-extract-compat
   get-current-version:
     - "git describe --abbrev=0"
-  create-archive:
-    - "wget https://github.com/p11-glue/p11-kit/releases/download/$PACKIT_PROJECT_VERSION/p11-kit-$PACKIT_PROJECT_VERSION.tar.xz"
-    - "wget https://github.com/p11-glue/p11-kit/releases/download/$PACKIT_PROJECT_VERSION/p11-kit-$PACKIT_PROJECT_VERSION.tar.xz.sig"
```
ueno, line 18: > * create-archive is not used for propose_downstream

The current workflow is that, after uploading a release tarball along with the GPG signature, I open an issue to ask packit-service to propose-update, where I thought `create-archive` is used. Is it not the case?

TomasTomecek: oh, I feel so silly now, you are correct, part of the downstream process is to upload tarballs

sorry about this, will drop that hunk

TomasTomecek: The workflow you described sounds fine. Btw, cockpit team is planning to use something similar - we're discussing with them what would be the most convenient way to do when an archive is created outside packit: https://github.com/packit/packit-service/issues/1017

ueno (approved):
Thank you!

PR #366, opened 2021-05-31
Support recursive attributes indicated with CKF_ARRAY_ATTRIBUTE

ueno:
This adds support for getting and setting attributes with array type, e.g., `CKA_WRAP_TEMPLATE`.


ueno:
The commits are already present in git; only missing part is RPC support.

PR #368, opened 2021-06-03
proxy: Support copying array of attributes

ueno:
Previously C_SetAttributeValue implementation in the proxy didn't copy array of attributes recursively and caused the subsequent C_GetAttributeValue receive the same memory address for the leaf attributes. This changes the behavior so the copying is always deep copy.

### DIFF
--- a/common/attrs.c
+++ b/common/attrs.c
@@ -82,8 +82,9 @@ p11_attrs_free (void *attrs)
 	if (!attrs)
 		return;
 
-	for (i = 0; !p11_attrs_terminator (ats + i); i++)
-		free (ats[i].pValue);
+	for (i = 0; !p11_attrs_terminator (ats + i); i++) {
+		p11_attr_clear (&ats[i]);
+	}
 	free (ats);
 }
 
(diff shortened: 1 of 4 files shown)


ueno:
@ansasaki would you perhaps like to review this?

ansasaki:
> @ansasaki would you perhaps like to review this?

Yes, I can review

ueno:
Thank you!

common/attrs.c
```diff
@@ -143,18 +144,17 @@ attrs_build (CK_ATTRIBUTE *attrs,
...
 		}
 
-		memcpy (attr, add, sizeof (CK_ATTRIBUTE));
-		if (!take_values && attr->pValue != NULL) {
-			if (attr->ulValueLen == 0)
-				attr->pValue = malloc (1);
-			else
-				attr->pValue = memdup (attr->pValue, attr->ulValueLen);
-			return_val_if_fail (attr->pValue != NULL, NULL);
+		if (take_values) {
```
ansasaki, line 152: When is the ``take_values`` flag passed? Is it possible to have the flag set and the attribute to be added be an array?

ueno: @ansasaki sorry for the long delay.

There are a couple of occasions where `take_values` is true: called from the internal API `p11_attrs_take` and `p11_attrs_merge`. Both of them take ownership of the operands, so even if they are an array, we don't need recursive copy; just moving the container should be sufficient.

onihboni:
autogen.sh

PR #372, opened 2021-06-11
meson: optionalise glib's development files for gtk_doc

sgn:
In p11-kit, only installation prefix of glib is required to build gtk-doc.

In some distro, those development files are splitted into separated
package.  Pulling those development files in will pull its development
dependencies, too.

Let's give our users an option to build gtk-doc without development
files of glib.


ueno (approved):
Makes sense; thank you!

PR #377, opened 2021-06-16
Meson: p11_system_config_modules instead of p11_package_config_modules

concatime:
A trivial typo.

### DIFF
--- a/p11-kit/meson.build
+++ b/p11-kit/meson.build
@@ -307,7 +307,7 @@ p11_kit_pc_variables = [
   'p11_module_configs=@0@'.format(prefix / p11_package_config_modules),
   'p11_module_path=@0@'.format(prefix / p11_module_path),
   'proxy_module=@0@/p11-kit-proxy.so'.format(prefix / libdir),
-  'p11_system_config_modules=@0@'.format(prefix / p11_package_config_modules)
+  'p11_system_config_modules=@0@'.format(prefix / p11_system_config_modules)
 ]
 
 if trust_paths != ''


ueno (approved):
Thank you.

p11-kit/meson.build
```diff
@@ -307,7 +307,7 @@ p11_kit_pc_variables = [
   'p11_module_configs=@0@'.format(prefix / p11_package_config_modules),
   'p11_module_path=@0@'.format(prefix / p11_module_path),
   'proxy_module=@0@/p11-kit-proxy.so'.format(prefix / libdir),
-  'p11_system_config_modules=@0@'.format(prefix / p11_package_config_modules)
+  'p11_system_config_modules=@0@'.format(prefix / p11_system_config_modules)
```
debarshiray, line 310: Was this really intended?  It changes the value of `p11_system_config_modules` from what it is in the Autotools build  Specifically:
```
-p11_system_config_modules=/usr/share/p11-kit/modules
+p11_system_config_modules=/etc/pkcs11/modules
```

I noticed this because of a breakage in Fedora's gnome-keyring [build](https://bugzilla.redhat.com/show_bug.cgi?id=2045447).  We were using the [previous value](https://src.fedoraproject.org/rpms/gnome-keyring/c/64189d5d8bac5015) with the Autotools build and then with Meson.

Feel free to ignore me if this was really intended, and please accept my apologies in that case.

debarshiray: Or is it possible that gnome-keyring is using the wrong path?  I am asking because `/etc` seems like something for system administrators, while `/usr/share` is better suited for files fully controlled by the operating system.

ueno: @debarshiray thanks for looking into it! Yeah indeed this is a backward incompatible change, though I would suggest that applications switch to using `p11_module_configs` instead of `p11_system_config_modules`. I'll revert this change for now with a similar comment we have in Autotools .pc [template](https://github.com/p11-glue/p11-kit/blob/master/p11-kit/p11-kit-1.pc.in#L14).

PR #383, opened 2021-08-24
Link libp11-kit so that it cannot unload

manu0401:
libp11-kit installs a pthread_atfork() callback to count forks. If the
library gets unloaded, a stale pointer is called on the next fork(),
leading to a crash.

This fix adds -z nodelete linker flag so that libp11-kit cannot be
unloaded. Unfortunately, that only fixes the problem for GNU ld.
More additionnal flags will be required for other linkers.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -44,6 +44,19 @@ LT_INIT([dlopen disable-static])
 dnl Check if -Wl,--version-script is supported by the linker
 gl_LD_VERSION_SCRIPT
 
+# --------------------------------------------------------------------
+# Check for -z nodelete link flag
+
+SAVE_LDFLAGS="$LDFLAGS"
+LDFLAGS="$LDFLAGS -Wl,-z -Wl,nodelete"
+AC_MSG_CHECKING([whether linker understands -z nodelete])
+AC_LINK_IFELSE([AC_LANG_PROGRAM([], [])],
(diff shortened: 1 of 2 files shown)


ueno (approved):
Thank you; that makes sense. Would you also like to add the same logic to:
https://github.com/p11-glue/p11-kit/blob/0.24.0/p11-kit/meson.build#L55

manu0401:
On Tue, Aug 24, 2021 at 10:36:33AM -0700, Daiki Ueno wrote:
> @ueno approved this pull request.
> 
> Thank you; that makes sense. Would you also like to add the same logic to:
> https://github.com/p11-glue/p11-kit/blob/0.24.0/p11-kit/meson.build#L55

I am less familiar with that than with autotools. I would be glad if 
you beat me at doing it.

-- 
Emmanuel Dreyfus
***@***.***

ueno:
OK, let me try; thank you for the patch!

ueno:
#384 is the follow-up for meson.

manu0401:
On Tue, Aug 24, 2021 at 10:36:33AM -0700, Daiki Ueno wrote:
> Thank you; that makes sense.

Thinking at it afterwards, I wonder if it deserves a comment
in configure.ac noting why we want it to get the undelete flag.

-- 
Emmanuel Dreyfus
***@***.***

ueno:
Sure, I've amended the comment in #384.

PR #389, opened 2021-10-16
build: Suppress cppcheck false-positive on array bounds

ueno:
Cppcheck 2.6 reports the following errors, when loops on arrays depend
on a sentinel:
```console
trust/builder.c:1103,error,arrayIndexOutOfBounds,Array 'schema->attrs[32]' accessed at index 9998, which is out of bounds.
trust/builder.c:1136,error,arrayIndexOutOfBounds,Array 'schema->attrs[32]' accessed at index 9998, which is out of bounds.
trust/builder.c:1301,error,arrayIndexOutOfBounds,Array 'ku_attribute_map[8]' accessed at index 9998, which is out of bounds.
trust/builder.c:1385,error,arrayIndexOutOfBounds,Array 'eku_attribute_map[9]' accessed at index 9998, which is out of bounds.
```
This adds an extra check to those loop conditions.  While it is
redundant, this should be harmless as it is not in the performance
critical path.

### DIFF
--- a/trust/builder.c
+++ b/trust/builder.c
@@ -62,6 +62,8 @@
 #define _(x) (x)
 #endif
 
+#define ELEMS(x) (sizeof (x) / sizeof (x[0]))
+
 struct _p11_builder {
 	p11_asn1_cache *asn1_cache;
 	p11_dict *asn1_defs;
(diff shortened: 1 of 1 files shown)


ansasaki (approved):
LGTM!

Issue #413, opened 2022-04-12
"trust dump --help" displays "usage: trust list"


nora-pxh:
"trust dump --help" displays usage of trust list incorrectly.
The usage string of p11_tool_desc usages in trust/dump.c uses “trust list”.
	p11_tool_desc usages[] = {
		{ 0, "usage: trust list --filter=<what>" },

winner658:
The usage string of p11_tool_desc usages in trust/dump.c  should  uses "trust dump" 
p11_tool_desc usages[] = {
{ 0, "usage: trust dump--filter=" },

winner658:
i think it was copy from trust/list.c, and author forget changed

ueno:
Yes, pull-request fixing it would be welcome :-)

ueno:
Should be fixed in #414.

PR #414, opened 2022-04-27
fix  "trust dump --help" displays "usage: trust list" #413

winner658:
The usage string of p11_tool_desc usages in trust/dump.c should uses "trust dump"
"trust dump" p11_tool_desc usages[] = { { 0, "usage: trust dump--filter=" }


ueno (approved):
Thank you!

PR #415, opened 2022-05-02
Add support for some IBM specific mechanisms supported by OpenCryptoki for the RPC protocol

ifranzki:
This patch series adds support for a couple of IBM specific mechanisms (and attributes), as well as some standard mechanisms that use mechanism parameters that were not supported so far for the RPC protocol.
All these mechanisms are supported by OpenCryptoki (https://github.com/opencryptoki/opencryptoki) and we would like to make sure that those mechanisms can be used with p11-kit's remote-PKCS#11 feature.

I have tested this by running OpenCryptoki's testcases (https://github.com/opencryptoki/opencryptoki/tree/master/testcases - slightly modified) over p11-kit's remote PKCS#11 interface. Some modifications in OpenCryptoki's testcases were required to let it use the p11-kit's p11-kit-client.so instead of OpenCryptkis's shared library, as well as to avoid the use of things not supported by p11-kit: Attribute-array attributes (CKA_WRAP_TEMPLATE etc), and PKCS#11 version 3.0 interfaces.


common/pkcs11.h
```diff
@@ -982,7 +982,6 @@ struct ck_aes_ctr_params {
 struct ck_gcm_params {
   unsigned char *iv_ptr;
   unsigned long iv_len;
-  unsigned long iv_bits;
```
ueno, line 985: Seems like there was a confusion in the PKCS#11 standard, where the doc says nothing about this field, while the official C [header](https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/include/pkcs11-v2.40/pkcs11t.h) has this (see the [comment](https://hg.mozilla.org/projects/nss/file/tip/lib/util/pkcs11t.h#l1877) in NSS for the rationale).

Given the definition is copied into SoftHSMv2 and used already, I'd suggest not touching it.

ueno (changes requested):

common/pkcs11.h
ifranzki: OK, I guess I then also have to adapt Opencryptoki to maybe support both structure versions.....

ifranzki: changed, see updated commit [Add support for CKM_AES_GCM](https://github.com/p11-glue/p11-kit/pull/415/commits/949daa1289ff484cad134edc20c98b12ee2395bc).

p11-kit/rpc-message.c
```diff
@@ -1660,6 +1660,72 @@ p11_rpc_buffer_get_aes_ctr_mechanism_value (p11_buffer *buffer,
...
+		return false;
+	if (!p11_rpc_buffer_get_uint64 (buffer, offset, &val))
+		return false;
+
+	if (value) {
+		CK_GCM_PARAMS params;
+
+		params.pIv = (void *) data1;
+		params.ulIvLen = len1;
+		params.ulIvBits = len1 * 8;
```
ifranzki, line 1715: Maybe I should better transport the `ulIvBits` field separately and not derive it from `ulIvLen`? I guess it does not necessarily correspond directly to `ulIvLen * 8`, but can be less.

What do you think?

common/pkcs11x.h
```diff
@@ -181,6 +181,65 @@ typedef CK_ULONG                        CK_TRUST;
 
 #endif /* CRYPTOKI_RU_TEAM_TC26_VENDOR_DEFINED */
 
+/* Define this if you want the IBM specific symbols */
+#define CRYPTOKI_IBM_VENDOR_DEFINED 1
+#ifdef CRYPTOKI_IBM_VENDOR_DEFINED
+
+#ifndef CKK_VENDOR_DEFINED
```
ueno, line 188: These are also defined in `common/pkcs11.h`.

p11-kit/rpc-message.c
ueno: According to PKCS#11 3.0, `ulIvLen` seems to take precedence over `ulIvBits` so I guess we could keep it.
By the way I wonder if we can handle AES-GCM properly with the PKCS#11 2.40 interface alone, which lacks message based [functions](https://www.oasis-open.org/committees/download.php/52895/PKCS11MessageBasedEncryption20140502.pdf).

ueno (approved):
Sorry for the delay. The changes look good to me, though I have a few concerns: whether those new mechanisms work without recursive attributes (#366) and need to wait for PKCS#11 3.0 support. Perhaps you could point me to the actual usage in openCryptoki?

 All in all, the RPC changes seem reasonable and we could merge them first.

common/pkcs11x.h
ifranzki: right. do you want me to remove the CKK_VENDOR_DEFINED definitions then?

ifranzki:
> By the way I wonder if we can handle AES-GCM properly with the PKCS#11 2.40 interface alone, which lacks message based [functions](https://www.oasis-open.org/committees/download.php/52895/PKCS11MessageBasedEncryption20140502.pdf).

Yes this works with just regular C_Encrypt/C_Decrypt functions as of PKCS v2.40.

ifranzki:
> whether those new mechanisms work without recursive attributes (https://github.com/p11-glue/p11-kit/pull/366) and need to wait for PKCS#11 3.0 support. Perhaps you could point me to the actual usage in openCryptoki?

None if these changes are dependent to recursive attributes. 

Commit https://github.com/p11-glue/p11-kit/pull/415/commits/a8f205c7945d19d30c768572fd65310f565198a8 is the only one that is related to attribute-array attributes, but is not really required for this PR.

Opencryptoki does support attribute-array attributes in general, and some of the opencryptoki testcases use them, so I thought I could get those tests to work without your PR, but it seems not. But this does not matter for these changes.

common/pkcs11x.h
ueno: If there is no use of this file (pkcs11x.h) alone, then that would be nice.

ifranzki: will check and force push if that works.

p11-kit/rpc-message.c
ifranzki: So do we want to keep it as is, or do we want to transport ulIvBits separately? 

This would be something like the following:

```
 p11-kit/rpc-message.c | 15 +++++++++------
 1 file changed, 9 insertions(+), 6 deletions(-)

diff --git a/p11-kit/rpc-message.c b/p11-kit/rpc-message.c
index 97b5a30..f785b79 100644
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -1675,8 +1675,8 @@ p11_rpc_buffer_add_aes_gcm_mechanism_value (p11_buffer *buffer,
 
 	memcpy (&params, value, value_length);
 
-	/* Check if params.ulTagBits can be converted to uint64_t. */
-	if (params.ulTagBits > UINT64_MAX) {
+	/* Check if params.ulTagBits/ulIvBits can be converted to uint64_t. */
+	if (params.ulTagBits > UINT64_MAX || params.ulIvBits > UINT64_MAX) {
 		p11_buffer_fail (buffer);
 		return;
 	}
@@ -1684,6 +1684,7 @@ p11_rpc_buffer_add_aes_gcm_mechanism_value (p11_buffer *buffer,
 	p11_rpc_buffer_add_byte_array (buffer,
 				       (unsigned char *)params.pIv,
 				       params.ulIvLen);
+	p11_rpc_buffer_add_uint64 (buffer, params.ulIvBits);
 	p11_rpc_buffer_add_byte_array (buffer,
 				       (unsigned char *)params.pAAD,
 				       params.ulAADLen);
@@ -1696,15 +1697,17 @@ p11_rpc_buffer_get_aes_gcm_mechanism_value (p11_buffer *buffer,
 					    void *value,
 					    CK_ULONG *value_length)
 {
-	uint64_t val;
+	uint64_t val1, val2;
 	const unsigned char *data1, *data2;
 	size_t len1, len2;
 
 	if (!p11_rpc_buffer_get_byte_array (buffer, offset, &data1, &len1))
 		return false;
+	if (!p11_rpc_buffer_get_uint64 (buffer, offset, &val1))
+		return false;
 	if (!p11_rpc_buffer_get_byte_array (buffer, offset, &data2, &len2))
 		return false;
-	if (!p11_rpc_buffer_get_uint64 (buffer, offset, &val))
+	if (!p11_rpc_buffer_get_uint64 (buffer, offset, &val2))
 		return false;
 
 	if (value) {
@@ -1712,10 +1715,10 @@ p11_rpc_buffer_get_aes_gcm_mechanism_value (p11_buffer *buffer,
 
 		params.pIv = (void *) data1;
 		params.ulIvLen = len1;
-		params.ulIvBits = len1 * 8;
+		params.ulIvBits = val1;
 		params.pAAD = (void *) data2;
 		params.ulAADLen = len2;
-		params.ulTagBits = val;
+		params.ulTagBits = val2;
 
 		memcpy (value, &params, sizeof (CK_GCM_PARAMS));
 	}
```

Let me know what you prefer.

ueno: Thanks; I like this change.

common/pkcs11x.h
ifranzki: changed

p11-kit/rpc-message.c
ifranzki: changed.

ueno:
Thank you!

PR #417, opened 2022-05-31
build: Use EXEEXT and SHLEXT

ueno:
For better compatibility with Windows, append EXEEXT when spawning binary executables as well as replace ".so" with SHLEXT determined at build time.


ansasaki (approved):
LGTM!

PR #422, opened 2022-07-16
export SHELL in shell script

doug-catsone:
specify the correct `SHELL` in the environment, overriding any existing and potentially csh-ish value, so `p11-kit-server` writes the env files correctly


ZoltanFridrich:
This issue has been resolved by #438 
Closing this MR

PR #423, opened 2022-07-26
fix remote on Windows

### LINKED ISSUE #376: CK_ULONG attribute RPC issue between Windows p11-kit-server and Linux p11-kit-client.
Since p11-kit 0.23.22, I'm encountering a problem with RPC between p11-kit-server running on Windows and p11-kit-client running on Linux. I'm seeing `C_GetAttributeValue` PKCS#11 API calls fail in p11-kit-client with `CKR_DEVICE_ERROR`. I've bisected this (on the 0.23 branch) and found that the issue started from https://github.com/p11-glue/p11-kit/commit/2617f3ef888e103324a28811886b99ed0a56346d. The issue occurs when a `CK_ULONG` attribute is sent from the server. In `p11_rpc_buffer_get_attribute()` on Linux p11-kit-client, the `ulValueLen` field of the `CK_ATTRIBUTE` instance is 8, but `length` (holding the value read from a previous `p11_rpc_buffer_get_uint32()` call) is 4. This causes the function to return false as `if (decode_length > length)` evaluates to true (which in turn leads to the `CKR_DEVICE_ERROR`).

I've looked further into the discrepancy between these length values. In `p11_rpc_buffer_add_attribute()`, the attribute length specified by `attr->ulValueLen` is first written to the RPC buffer, followed by the attribute value. For a `CK_ULONG` attribute on Windows p11-kit-server, `attr->ulValueLen` is 4 (`sizeof(unsigned long int)` on this platform). However, `p11_rpc_buffer_add_ulong_value()` always encodes values using 8 bytes, with `p11_rpc_buffer_add_uint64()`. In Linux p11-kit-client, `p11_rpc_buffer_get_ulong_value()` always reads an 8 byte value, and sets the `value_length` output parameter to `sizeof(CK_ULONG`) (which is 8 on this platform). This then yields the different length values observed in `p11_rpc_buffer_get_attribute()`, which now fails the validation check added by https://github.com/p11-glue/p11-kit/commit/2617f3ef888e103324a28811886b99ed0a56346d.

Perhaps the validation in `p11_rpc_buffer_get_attribute()` added by https://github.com/p11-glue/p11-kit/commit/2617f3ef888e103324a28811886b99ed0a56346d needs an exemption for ulong types (whose size may differ on server and client platforms)?

bhuztez:
sometimes msvcrt set stdin and stdout to text mode on start, we have to set them to binary mode. also fix #376.

### DIFF
--- a/p11-kit/remote.c
+++ b/p11-kit/remote.c
@@ -44,6 +44,7 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -113,6 +114,11 @@ main (int argc,
 		p11_message (_("the 'remote' tool is not meant to be run from a terminal"));
 		return 2;
 	}
+
+#ifdef OS_WIN32
+        _setmode(_fileno(stdin), _O_BINARY);
+        _setmode(_fileno(stdout), _O_BINARY);
+#endif
 
 	if (strncmp (argv[0], "pkcs11:", 7) == 0) {
 		CK_FUNCTION_LIST *module = NULL;
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -1059,6 +1059,27 @@ p11_rpc_buffer_add_byte_array_value (p11_buffer *buffer,
 }
 
 void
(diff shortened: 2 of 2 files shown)


ZoltanFridrich:
We have decided to split this MR into two parts. The p11-kit/remote.c change can be merged so I will file a separate MR to merge it. The rest needs to be a bit refined.

PR #424, opened 2022-07-27
pkcs11-gnu: Enable testing with <p11-kit/pkcs11x.h>

ueno:
This ensures that programs using <p11-kit/pkcs11x.h> can be compiled
with CRYPTOKI_GNU.  The previous coverage was partial as pkcs11-gnu.c
didn't include "pkcs11x.h" and Meson didn't supply -DCRYPTOKI_GNU=1.


ansasaki (approved):
LGTM!

Issue #425, opened 2022-07-27
Compat definitions of O_BINARY and O_CLOEXEC broken


kadler:
https://github.com/p11-glue/p11-kit/blob/master/common/compat.h#L78-L84

compat.h does not include fcntl.h, so the definitions of O_BINARY and O_CLOEXEC will never be defined already unless one of the includers of compat.h includes fcntl.h first (which AFAICT is never the case). This will cause O_BINARY and O_CLOEXEC to always be defined to 0 and then when fcntl.h is included, they will either get overridden by the system headers or left as 0 instead of using the system values if the system values are guarded by an ifndef (not sure any systems do this though).

Seems like compat.h should include fcntl.h before it checks whether these are defined.

ueno:
Thank you for the report. It should be fixed with #428.

PR #428, opened 2022-07-29
compat: Fix detection of O_BINARY and O_CLOEXEC

ueno:
Previously those macros were defined as 0 unless <fcntl.h> was
included before "compat.h".

Reported and fix suggested by Kevin Adler in:
https://github.com/p11-glue/p11-kit/issues/425

Fixes: #425 

### DIFF
--- a/common/compat.h
+++ b/common/compat.h
@@ -37,8 +37,9 @@
 
 #include "config.h"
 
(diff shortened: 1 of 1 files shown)


ansasaki (approved):
LGTM!

Issue #435, opened 2022-10-17
Partially untranslatable string


urbalazs:
Here: https://github.com/p11-glue/p11-kit/blob/master/p11-kit/modules.c#L2171-L2172

```
p11_message (_("%s: module failed to initialize%s: %s"),
    name, critical ? "" : ", skipping", p11_kit_strerror (rv));
```

The string `, skipping` is untranslatable. Please use this message like:

```
critical ? p11_message (_("%s: module failed to initialize: %s"), name, p11_kit_strerror (rv)) : p11_message (_("%s: module failed to initialize, skipping: %s"), name, p11_kit_strerror (rv));
```

PR #436, opened 2022-10-18
modules: Make logging message more translatable

ueno:
Reported and fix suggested by Balázs Úr in:
https://github.com/p11-glue/p11-kit/issues/435

Fixes: #435


ueno:
As this change is only about translatable strings, I'm merging it without approval.

Issue #437, opened 2022-10-18
Environment in p11-kit/test-server.sh


A-Bld:
Hello,

In the p11-kit/test-server.sh, call 

"$abs_top_builddir"/p11-kit/p11-kit-server <etc...>

is dependent on $SHELL environment variable.
I use tcsh which prompts the generated start.env and stop.env to be created with tcsh syntax and leads to an error on execution of the latters as the script is (basic) sh .

Fix is simple: just add 

unset SHELL

for example (in my case novel line 20) before the lines

unset P11_KIT_SERVER_ADDRESS
unset P11_KIT_SERVER_PID

(I was installing version 0.24.1_1 , error occured at make check stage of p11-kit built).

   Best, A-Bld

A-Bld:
Re-Hello,
Seems to be a bit general in the various scripts used in the make check of the different modules p11-kit...

Best,

  A-Bld.

ueno:
Thank you for the report. I think it's indeed an issue in p11-kit server; #438 should fix it.

PR #438, opened 2022-10-19
server: Check SHELL only when neither --sh nor --csh is specified

ueno:
Previously C shell syntax is used for printing environment variable setup even if --sh is given when SHELL is csh.

Reported in:
https://github.com/p11-glue/p11-kit/issues/437

Fixes: #437


ZoltanFridrich (approved):
Makes sense. LGTM

PR #446, opened 2022-11-28
Add command to print merged configuration

### LINKED ISSUE #395: Add ability to dump the merged configuration
p11-kit provides 3 levels of configuration: system wide, per-module, and per-user. At run time these are merged into a single configuration (dictionary). It would be nice if there is a way to output the final configuration in a human (and/or machine) readable form.

ZoltanFridrich:
Add an option to dump merged configuration to standard output. Currently its a draft which uses test configs as an input.

Closes #395 

Signed-off-by: Zoltan Fridrich <zfridric@redhat.com>

### DIFF
--- a/p11-kit/Makefile.am
+++ b/p11-kit/Makefile.am
@@ -237,13 +237,17 @@ EXTRA_DIST += \
 
 bin_PROGRAMS += p11-kit/p11-kit
 
+p11_kit_p11_kit_CFLAGS = $(COMMON_CFLAGS)
+
 p11_kit_p11_kit_SOURCES = \
 	p11-kit/lists.c \
 	p11-kit/p11-kit.c \
+	p11-kit/print-config.c \
 	$(NULL)
 
 p11_kit_p11_kit_LDADD = \
 	libp11-kit.la \
+	libp11-kit-internal.la \
 	libp11-tool.la \
 	libp11-common.la \
 	$(LTLIBINTL) \
--- a/p11-kit/meson.build
+++ b/p11-kit/meson.build
@@ -141,14 +141,15 @@ endif
 
 p11_kit_sources = [
   'lists.c',
-  'p11-kit.c'
+  'p11-kit.c',
+  'print-config.c'
 ]
 
 executable('p11-kit',
            p11_kit_sources,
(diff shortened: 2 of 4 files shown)


p11-kit/meson.build
```diff
@@ -140,13 +140,15 @@ if get_option('test')
 endif
 
 p11_kit_sources = [
+  'conf.c',
```
ueno, line 143: This means that the same file `p11-kit/conf.c` is compiled twice. It would be sufficient to just link to `libp11-kit-internal.a`.

p11-kit/Makefile.am
```diff
@@ -237,9 +237,13 @@ EXTRA_DIST += \
 
 bin_PROGRAMS += p11-kit/p11-kit
 
+p11_kit_p11_kit_CFLAGS = $(COMMON_CFLAGS)
+
 p11_kit_p11_kit_SOURCES = \
+	p11-kit/conf.c \
```
ueno, line 243: Same here, you could simply link to `libp11-kit-internal.la`.

ueno (changes requested):

ZoltanFridrich:
@ueno is the output format ok?

ueno:
> @ueno is the output format ok?

Looks good to me as the default (human readable) output. Eventually, we could support a structured format (e.g., JSON, TOML, etc.) with schema validation, that could be through a new option, e.g., `--format=json`.

ueno (approved):

ueno:
Thanks!

p11-kit/print-config.c
```diff
@@ -0,0 +1,137 @@
...
+		case '?':
+			return 2;
+		default:
+			assert_not_reached ();
+			break;
+		}
+	}
+
+	if (argc - optind != 0) {
+		p11_message (_("extra arguments specified"));
```
piotrdrag, line 132: `p11-kit/print-config.c` is missing from `po/POTFILES.in`.

ZoltanFridrich: I have created MR: https://github.com/p11-glue/p11-kit/pull/448/files

PR #459, opened 2023-03-02
rpc-server: Check pointer value returned by proto_read_byte_array

ueno:
The commit b58b725f30ea833ce8ff08879f616b32be44cbdb added a case where proto_read_byte_array may return non-zero length while the pointer location is set to NULL.  This is to accommodate the change in rpc_C_Initialize.

Spotted by oss-fuzz:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=56504


ZoltanFridrich (approved):
No problems with the change. Approved.

PR #461, opened 2023-03-03
Add PKCS#11 3.0 support to proxy

ZoltanFridrich:
Expands the proxy modules function list to support pkcs#11 3.0 functions.


p11-kit/client.c
```diff
@@ -49,7 +49,7 @@
 typedef struct _State {
```
ueno, line 49: This file is unrelated to the proxy.

p11-kit/virtual.c
```diff
@@ -3062,24 +4177,30 @@ static CK_FUNCTION_LIST *
 p11_virtual_wrap_fixed (p11_virtual *virt,
```
ueno, line 4177: We can change the return type of this function to `CK_FUNCTION_LIST_3_0`, given the PKCS#11 3.0 functions are always present.

ueno (commented):
I've added the first round of comments; maybe good to think about when to use `CK_FUNCTION_LIST` or `CK_FUNCTION_LIST_3_0`.

p11-kit/proxy.c
```diff
@@ -67,7 +68,7 @@
 typedef struct _Mapping {
 	CK_SLOT_ID wrap_slot;
 	CK_SLOT_ID real_slot;
-	CK_FUNCTION_LIST_PTR funcs;
+	CK_FUNCTION_LIST_3_0_PTR funcs;
```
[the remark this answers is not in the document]
ZoltanFridrich: if I change it back to CK_FUNCTION_LIST_PTR it will need to be retyped inside all 3.0 funcs, is that ok?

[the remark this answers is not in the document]
ueno: I would say that should be ok. The point is to prevent accessing `CK_FUNCTION_LIST_PTR` as `CK_FUNCTION_LIST_3_0_PTR`; inside 3.0 funcs, we know what we are accessing.

p11-kit/virtual.c
```diff
@@ -2875,6 +3927,23 @@ p11_virtual_wrap (p11_virtual *virt,
 
 #endif /* !FFI_CLOSURES */
 
+CK_FUNCTION_LIST *
+p11_virtual_wrap_version (p11_virtual *virt,
```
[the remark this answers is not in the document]
ZoltanFridrich: It does seem that we indeed don't need it. I will remove it completely.

p11-kit/proxy.c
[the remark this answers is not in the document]
ZoltanFridrich: done

p11-kit/p11v3-module.map
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
ueno: I guess we could remove this file as well as the .def counterpart.

ZoltanFridrich:
All threads have been addressed.

p11-kit/proxy.c
```diff
@@ -1777,6 +2273,10 @@ C_GetFunctionList (CK_FUNCTION_LIST_PTR_PTR list)
 				rv = CKR_GENERAL_ERROR;
 
 			} else {
+				if (version) {
+					module->version.major = version->major;
+					module->version.minor = version->minor;
```
ueno, line 2278: nit: why not just assign like the above line 2265? `module->version = *version;`?

p11-kit/proxy.c
```diff
@@ -1785,11 +2285,153 @@ C_GetFunctionList (CK_FUNCTION_LIST_PTR_PTR list)
...
+		*pulCount = NUM_INTERFACES;
+		return CKR_BUFFER_TOO_SMALL;
+	}
+
+	p11_library_init_once ();
+	p11_lock ();
+
+	rv = internal_init_interfaces ();
+	if (rv == CKR_OK) {
+		memcpy (pInterfacesList, interfaces, NUM_INTERFACES * sizeof(CK_INTERFACE));
```
ueno, line 2373: Can't `internal_init_interfaces` be extended to do this `memcpy` internally?

p11-kit/proxy.c
```diff
@@ -1785,11 +2285,153 @@ C_GetFunctionList (CK_FUNCTION_LIST_PTR_PTR list)
...
+/* We are not going to support any special interfaces */
+#define NUM_INTERFACES 2
+#define DEFAULT_INTERFACE 0
+CK_INTERFACE interfaces[NUM_INTERFACES] = {
+        {"PKCS 11", NULL, 0}, /* 3.0 */
+        {"PKCS 11", NULL, 0}  /* 2.4 */
+};
+
+static int
+internal_init_interfaces (void)
```
ueno, line 2307: nit: `internal_` seems redundant as it is a static function?

ueno (approved):
Thank you; looks good to me.

p11-kit/proxy.c
ZoltanFridrich: I wouldnt do that. Right now the init_interfaces function does exactly what it is supposed to: initializes internal interfaces array. I think it makes more sense this way.

ueno: If we do this way (i.e., copy the result), we don't need to define `interfaces` as global; it can be a local variable in this function.

ZoltanFridrich: I am keeping it as is. I cant make it into local variable without reworking a lot more then just `init_interfaces`, for example this part inside `C_GetInterface` expects it to be global as well:
`*ppInterface = &interfaces[i];`

PR #463, opened 2023-03-09
Add PKCS#11 3.0 support to RPC

ZoltanFridrich:
Adds RPC calls for pkcs#11 3.0 functions


p11-kit/client.c
```diff
@@ -163,7 +163,7 @@ p11_client_module_cleanup (void)
 	for (; state != NULL; state = next) {
 		next = state->next;
 		p11_rpc_transport_free (state->rpc);
-		p11_virtual_unwrap (state->wrapped);
+		p11_virtual_unwrap ((CK_FUNCTION_LIST *)state->wrapped);
```
ueno, line 166: Not a fault of this PR, but we probably should change `p11_virtual_unwrap` to take `CK_FUNCTION_LIST_3_0 *` instead of `CK_FUNCTION_LIST *` to be consistent with `p11_virtual_wrap`.

p11-kit/rpc-client.c
```diff
@@ -595,6 +596,10 @@ proto_read_sesssion_info (p11_rpc_message *msg,
 	if (!p11_rpc_message_write_byte_buffer (&_msg, arr ? (*len > 0 ? *len : (uint32_t)-1) : 0)) \
 		{ _ret = CKR_HOST_MEMORY; goto _cleanup; }
 
+#define IN_BYTE_BUFFER_NULL(arr, len) \
+	if (!p11_rpc_message_write_byte_buffer_null (&_msg, len)) \
```
ueno, line 600: These macros (`IN_BYTE_BUFFER_NULL` and `OUT_BYTE_ARRAY_NULL`) and the support function (`p11_rpc_message_write_byte_buffer_null`) are only used to handle the special case in [C_SignMessageNext](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/pkcs11-spec-v3.1.pdf#page=145), which I would rather expand at the caller.

p11-kit/client.c
ZoltanFridrich: but `p11_virtual_wrap` returns `CK_FUNCTION_LIST *` should I change all of these?

ZoltanFridrich: this change would be huge and would require to retype the entire module.c and probably another modules as well, like proxy.c ... I think if we want to do this, then in a separate PR

p11-kit/rpc-client.c
```diff
@@ -1883,6 +1884,379 @@ rpc_C_GenerateRandom (CK_X_FUNCTION_LIST *self,
...
+		/* Make sure this is in the right order */
+		assert (!_msg.signature || p11_rpc_message_verify_part (&_msg, "fy"));
+		/* Validity byte */
+		p11_rpc_buffer_add_byte (_msg.output, signature_len ? 1 : 0);
+		if (signature_len)
+			p11_rpc_buffer_add_uint32 (_msg.output, *signature_len);
+		if (p11_buffer_failed (_msg.output)) {
+			_ret = CKR_HOST_MEMORY;
+			goto _cleanup;
+		}
```
ueno, line 2166: Sorry, what I really meant was to change the protocol [signature](https://github.com/p11-glue/p11-kit/pull/463/files#diff-4c71e18b8f308838a79477ed5669813f625782daf55b2dd28714fad582afa091R245) of C_SignMessageNext from "uayayfy" to "uayayyfy" (note the extra "y" before "fy"). Then you could write this part something like:
```c
IN_BYTE (signature_present)
if (signature_present)
  IN_BYTE_BUFFER (signature, signature_len)
```

p11-kit/rpc-server.c
```diff
@@ -1834,6 +1878,416 @@ rpc_C_GenerateRandom (CK_X_FUNCTION_LIST *self,
...
+		IN_ULONG (session);
+		IN_BYTE_ARRAY (parameter, parameter_len);
+		IN_BYTE_ARRAY (data, data_len);
+		_ret = proto_read_byte_buffer_null (msg, &signature, &signature_len);
+		if (_ret != CKR_OK)
+			goto _cleanup;
+	PROCESS_CALL ((self, session, (void *)parameter, parameter_len, data, data_len,
+	               signature, signature_len));
+		OUT_BYTE_ARRAY (signature, signature_len ? *signature_len : 0);
+	END_CALL;
```
ueno, line 2189: Sorry, what I really meant was to change the protocol [signature](https://github.com/p11-glue/p11-kit/pull/463/files#diff-4c71e18b8f308838a79477ed5669813f625782daf55b2dd28714fad582afa091R245) of C_SignMessageNext from "uayayfy" to "uayayyfy" (note the extra "y" before "fy"). Then you could write this part something like:

```suggestion
	BEGIN_CALL (SignMessageNext);
		IN_ULONG (session);
		IN_BYTE_ARRAY (parameter, parameter_len);
		IN_BYTE_ARRAY (data, data_len);
		IN_BYTE (signature_expected);
		IN_BYTE_BUFFER (signature, signature_len);
	PROCESS_CALL ((self, session, (void *)parameter, parameter_len, data, data_len,
	               signature_expected ? signature : NULL, signature_expected ? signature_len : NULL));
		OUT_BYTE_ARRAY (signature, signature_expected ? *signature_len : 0);
	END_CALL;
```

ZoltanFridrich: could you take a look at this part again? I think rpc-server should be good now. Dont know about rpc-client

p11-kit/client.c
ueno: Let's postpone it (or do it in a separate PR).

ueno (approved):
Looks good to me!

PR #465, opened 2023-03-15
.github/workflows/test.yaml: Switch to using macOS 12 image

ueno:


trust/token.c
```diff
@@ -304,19 +310,18 @@ loader_load_path (p11_token *token,
...
 
 	if (stat (path, &sb) < 0) {
 		if (errno != ENOENT)
 			p11_message_err (errno, _("cannot access trust certificate path: %s"), path);
 		loader_gone_file (token, path);
 		*is_dir = false;
-		ret = 0;
+		return 0;
 
 	} else if (S_ISDIR (sb.st_mode)) {
```
ZoltanFridrich, line 323: doesnt have to be if else when we return from previous block

ZoltanFridrich (approved):
Only one nit, otherwise LGTM

Issue #466, opened 2023-03-15
test-managed failed


jkr1987:
hello i'm trying to build p11-kit on ubuntu 22

when i run : 
`$ meson test -C _build`

that give me :

```
=================================== 27/33 ====================================
test:         test-managed
start time:   13:51:51
duration:     0.21s
result:       killed by signal 6 SIGABRT
command:      MALLOC_PERTURB_=13 /home/p11-kit/_build/p11-kit/test-managed
----------------------------------- stderr -----------------------------------
p11-kit: 'result != NULL' not true at p11_virtual_wrap
==============================================================================
```

Thankful for any wise thoughts.

ueno:
Thank you for the report. I can reproduce if I compile p11-kit without libffi, even with an increased number of closures. We probably should exercise the configuration in the CI as well.

jkr1987:
tank you ! i installed libffi and that works !

PR #473, opened 2023-03-18
tests: Make test-managed succeed without libffi

ueno:
The test_max_session_load test was assuming that the closure assignment would succeed even if the fixed closures are exhausted, thanks to the fallback mechanism when libffi is enabled.  This conditionalizes the test with WITH_FFI and do not exceed the limit if libffi is not enabled.

Fixes: #466


ZoltanFridrich (approved):
Seems good to me.

PR #475, opened 2023-03-21
Add tests for pkcs#11 3.0

ZoltanFridrich:


p11-kit/test-rpc3.c
```diff
@@ -0,0 +1,771 @@
...
+
+#include "test-mock.c"
+
+CK_VERSION test_version_three = {CRYPTOKI_VERSION_MAJOR, CRYPTOKI_VERSION_MINOR};
+
+int
+main (int argc,
+      char *argv[])
+{
+	CK_MECHANISM_TYPE mechanisms[] = {
```
ueno, line 734: Let's use `static const` so the array will be part of the text segment instead of a stack. Also I have a little preference on defining it outside the function.

p11-kit/test-rpc3.c
```diff
@@ -0,0 +1,771 @@
...
+#include <assert.h>
+#include <limits.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#ifdef OS_UNIX
+#include <unistd.h>
+#endif
+
+static p11_virtual base;
```
ueno, line 63: This file seems to be almost identical to the latter half of `p11-kit/test-rpc.c`. Can we somehow consolidate them?

For example:
- create a new `p11-kit/test-rpc-message.c`, to test the low-level message handling (the first part of `p11-kit/test-rpc.c`)
- merge `p11-kit/test-rpc3.c` into `p11-kit/test-rpc.c`

p11-kit/test-log3.c
```diff
@@ -0,0 +1,115 @@
...
+	rv = p11_module_release_inlock_reentrant (module);
+	assert (rv == CKR_OK);
+
+	p11_unlock ();
+}
+
+/* Bring in all the mock module tests */
+#include "test-mock.c"
+
+CK_VERSION test_version_three = {CRYPTOKI_VERSION_MAJOR, CRYPTOKI_VERSION_MINOR};
```
ueno, line 100: `static const`

ZoltanFridrich: Cant be const. `test_mock_add_tests` requires takes non-const CK_VERSION

ZoltanFridrich: actually... I can just change the function parameter to take const

p11-kit/test-rpc3.c
ZoltanFridrich: Cant be const. But I can move it outside and make it static.

p11-kit/test-transport3.c
```diff
@@ -0,0 +1,452 @@
...
+	char *directory;
+	char *user_config;
+	char *user_modules;
+#ifdef OS_UNIX
+	pid_t pid;
+#endif
+} test;
+
+static void
+setup_remote (void *unused)
```
ueno, line 70: This test also shares quite a lot with `p11-kit/test-transport.c`. Can these two be consolidated as well?

ZoltanFridrich: ok

ZoltanFridrich: I think there is a bit of a problem with this.
In test-rpc I was able to merge them into one because I used v3 module but tested both v2 and v3 interfaces.
However in test-transport only the used module changes. So either I keep it separate to test v2 module and v3 module or I merge it such that I only test v3 module. I dont know how else I would do it.

PR #478, opened 2023-03-28
server: Fix umask setting when --group is specified

ueno:
p11-kit server previously set umask(066) unconditionally before chown and that resulted in removing group read/write bits, even if --group is specified.  This fixes the logic to respect --group and also allows it to be specified without --user.

Originally reported in:
https://lists.freedesktop.org/archives/p11-glue/2023-January/000725.html


ueno:
I'm filing this on behalf of the original submitter, and the change looks good; merging without approval.

Issue #477, opened 2023-03-24
Mechanisms may use different parameter structures depending on the PKCS#11 function being called


uhhhh2:
According to [PKCS 3.1 Section 6.13.5 (AES GCM and CCM Mechanism Parameters)](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/cs01/pkcs11-spec-v3.1-cs01.html#_Toc111203499)

The `CKM_AES_GCM` mechanism uses the `CK_GCM_PARAMS` structure when calling `C_EncryptInit` or `C_DecryptInit`

```C
typedef struct CK_GCM_PARAMS {
   // Initialization Vector (IV)
   CK_BYTE_PTR   pIv;  
   // Length of IV in bytes
   CK_ULONG     ulIvLen; 
   // Length of IV in bits 
   CK_ULONG     ulIvBits; 
   // pointer to additional authenticated data (AAD)
   CK_BYTE_PTR   pAAD; 
   // length of AAD in bytes
   CK_ULONG     ulAADLen;
   // length of authentication tag in bits 
   CK_ULONG     ulTagBits;
}  CK_GCM_PARAMS;
```

However, the `CKM_AES_GCM` mechanism uses the `CK_GCM_MESSAGE_PARAMS` when calling `C_MessageEncryptInit`  or `C_MessageDecryptInit`

```C
typedef struct CK_GCM_MESSAGE_PARAMS {
    // Initialization Vector (IV)
   CK_BYTE_PTR   pIv;
   // length of IV in bytes
   CK_ULONG     ulIvLen; 
   // Number of bits to preserve when generating a new IV
   CK_ULONG     ulIvFixedBits; 
    // (Numeric ID of a) Function to use when generating a new IV
   CK_GENERATOR_FUNCTION   ivGenerator; 
    // location of authentication tag 
   // (returned in MessageEncrypt, provided in MessageDecrypt)
   CK_BYTE_PTR   pTag;
   // length of authentication tag in bits 
   CK_ULONG     ulTagBits; 
}  CK_GCM_MESSAGE_PARAMS;

typedef CK_ULONG CK_GENERATOR_FUNCTION;
```

The latter struct adds a `pTag` byte array for storing the authentication tag separately from the ciphertext, and the (numeric ID of) an `ivGenerator` function to use when generating IVs.  

[...]

PR #481, opened 2023-04-02
Add AES-GCM mechanism parameters for message-based encryption

ueno:
When used with C_MessageEncrypt and C_MessageDecrypt, CKM_AES_GCM takes different parameters than CK_GCM_PARAMS.

Fixes: #477


ZoltanFridrich (approved):
I havent find any problems, all the changes look good.

PR #483, opened 2023-04-06
trust: Use the same parser code for parsing and checking

ueno:
The trust check-format command used a dedicated parsing code for the .p11-kit format, which behaves differently from the original parsing code: sometimes it is stricter (e.g., PEM block), while other times it is not (e.g., constant names and OIDs). This makes it use the same single parser to align with the original behavior.

### DIFF
--- a/trust/check-format.c
+++ b/trust/check-format.c
@@ -47,7 +47,6 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
 
 #ifdef ENABLE_NLS
 #include <libintl.h>
--- a/trust/check-format.c
+++ b/trust/check-format.c
@@ -39,14 +39,14 @@
 #include "check-format.h"
 #include "debug.h"
 #include "message.h"
+#include "persist.h"
 #include "tool.h"
 
 #include <assert.h>
-#include <ctype.h>
+#include <errno.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <string.h>
 #include <unistd.h>
 
 #ifdef ENABLE_NLS
@@ -56,20 +56,12 @@
 #define _(x) (x)
 #endif
 
-#define IS_BASE64(C) (isalnum ((int)(C)) || (C) == '+' || (C) == '/')
-#define PEM_LABEL_MAX 256
-
 enum format_result {
 	FORMAT_OK,
 	FORMAT_FAIL,
 	FORMAT_ERROR
 };
 
-static char pem_label[PEM_LABEL_MAX];
-static int pem_label_length;
-static int inside_pem;
-static int inside_section;
-static size_t line_num;
 static bool color_out;
 static bool color_err;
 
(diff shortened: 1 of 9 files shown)


ueno:
With this (bogus) input:
```text
[p11-kit-object-v1]
class: bogus-class
label: bogus-label
object-id: 2.5.29.30.X
value: %30%2e%06%03%55%1d%1e%04%27%30%25%a0%23%30%0d%82%0b%65%78%61%6d%70%6c%65%2e%63%6f%6d%30%12%82%10%63%6f%72%70%2e%65%78%61%6d%70%6c%65%2e%63%6f%6d
-----BEGIN PUBLIC KEY-----
MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDRtTajie6qgC9T/RJ1PvN6ntav
+rwcYBBLJoETGlnj/kVsOAQ5J0ZX/dW8jYoQtjvUCoFaRS/sPoHw2U5Pl99LMg8I
sSaivWlhXWY5Yy8QcDX7B4UK/1cSwfSDHfnG06S2cCuAoUB/SE7ZreuAzM+SwdGD
ZAEjR469MZgFa2t8NwIDAQAB
-----END TEST KEY-----
```
Old:
```console
$ _build/trust/trust check-format bogus.p11-kit
Failed at line 11: label in the END block does not match the label in the BEGIN block
bogus.p11-kit: FAIL
```
New:
```console
$ _build/trust/trust check-format bogus.p11-kit
p11-kit: bogus.p11-kit:2: class: invalid value
p11-kit: bogus.p11-kit:3: label: invalid value
p11-kit: bogus.p11-kit:4: object-id: invalid value
p11-kit: bogus.p11-kit:5: value: invalid value
p11-kit: bogus.p11-kit:11: BEGIN ...: invalid pem block
bogus.p11-kit: FAIL
```

trust/persist.h
```diff
@@ -36,8 +36,10 @@
 #define P11_PERSIST_H_
 
 #include "array.h"
+#include "buffer.h"
```
ZoltanFridrich, line 39: I would put these includes into the source file

trust/check-format.c
```diff
@@ -39,10 +39,12 @@
 #include "check-format.h"
 #include "debug.h"
 #include "message.h"
+#include "persist.h"
 #include "tool.h"
 
 #include <assert.h>
 #include <ctype.h>
+#include <errno.h>
```
ZoltanFridrich, line 47: I think you can remove some includes, like string.h

ZoltanFridrich (approved):
I would just fixup some includes but looks good overall.

trust/persist.h
ueno: "buffer.h" needs to be included, as `p11_persist_write` takes a `p11_buffer` argument.

trust/check-format.c
ueno: Done; that also reminds me that we should provide a stub for `isatty` on non-Unix platforms, fixed that as well.

PR #485, opened 2023-04-09
trust: Lower Cryptoki version from 3.0 to 2.40

ueno:
The trust module currently stays at Cryptoki version 2.40, while 3.0 was advertised through the `version` field in `CK_FUNCTION_LIST` and `C_GetInfo`.


ZoltanFridrich (approved):
OK.

PR #486, opened 2023-04-09
client: Support PKCS#11 3.0

ueno:
This exports `C_GetInterfaceList` and `C_GetInterface` from the client module.

### DIFF
--- a/p11-kit/client.c
+++ b/p11-kit/client.c
@@ -36,6 +36,7 @@
 
 #include "client.h"
 #include "compat.h"
+#include "debug.h"
 #include "library.h"
 #include "runtime.h"
 #include "path.h"
@@ -49,7 +50,7 @@
 typedef struct _State {
 	p11_virtual virt;
 	p11_rpc_transport *rpc;
-	CK_FUNCTION_LIST_3_0 *wrapped;
+	CK_INTERFACE wrapped;
 	struct _State *next;
 } State;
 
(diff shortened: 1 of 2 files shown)


p11-kit/client.c
```diff
@@ -97,57 +97,176 @@ get_server_address (char **addressp)
...
+	}
+
+	state->wrapped.pInterfaceName = (char *)p11_interface_name;
+
+	state->wrapped.pFunctionList = (CK_FUNCTION_LIST_3_0 *)module;
+	module = NULL;
+
+	state->wrapped.flags = flags;
+
+	*interface = &state->wrapped;
```
ZoltanFridrich, line 158: I would put an assert on interface not being NULL

p11-kit/proxy.c
```diff
@@ -2270,48 +2273,59 @@ get_interface_inlock(CK_INTERFACE **interface, const CK_VERSION *version, CK_FLA
...
+		module->version = *version;
+
+	state->wrapped.pInterfaceName = (char *)p11_interface_name;
+
+	state->wrapped.pFunctionList = module;
+	module = NULL;
+
+	state->wrapped.flags = flags;
+
+	*interface = &state->wrapped;
```
ZoltanFridrich, line 2310: assert that interface is not NULL

ZoltanFridrich (changes requested):
Looks good, but I have some comments.

p11-kit/client.c
```diff
@@ -97,57 +97,176 @@ get_server_address (char **addressp)
...
 	}
 
-	p11_unlock ();
+	p11_library_init_once ();
+	p11_lock ();
 
-	free (address);
+	if (pVersion &&
+	    memcmp (pVersion, &version_three, sizeof(*pVersion)) == 0)
+		pVersion = NULL;
```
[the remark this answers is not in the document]
ueno: Indeed, that's a good point; I've removed the special case and made `get_interface_inlock` always expects non-NULL `version`.

ZoltanFridrich (approved):

PR #489, opened 2023-04-11
Document print-config command

### LINKED ISSUE #484: No documentation for `p11-kit print-config`
Since #446 we have `p11-kit print-config` command to show the actual configuration used. However, the command is not documented in the manual (also not appear in the bash-completion).

ZoltanFridrich:
* adds documenation for print-config command
* adds bash-completion for print-config command
* adds bash-completion for check-format command

Closes #484

### DIFF
--- a/bash-completion/trust
+++ b/bash-completion/trust
@@ -7,7 +7,7 @@ _trust()
 
     local commands command
 
-    commands='list extract extract-compat anchor dump'
+    commands='list extract extract-compat anchor dump check-format'
 
     if [[ $cword -eq 1 ]]; then
         COMPREPLY=( $(compgen -W "$commands" -- "$cur") )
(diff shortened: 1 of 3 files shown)


bash-completion/trust
```diff
@@ -7,7 +7,7 @@ _trust()
 
     local commands command
 
-    commands='list extract extract-compat anchor dump'
+    commands='list extract extract-compat anchor dump check-format'
```
ueno, line 10: Haven't checked, but does this auto-completes filename?

ueno (approved):
Looks good to me except a minor question.

bash-completion/trust
ZoltanFridrich: I believe it should.

PR #491, opened 2023-04-12
.github/actions: Don't swallow error exit code when printing logs

ueno:
It turned out that the CI didn't fail in meson builds, as the exit code was swallowed after printing the error log. This series fixes it as well as actual issues found by ASan and clang-analyzer.


ueno:
Thank you for the review!

PR #492, opened 2023-04-12
virtual: Fix libffi type signatures for PKCS#11 3.0 functions

ueno:
There were some mismatches between the type signatures of `binding_*`
functions and the `BindingInfo` definitions, causing test failures
when compiled without fixed closures.


Jakuje (approved):
lgtm. Good to have this covered by CI too now!

ueno:
Thanks for the review!

PR #496, opened 2023-04-17
build: Fix meson build & tests on macOS

neverpanic:
This PR consists of two parts:

- Fixing linker errors when building with meson on macOS
- Ensuring loadable modules used for tests are correctly built with an `.so` filename extension on macOS

### build: Fix linker errors on macOS
This fixes a few linker errors on macOS due to an undefined `p11_library_mutex` symbol:

```
cc  -o p11-kit/p11-kit-server-testable p11-kit/p11-kit-server-testable.p/server.c.o -Wl,-dead_strip_dylibs -Wl,-headerpad_max_install_names -Wl,-undefined,error -Wl,-rpath,/opt/local/lib p11-kit/libp11-kit-testable.a common/libp11-library.a common/libp11-common.a common/libp11-tool.a /opt/local/lib/libffi.dylib
Undefined symbols for architecture x86_64:
  "_p11_library_mutex", referenced from:
      _p11_kit_be_quiet in libp11-kit-testable.a(util.c.o)
      _p11_kit_be_loud in libp11-kit-testable.a(util.c.o)
      _p11_kit_set_progname in libp11-kit-testable.a(util.c.o)
ld: symbol(s) not found for architecture x86_64
```

### build: Use .so as module suffix on macOS

macOS uses `.so` as filename extension for loadable modules. For some reason, meson does not use this by default, but uses `.dylib`. This inconsistency between the autoconf build system (which uses `.so`) and the meson build system caused the tests to fail, because the mock modules would not be found by the tests.

This is a bug in meson, as loadable modules (filetype `BUNDLE` in `otool -hv` output) should have an `.so` extension on macOS. See [[1][1],[2][2]] for discussion in meson.

Port the same fix that was used by the GNOME developers in gtk [[3][3],[4][4]] and glib-networking [\[5\]][5]. See also a similar fix in glib [\[6\]][6].

[1]: https://github.com/mesonbuild/meson/issues/1160
[2]: https://github.com/mesonbuild/meson/issues/3053
[3]: https://gitlab.gnome.org/GNOME/gtk/-/issues/3645
[4]: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3162
[5]: https://gitlab.gnome.org/GNOME/glib-networking/-/merge_requests/1
[6]: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/280

### DIFF
--- a/trust/meson.build
+++ b/trust/meson.build
@@ -142,12 +142,23 @@ if get_option('test')
     'test-jks'
   ]
 
+  # Some tests fail to link on macOS because they need p11_library_mutex, but
+  # it isn't included unless libp11_kit_testable is linked with
+  # -Wl,--whole-archive or -Wl,-force_load.
+  trust_tests_whole = [
+    'test-index',
+    'test-builder',
+    'test-token'
+  ]
+
   foreach name : trust_tests
     t = executable(name, '@0@.c'.format(name),
                    c_args: common_c_args + tests_c_args + libtrust_testable_c_args,
                    dependencies: [asn_h_dep,
                                   libp11_kit_dep,
-                                  libp11_library_dep,
+                                  trust_tests_whole.contains(name) ?
+                                    libp11_library_whole_dep :
+                                    libp11_library_dep,
                                   libp11_test_dep] + libtasn1_deps + dlopen_deps,
                    link_with: [libtrust_testable, libtrust_data, libtrust_test])
     test(name, t)
@@ -165,12 +176,21 @@ if get_option('test')
     'frob-oid'
   ]
 
+  # Some tests fail to link on macOS because they need p11_library_mutex, but
+  # it isn't included unless libp11_kit_testable is linked with
+  # -Wl,--whole-archive or -Wl,-force_load.
+  trust_progs_whole = [
+    'frob-token'
+  ]
+
   foreach name : trust_progs
     t = executable(name, '@0@.c'.format(name),
                    c_args: tests_c_args,
                    dependencies: [asn_h_dep,
                                   libp11_kit_dep,
-                                  libp11_library_dep,
+                                  trust_progs_whole.contains(name) ?
+                                    libp11_library_whole_dep :
+                                    libp11_library_dep,
                                   libp11_test_dep] + libtasn1_deps + libffi_deps + dlopen_deps,
                    link_with: [libtrust_testable, libtrust_data, libtrust_test])
   endforeach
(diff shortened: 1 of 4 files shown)


mouse07410:
With this PR applied, all tests pass, except for one (57) that timeouts:
```text
57/57 test-server              TIMEOUT        30.01s   exit status 7
>>> MALLOC_PERTURB_=146 /Users/ur20980/src/p11-kit/_build/p11-kit/test-server


Ok:                 56  
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            1   
```

Testlog: [testlog.txt](https://github.com/p11-glue/p11-kit/files/11250530/testlog.txt)

MacOS Ventura 13.3.1, Xcode-14.3.

meson.build
```diff
@@ -1,6 +1,6 @@
 project('p11-kit', 'c',
         version: '0.24.1',
-        meson_version: '>= 0.49')
+        meson_version: '>= 0.56')
```
ueno, line 3: I suppose this is for `as_link_whole`; perhaps we could simply declare a new dependency with the `link_whole` keyword to make it work with 0.49?

IIRC we stick to 0.49 for compatibility reasons, as some distos only have that version. Maybe it's time to bump it anyway.

trust/meson.build
```diff
@@ -142,12 +142,25 @@ if get_option('test')
...
+    'test-index',
+    'test-builder',
+    'test-token'
+  ]
+
   foreach name : trust_tests
+    libp11 = libp11_library_dep
+    if trust_tests_whole.contains(name)
+      libp11 = libp11_library_dep.as_link_whole()
+    endif
```
ueno, line 158: nit: can this be simply written with a [ternary operator](https://mesonbuild.com/Syntax.html#ternary-operator)?

ueno (approved):
Looks good to me. Only one question regarding Meson version bump.

neverpanic:
> While you are at it, could you also update the CI to use Meson on macOS?

I tried that but stopped the attempt, because your Meson GitHub Action expects `useradd` and `runuser`, which don't exist on macOS, and my attempt to avoid them failed in lots of places. I think that's something that can be done in an additional PR.

meson.build
neverpanic: RHEL 8 has 0.58, but I changed it anyway.

ueno:
Thank you!

PR #499, opened 2023-04-26
.github/workflows: Switch macOS CI to using Meson

ueno:


ueno:
@neverpanic could you review this?

neverpanic (approved):
Looks good to me!

.github/workflows/test.yaml
```diff
@@ -91,19 +94,21 @@ jobs:
...
+    env:
+      MESON_BUILD_OPTS: -Dclosures=0
     steps:
       # Checkout repo
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: |
           brew update
-          for pkg in autoconf automake coreutils gettext libffi libtasn1 libtool; do
+          for pkg in autoconf automake coreutils gettext libffi libtasn1 libtool meson; do
```
neverpanic, line 105: Actually, since you're now using meson, I guess you could remove autoconf and automake?

.github/workflows/test.yaml
```diff
@@ -91,19 +94,21 @@ jobs:
...
             brew install $pkg || true
```
neverpanic, line 106: Any reason you don't just use `brew install coreutils gettext libffi libtasn1 meson` instead of the loop?

ueno:
Thanks for the suggestions!

Issue #394, opened 2021-11-13
Add basic key and object manipulation subcommands to p11-kit


ueno:
It would be nice if the following operations could be done with `p11-kit` alone without installing other packages (e.g., `p11tool` from GnuTLS or `pkcs11-tool` from OpenSC):

- [x] key pair generation
- [x] enumerate objects (keys, certificates)
- [ ] import object
- [x] export object
- [x] delete object

Other operations, such as certificate generation, are out of scope.

PR #500, opened 2023-04-26
Add list-profiles sub-command to p11-kit cmd tool

ZoltanFridrich:
Adds list-profiles command for listing profiles that are supported by a token
usage: p11-kit list-profiles pkcs11:token

### DIFF
--- a/common/constants.c
+++ b/common/constants.c
@@ -636,6 +636,15 @@ const p11_constant p11_constant_mechanisms[] = {
 	{ CKA_INVALID },
 };
 
+const p11_constant p11_constant_profiles[] = {
+	CT (CKP_BASELINE_PROVIDER, "baseline-provider")
+	CT (CKP_EXTENDED_PROVIDER, "extended-provider")
+	CT (CKP_AUTHENTICATION_TOKEN, "authentication-token")
+	CT (CKP_PUBLIC_CERTIFICATES_TOKEN, "public-certificates-token")
+	CT (CKP_VENDOR_DEFINED, "vendor-defined")
+	{ CKA_INVALID },
+};
+
 #undef CT
 
 struct {
@@ -653,6 +662,7 @@ struct {
 	{ p11_constant_states, ELEMS (p11_constant_states) - 1 },
 	{ p11_constant_users, ELEMS (p11_constant_users) - 1 },
 	{ p11_constant_returns, ELEMS (p11_constant_returns) - 1 },
+	{ p11_constant_profiles, ELEMS (p11_constant_profiles) - 1 },
 };
 
 static int
--- a/p11-kit/list-profiles.c
+++ b/p11-kit/list-profiles.c
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) 2023, Red Hat Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the
+ *       following disclaimer.
+ *     * Redistributions in binary form must reproduce the
+ *       above copyright notice, this list of conditions and
+ *       the following disclaimer in the documentation and/or
+ *       other materials provided with the distribution.
+ *     * The names of contributors to this software may not be
+ *       used to endorse or promote products derived from this
+ *       software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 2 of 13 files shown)


p11-kit/list-profiles.c
```diff
@@ -0,0 +1,232 @@
...
+p11_kit_list_profiles (int argc,
+		       char *argv[]);
+
+typedef struct {
+	CK_PROFILE_ID id;
+	const char *name;
+} profile_st;
+
+static profile_st profile_map[] = {
+	{ CKP_BASELINE_PROVIDER, "baseline_provider" },
```
ueno, line 66: As I mentioned previously, it would make sense to integrate this mapping and `profile_to_str` below to `common/constants.c`.

p11-kit/meson.build
```diff
@@ -146,6 +146,7 @@ if get_option('test')
 endif
 
 p11_kit_sources = [
+  'list-profiles.c',
```
ueno, line 149: Add this file to `po/POTFILES.in` as well.

common/constants.c
```diff
@@ -636,6 +636,15 @@ const p11_constant p11_constant_mechanisms[] = {
 	{ CKA_INVALID },
 };
 
+const p11_constant p11_constant_profiles[] = {
+	CT (CKP_BASELINE_PROVIDER, "baseline_provider")
```
ueno, line 640: Let's use "kebab-case", rather than "snail_case" for consistency with other constants.

common/constants.c
```diff
@@ -636,6 +636,15 @@ const p11_constant p11_constant_mechanisms[] = {
 	{ CKA_INVALID },
 };
 
+const p11_constant p11_constant_profiles[] = {
```
ueno, line 639: This needs to be registered in `tables`.

ueno (changes requested):
I think this PR really deserves a test case. The below is what I used for testing it locally, but ideally there should be a test integrated into the test suite:

1. Modify p11-kit/mock-module-ep.c
```diff
diff --git a/p11-kit/mock-module-ep.c b/p11-kit/mock-module-ep.c
index 4324433..4845dcd 100644
--- a/p11-kit/mock-module-ep.c
+++ b/p11-kit/mock-module-ep.c
@@ -39,6 +39,14 @@
 
 #include "mock.h"
 
+static CK_RV
+override_initialize (CK_VOID_PTR init_args)
+{
+       CK_RV rv = mock_C_Initialize (init_args);
+       mock_module_add_profile (MOCK_SLOT_ONE_ID, CKP_PUBLIC_CERTIFICATES_TOKEN);
+       return rv;
+}
+
 #ifdef OS_WIN32
 __declspec(dllexport)
 #endif
@@ -47,6 +55,7 @@ C_GetFunctionList (CK_FUNCTION_LIST_PTR_PTR list)
 {
        mock_module_init ();
        mock_module.C_GetFunctionList = C_GetFunctionList;
+       mock_module.C_Initialize = override_initialize;
        if (list == NULL)
                return CKR_ARGUMENTS_BAD;
        *list = &mock_module;
```
2. Build
```console
meson setup _build -Dprefix=/usr
meson compile -C _build
```
4. Copy `mock-one.so` and `one.module`
```console
cp _build/p11-kit/mock-one.so /usr/lib64/pkcs11
cp p11-kit/fixtures/system-modules/one.module /usr/share/p11-kit/modules
```
5. Run `p11-kit list-profiles`
```console
_build/p11-kit/p11-kit list-profiles 'pkcs11:model=TEST%20MODEL;manufacturer=TEST%20MANUFACTURER;serial=TEST%20SERIAL;token=TEST%20LABEL'
```

ueno:
One way to test this would be to create a separate `p11-kit` executable for testing (e.g., `p11-kit-testable`), which links to `libp11-kit-testable.a`. The it should be able to access the mock modules installed under `p11-kit/fixtures/`.

ZoltanFridrich:
> One way to test this would be to create a separate `p11-kit` executable for testing (e.g., `p11-kit-testable`), which links to `libp11-kit-testable.a`. The it should be able to access the mock modules installed under `p11-kit/fixtures/`.

I thought that I might do it similar to test-iter where I would include the list-profiles.c and supply my own argv argc to the p11_kit_list_profiles, then I would redirect the stdout through pipe and read whether it contains correct profiles.

ueno:
> I thought that I might do it similar to test-iter where I would include the list-profiles.c and supply my own argv argc to the p11_kit_list_profiles, then I would redirect the stdout through pipe and read whether it contains correct profiles.

I'm not sure if it makes sense to test commands in C rather than simple shell scripts, given we intend to add more subcommands (e.g., #394). With shell script, you can simply diff the output against the expected one.

p11-kit/meson.build
```diff
@@ -158,6 +159,14 @@ executable('p11-kit',
            link_with: [libp11_kit, libp11_kit_internal],
            install: true)
 
+if get_option('test')
+  executable('p11-kit-list-profiles-testable',
```
ueno, line 163: Is there any reason this needs to be specific to the list-profiles subcommand? I mean, can we just create a p11-kit-testable executable from the same source files as p11-kit?

ZoltanFridrich: when I run the p11-kit-testable I get this error message:
`p11-kit: invalid config filename, will be ignored in the future: /home/zfridric/upstream/p11-kit/p11-kit/fixtures/system-modules/two.badname`

ZoltanFridrich: yes, it does affect the test execution. At least it shows up in the testlog.txt

PR #503, opened 2023-05-04
Add command to delete profiles from a token

ZoltanFridrich:
Adds the ability to delete profile objects from given PKCS#11 token.

$ p11-kit delete-profile --profile public-certificates-token 'pkcs11:token=MyToken'

### DIFF
--- a/p11-kit/delete-profile.c
+++ b/p11-kit/delete-profile.c
@@ -1,0 +1,211 @@
+/*
+ * Copyright (c) 2023, Red Hat Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the
+ *       following disclaimer.
+ *     * Redistributions in binary form must reproduce the
+ *       above copyright notice, this list of conditions and
+ *       the following disclaimer in the documentation and/or
+ *       other materials provided with the distribution.
+ *     * The names of contributors to this software may not be
+ *       used to endorse or promote products derived from this
+ *       software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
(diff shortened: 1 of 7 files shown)


p11-kit/delete-profile.c
```diff
@@ -0,0 +1,224 @@
...
+			if (profile == CKP_INVALID_ID)
+				profile = strtol (optarg, NULL, 16);
+			if (profile == CKP_INVALID_ID) {
+				p11_message (_("failed to convert profile argument: %s"), optarg);
+				goto cleanup;
+			}
+
+			if (n_profiles >= alloc_profiles) {
+				alloc_profiles += 4;
+				tmp = realloc (profiles, alloc_profiles * sizeof (CK_PROFILE_ID));
```
ueno, line 191: Would it be useful to support multiple profiles specified at once, given the user could always call `p11-kit delete-profile` multiple times from scripts / terminal? I would have it always take a single profile, following the "Do One Thing and Do It Well" [principle](https://en.wikipedia.org/wiki/Unix_philosophy#Do_One_Thing_and_Do_It_Well).

p11-kit/delete-profile.c
```diff
@@ -0,0 +1,224 @@
...
+
+	argc -= optind;
+	argv += optind;
+
+	if (argc != 1) {
+		p11_tool_usage (usages, options);
+		goto cleanup;
+	}
+
+	ret = delete_profile (*argv, profiles, n_profiles);
```
ueno, line 217: Should we check at least one profile is specified?

ueno (commented):
Can we add tests in a similar manner to list-profiles?

p11-kit/delete-profile.c
```diff
@@ -0,0 +1,224 @@
...
+			p11_kit_be_quiet ();
+			break;
+		case opt_help:
+			p11_tool_usage (usages, options);
+			ret = 0;
+			goto cleanup;
+		case opt_profile:
+			profile = p11_constant_resolve (profile_nicks, optarg);
+			if (profile == CKP_INVALID_ID)
+				profile = strtol (optarg, NULL, 16);
```
ueno, line 183: I would suggest relying on `strtol(..., 0)`, so the user can specify the profile id in both hexadecimal and decimal format.

ZoltanFridrich: ok

p11-kit/delete-profile.c
ZoltanFridrich: probably yes.

p11-kit/delete-profile.c
ZoltanFridrich: In such case why do we even have an option --profile when the option is mandatory and you cant specify more then one. The command might as well just take two arguments <profile> and <token>

ZoltanFridrich:
> Can we add tests in a similar manner to list-profiles?

I don't think so. The mock modules are loaded when the command is executed and the artificial profile is created. The problem is that I cant check with list-profiles whether the profile has been deleted because that command loads the mock modules again and therefore creates the artificial profile again.

ueno:
> I don't think so. The mock modules are loaded when the command is executed and the artificial profile is created. The problem is that I cant check with list-profiles whether the profile has been deleted because that command loads the mock modules again and therefore creates the artificial profile again.

Yeah, I also can't think of any easy way to test it so let's not block on this. Maybe we could have a special mock module backed by a file to load/store profile objects, or perhaps it might be sufficient to simply check `C_DestroyObject` is called with `log-calls: yes` in the module configuration.

ueno:
> I don't think so. The mock modules are loaded when the command is executed and the artificial profile is created. The problem is that I cant check with list-profiles whether the profile has been deleted because that command loads the mock modules again and therefore creates the artificial profile again.

Yeah, I also can't think of any easy way to test it so let's not block on this. Maybe we could have a special mock module backed by a file to load/store profile objects, or perhaps it might be sufficient to simply check `C_DestroyObject` is called with `log-calls: yes` in the module configuration.

Issue #495, opened 2023-04-17
p11-kit 0.24.1 fails assertions


mouse07410:
Apple Silicon M2, MacOS Ventura 13.3.1, Xcode-14.3, OpenSSL-3.1.0, pkcs11-provider (current master), p11-kit v0.24.1, installed via Macports.

```sh
$ openssl pkeyutl  -decrypt -inkey "pkcs11:id=%03;type=private" -pkeyopt rsa_padding_mode:oaep -pkeyopt rsa_oaep_md:sha384 -pkeyopt rsa_mgf1_md:sha384 -in /tmp/derive.25519.key.enc -out /tmp/derive.25519.key.dec
Enter pass phrase for PKCS#11 Token (Slot 17 - Yubico YubiKey OTP+FIDO+CCID):
p11-kit: 'bound != NULL' not true at fixed4_C_CloseSession
p11-kit: 'bound != NULL' not true at fixed4_C_CloseSession
p11-kit: 'bound != NULL' not true at fixed4_C_CloseSession
p11-kit: 'bound != NULL' not true at fixed4_C_CloseSession
p11-kit: 'bound != NULL' not true at fixed4_C_CloseSession
p11-kit: 'bound != NULL' not true at fixed4_C_Finalize
```

Despite the above bunch of failed  assertions, apparently the decryption succeeds and produces correct output.

Still, these assertions are rather unnerving, and it would be great to fix their cause or get rid of them. ;-)

### Update
Cloning this repo, building, and running tests (as your README suggests), showed a few failed tests:
```text
18/57 test-modules             FAIL            1.75s   exit status 12
>>> MALLOC_PERTURB_=12 /Users/ur20980/src/p11-kit/_build/p11-kit/test-modules

19/57 test-init                FAIL            1.97s   killed by signal 6 SIGABRT
>>> MALLOC_PERTURB_=190 /Users/ur20980/src/p11-kit/_build/p11-kit/test-init

20/57 test-deprecated          FAIL            1.62s   killed by signal 6 SIGABRT
>>> MALLOC_PERTURB_=11 /Users/ur20980/src/p11-kit/_build/p11-kit/test-deprecated

21/57 test-proxy               FAIL            1.59s   killed by signal 6 SIGABRT
>>> MALLOC_PERTURB_=36 /Users/ur20980/src/p11-kit/_build/p11-kit/test-proxy

22/57 test-proxy3              FAIL            1.60s   exit status 48
[...]

PR #505, opened 2023-05-07
libffi: Fix handling of pulCount parameter of C_GetInterfaceList

ueno:
When `C_GetInterfaceList` is called with NULL `pInterfaceList`, It doesn't make any sense to check the content of `pulCount`.

Fixes: #495


ZoltanFridrich (approved):
From the specification:
> If pInterfaceList is NULL_PTR, then all that C_GetInterfaceList does is return (in *pulCount)  the number of interfaces, without actually returning a list of interfaces.  The contents of  *pulCount on entry to C_GetInterfaceList has no meaning in this case, and the call returns the value CKR_OK.
Makes sense. Changes look good.

PR #509, opened 2023-05-24
Make terminal output nicer

### LINKED ISSUE #393: Better terminal output for p11-kit and trust commands
When running under a terminal (i.e., `isatty(0)` returns 1), those tools could produce better output, using control sequences, such as Emojis, bold, and [OSC 8](https://bugzilla.gnome.org/show_bug.cgi?id=779734) (as in `ls --hyperlink`).

ZoltanFridrich:
Closes #393


p11-kit/lists.c
```diff
@@ -119,35 +120,49 @@ print_token_info (CK_FUNCTION_LIST_PTR module, CK_SLOT_ID slot_id)
...
-	printf ("    token: %s\n", value);
+	printf (isatty_out ?
+		"    \033[32;1mtoken:\033[0m %s\n" :
+		"    token: %s\n", value);
 	free (value);
 
 	value = p11_kit_space_strdup (info.manufacturerID, sizeof (info.manufacturerID));
-	printf ("        manufacturer: %s\n", value);
+	printf (isatty_out ?
+		"        \033[1mmanufacturer:\033[0m %s\n" :
```
ueno, line 130: To avoid repetition, I would suggest creating a helper function something like:
```c
enum p11_print_level { ... };
static const char *level_to_sgr[] = { "\033[1m" };
#define SGR_RESET "\033[0m"

static void
p11_print_value (FILE *fp, const char *name, const char *value, size_t indent, enum p11_print_level level)
{
  // print indent

  if (isatty (fileno (fp)))
    // print SGR designation
  fprintf (fp, "%s:", name);
  if (isatty (fileno (fp)))
    // print SGR reset
  fprintf (fp, " %s\n", value);
}
```

ZoltanFridrich: I have refactored the code and used helper functions, for now I am keeping the refactoring in a separate commit but I will squash them afterward.

common/print.c
```diff
@@ -0,0 +1,128 @@
...
+		fprintf (fp, "\e]8;;%s\e\\\033[36m%s\033[0m\e]8;;\e\\\n", string, string);
+	else
+		fputs (string, fp);
+}
+
+void
+p11_print_word (FILE *fp,
+		const char *string,
+		int n_styles,
+		...)
```
ueno, line 73: Using varargs to just enumerate styles seems like a bit overkill to me. If we do this I would at least omit `n_styles` and let the caller to add a terminator, e.g., `p11_print_word (fp, string, P11_PRINT_BOLD, P11_PRINT_BLACK, P11_PRINT_INVALID)`.

Or, given only a single color would be specified, it could be turned into a flag:
```c
p11_print_word (fp, string, P11_PRINT_BLACK, P11_PRINT_BOLD|P11_PRINT_UNDERLINE);
```

trust/list.c
```diff
@@ -158,7 +159,8 @@ list_iterate (p11_enumerate *ex,
 				p11_buffer_init (&buf, 1024);
 				bytes = attr->pValue;
 				p11_url_encode (bytes, bytes + attr->ulValueLen, "", &buf);
-				printf ("    public-key-info: %.*s\n", (int)buf.len, (char *)buf.data);
+				p11_print_value (stdout, "public-key-info:", NULL, 4, 1, P11_PRINT_BOLD);
+				printf (" %.*s\n", (int)buf.len, (char *)buf.data);
```
ueno, line 163: For this, I would rather use varargs, calling `vfprintf` internally: `p11_printf_value (stdout, "public-key-info", 4, P11_PRINT_BOLD, "%.*s", buf.len, buf.data)`.

common/print.h
```diff
@@ -0,0 +1,75 @@
...
+#define P11_PRINT_H_
+
+#include <stdio.h>
+
+#define P11_PRINT_VALUE_BOLD(INDENT, KEY, VALUE) \
+	p11_print_value (stdout, INDENT, KEY, P11_COLOR_DEFAULT, P11_FONT_BOLD, " %s\n", VALUE)
+
+#define P11_FONT_DEFAULT   0
+#define P11_FONT_BOLD      1<<0
+#define P11_FONT_UNDERLINE 1<<1
```
ueno, line 45: Why not using enum?

common/print.c
```diff
@@ -0,0 +1,108 @@
...
+		 ...)
+{
+	size_t i;
+	va_list args;
+
+	for (i = 0; i < indent; ++i)
+		fputc (' ', fp);
+
+	p11_print_word (fp, key, color, font);
+	p11_print_word (fp, ":", color, font);
```
ueno, line 103: What about always put a space after a colon (and possibly print a newline), so the caller can call this function with "%s" instead of " %s\n"?

ueno (approved):
I might add some minor adjustment later, but overall it looks good to me.

common/print.h
ZoltanFridrich: I thought that when I use `P11_FONT_BOLD | P11_FONT_UNDERLINE` than it creates invalid enum member and that it would be a problem, but it looks like the compiler does not complain.

PR #514, opened 2023-06-02
RPC module: use space string in C_InitToken

### LINKED ISSUE #497: RPC module: `InitToken` `pLabel` parameter differs from spec
According to [5.5.7 C_InitToken](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/cs01/pkcs11-spec-v3.1-cs01.html#_Toc111203268): 

```
CK_DECLARE_FUNCTION(CK_RV, C_InitToken)(
      CK_SLOT_ID slotID,
      CK_UTF8CHAR_PTR pPin,
      CK_ULONG ulPinLen,
      CK_UTF8CHAR_PTR pLabel
);
```

> `C_InitToken` initializes a token. `slotID` is the ID of the token’s slot; `pPin` points to the SO’s initial PIN (which need not be null-terminated); `ulPinLen` is the length in bytes of the PIN; `pLabel` points to the 32-byte label of the token (which MUST be padded with blank characters, and which MUST NOT be null-terminated). This standard allows PIN values to contain any valid UTF8 character, but the token may impose subset restrictions.

From the above description, the `pLabel` is a 32-byte space padded string. However, the ~~proxy~~ RPC module seems to be using the `zero_string` functions rather than `space_string` functions for the `pLabel`

ZoltanFridrich:
Closes #497


configure.ac
```diff
@@ -20,7 +20,7 @@ P11KIT_AGE=3
 # ------------------------------------------------------------------------------
 # p11-kit RPC protocol versions
 P11KIT_RPC_MIN=0
-P11KIT_RPC_MAX=0
+P11KIT_RPC_MAX=1
```
ueno, line 23: Please also update `meson_options.txt`.

p11-kit/test-filter.c
```diff
@@ -138,10 +138,10 @@ test_allowed (void)
 	rv = (module->C_GetMechanismInfo) (slots[0], mechs[0], &mech);
 	assert_num_eq (CKR_OK, rv);
 
-	rv = (module->C_InitToken) (99, (CK_UTF8CHAR_PTR)"TEST PIN", 8, (CK_UTF8CHAR_PTR)"TEST LABEL");
+	rv = (module->C_InitToken) (99, (CK_UTF8CHAR_PTR)"TEST PIN", 8, (CK_UTF8CHAR_PTR)"TEST LABEL                      ");
```
ueno, line 141: Should we guard this with:
```c
#if P11_RPC_PROTOCOL_VERSION_MAXIMUM == 1
...
#endif
```

ZoltanFridrich: it shouldnt be necessary as the strings are both NULL terminated and space padded

ueno (approved):

PR #516, opened 2023-06-09
Make the proxy_module extension in pkg-config .pc file adaptable

### LINKED ISSUE #515: pkg-config reports wrong proxy_module on MacOS
Running `pkg-config --variable proxy_module p11-kit-1` should give the name of the p11-kit-proxy provider DSO.

It seems to be hard-coded to use the `.so` prefix though, not `.dylib` on MacOS as it should be. (Or `.dll` if/when we support building on Windows)

Spotted as it was being worked around in https://github.com/smallstep/crypto/pull/260

ZoltanFridrich:
Closes #515

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -624,6 +624,9 @@ AC_SUBST(SHLEXT)
 
 AC_DEFINE_UNQUOTED(EXEEXT, ["$EXEEXT"], [File extension for executables])
 
+p11_shlext=$SHLEXT
+AC_SUBST(p11_shlext)
+
 privatedir='${libexecdir}/p11-kit'
 AC_SUBST(privatedir)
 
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -48,6 +48,8 @@ endif
 conf.set_quoted('SHLEXT', shlext)
 conf.set_quoted('EXEEXT', exeext)
 
+p11_shlext = shlext
```
ueno, line 51: I think this is not sufficient as Meson doesn't use `p11-kit-1.pc.in` as a template. Check how `p11_kit_pc_variables` is defined.

ZoltanFridrich: fixed

ueno (approved):

PR #520, opened 2023-06-21
Add test for pkcs#11 3.0 client support

ZoltanFridrich:


ueno (approved):
LGTM, but probably we need to change `p11-kit/test-server.c` to cover PKCS#11 3.0 interface.

ZoltanFridrich:
I think I have goofed this quite bad. I have overwritten my branch by another commits. Closing this MR.

PR #521, opened 2023-06-22
Handle both XSI and GNU versions of strerror_r

### LINKED ISSUE #336: configure doesn't detect strerror_r() in uClibc
In GNU-ish systems, there are two versions of `strerror_r()`: the standard version and the GNU version. In a particular  target system with uClibc (read-only), the technique used by `autoconf` to verify that a system function is available fails because the identifier `strerror_r` is not an entry-point in the C library: instead, it is defined through `<string.h>` to either `__xpg_strerror_r` or `__glibc_strerror_r` depending on feature macros. `autoconf`'s test program doesn't include `<string.h>`, which is good in one way, as that would result in incompatible declarations, but of course equally its reference to `strerror_r()` is invalid.

The `autoconf` manual says that its underlying strategy is that programs should be written using standard APIs and the role of `autoconf` is to discover which used APIs are missing so that they can be shimmed. However the approach (as implemented by the function check macro in the package) fails since each API is only meaningful in the context of its mandated header file(s).

Because of the conflicting definitions of `strerror_r()`, it is recommended to [avoid using it](http://www.club.cc.cmu.edu/~cmccabe/blog_strerror.html), so avoiding the above problem.

ZoltanFridrich:
Closes #336


common/compat.c
```diff
@@ -1037,6 +1038,33 @@ fdwalk (int (* cb) (void *data, int fd),
...
+	intptr_t r = 0;
+
+	assert (buf != NULL);
+	assert (buflen != 0);
+
+	buf[0] = 0xff;
+	errno = 0;
+	r = (intptr_t)strerror_r (errnum, buf, buflen);
+
+	if (r == 0) // XSI success, GNU should never return NULL
```
ueno, line 1055: IMO, it would be better to check the behavior difference at compile time at run time. For Meson you could use [compiler.run](https://mesonbuild.com/Compiler-properties.html#compile-and-run-test-application), for Autotools maybe [AC_RUN_IFELSE](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/Runtime.html) could be used.

ueno (approved):
LGTM

PR #524, opened 2023-06-26
Fix memory leak inside p11_kit_remote_serve_tokens

### LINKED ISSUE #522: Memory leak inside p11_kit_remote_serve_tokens
```console
==396953==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 720 byte(s) in 1 object(s) allocated from:
#0 0x7f3c3fcd8cd7 in calloc (/lib64/libasan.so.8+0xd8cd7) (BuildId: dc689b05ca2577037af24700212bb5cce1f91c8a)
#1 0x5e9a30 in p11_kit_remote_serve_tokens ../p11-kit/rpc-server.c:2758
#2 0x405f3e in main ../p11-kit/remote.c:133
#3 0x7f3c3fa49b49 in __libc_start_call_main (/lib64/libc.so.6+0x27b49) (BuildId: 245240a31888ad5c11bbc55b18e02d87388f59a9)
#4 0x7f3c3fa49c0a in __libc_start_main_alias_2 (/lib64/libc.so.6+0x27c0a) (BuildId: 245240a31888ad5c11bbc55b18e02d87388f59a9)
#5 0x405834 in _start (/home/zfridric/upstream/p11-kit/_build/p11-kit/p11-kit-remote-testable+0x405834) (BuildId: 3207fed024af4b90b1d46ed26781ab2ccf8521d9)

SUMMARY: AddressSanitizer: 720 byte(s) leaked in 1 allocation(s).
```
https://github.com/p11-glue/p11-kit/blob/master/p11-kit/rpc-server.c#L2758

ZoltanFridrich:
The address sanitizer reports a memory leak in the log when running test-server. At first I thought it was related to the test-server.sh failure but it didn't fix the test. The test-server.sh fails immediately after running p11tool and the log does not say anything.

This patch fixes the memory leak reported in the log even though the address sanitizer does not seem to care about it.

Closes #522

### DIFF
--- a/p11-kit/rpc-server.c
+++ b/p11-kit/rpc-server.c
@@ -2761,7 +2761,7 @@ p11_kit_remote_serve_tokens (const char **tokens,
 				goto out;
 			}
 			p11_virtual_init (lower, &p11_virtual_base, module, NULL);
-			filter = p11_filter_subclass (lower, NULL);
+			filter = p11_filter_subclass (lower, free);
 			if (filter == NULL) {
 				error = EINVAL;
 				p11_message_err (error, "couldn't subclass filter");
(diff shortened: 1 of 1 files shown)


p11-kit/rpc-server.c
```diff
@@ -2761,7 +2761,7 @@ p11_kit_remote_serve_tokens (const char **tokens,
 				goto out;
 			}
 			p11_virtual_init (lower, &p11_virtual_base, module, NULL);
-			filter = p11_filter_subclass (lower, NULL);
+			filter = p11_filter_subclass (lower, free);
```
ueno, line 2764: Isn't it sufficient to supply `free` as the `lower_destroyer` parameter of `p11_virtual_init` above?

ZoltanFridrich: Just tried. It is not.

PR #529, opened 2023-07-02
Improve GNU/XSI strerror_r check

atsampson:
The new test that was added to distinguish GNU/XSI strerror_r ran a compiled program, which doesn't work when cross-compiling. The only difference at compile time is that the GNU version returns char * and the XSI version returns int, so detect it by compiling a program that dereferences the return value.

Testing this revealed that the autoconf and meson version of the checks produce different results, because the meson version doesn't define _GNU_SOURCE (even though the rest of the code is compiled with it) - the second commit fixes that.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -146,19 +146,14 @@ if test "$os_unix" = "yes"; then
 
 	AC_CHECK_FUNC(
 		[strerror_r],
-		[AC_RUN_IFELSE(
-			[AC_LANG_SOURCE([[
-				#include <errno.h>
-				#include <string.h>
-
-				int main (void)
-				{
-					char buf[32];
-					return strerror_r (EINVAL, buf, 32);
-				}
-			]])],
-                        [AC_DEFINE([HAVE_XSI_STRERROR_R], 1, [Whether XSI-compliant strerror_r() is available])],
+		[AC_COMPILE_IFELSE(
+			[AC_LANG_PROGRAM([[#include <errno.h>
+					   #include <string.h>]],
+					 [[/* GNU strerror_r returns char *, XSI returns int */
+					    char buf[32];
+					    return *strerror_r (EINVAL, buf, 32);]])],
 			[AC_DEFINE([HAVE_GNU_STRERROR_R], 1, [Whether GNU-specific strerror_r() is available])],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -146,19 +146,20 @@ if test "$os_unix" = "yes"; then
 
 	AC_CHECK_FUNC(
 		[strerror_r],
-		[AC_RUN_IFELSE(
+		[AC_COMPILE_IFELSE(
 			[AC_LANG_SOURCE([[
```
ueno, line 150: It would be slightly simpler to write this with [`AC_LANG_PROGRAM`](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/Generating-Sources.html#index-AC_005fLANG_005fPROGRAM).

meson.build
```diff
@@ -301,20 +301,21 @@ endforeach
 
 if cc.has_function('strerror_r', prefix: '#include <string.h>')
   strerror_r_code = '''
+#define _GNU_SOURCE
```
ueno, line 304: Is this necessary, given that we supply `-D_GNU_SOURCE` through `add_project_arguments`?

ueno (approved):
Looks good to me.

configure.ac
atsampson: OK, I've rewritten it with `AC_LANG_PROGRAM`.

meson.build
atsampson: Yes - note in [the documentation for add_project_arguments](https://mesonbuild.com/Adding-arguments.html), it says <q>Compilation tests that are run as part of your project configuration do not use these flags.</q>

Using current master on my machine with GCC 13.1 and glibc 2.37, Meson builds this test with ```cc .../testfile.c -o /home/ats/src/ .../output.exe -D_FILE_OFFSET_BITS=64 -O0```, so it detects the XSI version, but then goes on to build the project itself with `-D_GNU_SOURCE`, so the code that uses `strerror_r` will be wrong if built with Meson as it stands.

ueno: I see, thank you.

ueno:
Thank you for the PR!

PR #532, opened 2023-07-18
p11-kit: extend the list-modules sub-command of p11-kit cmd tool to print URIs for modules and tokens

ZoltanFridrich:
Adds URIs into list-modules command output.
```
$ p11-kit list-modules
module: ...
  uri: ...
  ...
  token: ...
    uri: ...
```


ueno (approved):
LGTM

PR #533, opened 2023-07-18
Add command to list objects of a token

ZoltanFridrich:
The following is the example output from p11-kit list-objects:
`$ p11-kit list-objects 'pkcs11:token=MyToken'`
By default, the results would only contain objects visible without authentication. To start an authenticated session before listing, the pin-value attribute could be added to the PKCS#11 URI.
`$ p11-kit list-objects 'pkcs11:token=MyToken?pin-value=123456'`


ueno (approved):
Nice, looks good to me

Issue #355, opened 2021-02-14
static library from p11-kit


bulbalion:
Hello,
I'm trying to build elinks-0.14 with gnutls statically. The reason is GNU license. I would like to use elinks for a project and static binary would be nice.
When I try to compile it I run into error due to p11-kit as follows:
/usr/bin/ld: cannot find -lp11-kit
It's compile using the  -static gcc option.
I tried to download p11-kit-0.23.22 and compile it using:
./configure --enable-static
and I run into:
configure: error: p11-kit cannot be used as a static library
Is that intentional behavior or bug?
Thank You & take care and thank You for Your work

ueno:
I'm afraid to say it's intentional. p11-kit uses ELF constructor for library initialization and that wouldn't work otherwise. Although it is possible to expose an explicit initialization function as part of the API (currently used in tests but hidden), it would require the calling sites (e.g., gnutls) to call that function explicitly.

bulbalion:
Thank You very much. That answers my question.

PR #538, opened 2023-08-03
Enable static compilation of libp11-kit

hugelgupf:
I'm not a meson expert, but this enables static compilation if using the `default_library` feature in meson, as documented [here](https://mesonbuild.com/Build-targets.html) and [here](https://mesonbuild.com/Build-targets.html).

```
meson --default-library=static _build 
meson compile -C _build
```

produces static library `_build/p11-kit/libp11-kit.a`

(for some reason, the first command cannot be `meson setup --default-library=static -C build` -- I don't understand how meson parses the default library args, but something's messed up there. anyway.)

### DIFF
--- a/p11-kit/meson.build
+++ b/p11-kit/meson.build
@@ -63,20 +63,20 @@ libp11_kit_ldflags += cc.get_supported_link_arguments([
   '-Wl,-z,nodelete'
 ])
 
-libp11_kit = shared_library('p11-kit',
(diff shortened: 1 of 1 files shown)


hugelgupf:
Heh, I found #355. I admit I haven't used this yet, I just got stuck trying to compile QEMU statically, which depends on p11-kit. So apparently that's just irreparably broken for the time being??

PR #540, opened 2023-08-14
virtual: Factor out code generation templates

ueno:
This makes the current code generation script more generic by factoring out the templates into a separate file.


p11-kit/gen-fixed-closures.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/python
```
neverpanic, line 1: Should we use `/usr/bin/env python` to support distributions which don't have it in `/usr/bin`, do autodetection of a Python interpreter in configure?

Should we also use `python3` in the shebang? You are relying on f-strings, [which are a Python 3.6+ feature](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals) anyway. [PEP 394](https://peps.python.org/pep-0394/#for-python-script-publishers) says this:

> In cases where the script is expected to be executed outside virtual environments, developers will need to be aware of the following discrepancies across platforms and installation methods:
> 
> - Older Linux distributions will provide a python command that refers to Python 2, and will likely not provide a python2 command.
> - Some newer Linux distributions will provide a python command that refers to Python 3.
> - Some Linux distributions will not provide a python command at all by default, but will provide a python3 command by default.

p11-kit/gen-fixed-closures.py
```diff
@@ -0,0 +1,46 @@
...
+        output.write(f"""\
+P11_VIRTUAL_FIXED_GET_FUNCTION_LIST({i})
+P11_VIRTUAL_FIXED_GET_INTERFACE_LIST({i})
+P11_VIRTUAL_FIXED_GET_INTERFACE({i})
+""")
+    return output.getvalue()
+
+
+if __name__ == "__main__":
+    import argparse
```
neverpanic, line 37: Any reason why this isn't a global import? Are you planning to support using this as a module?

p11-kit/gen-fixed-closures.py
```diff
@@ -0,0 +1,46 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--closures", type=int,
+                        default=64)
+    parser.add_argument("--infile", type=argparse.FileType("r"))
```
neverpanic, line 41: The `--infile` argument seems to be unused.

p11-kit/gen-wrappers.py
```diff
@@ -0,0 +1,145 @@
...
+    wrappers = separator.join([
+        emit_wrapper_function(
+            function,
+            templates,
+            args.concat_lines,
+        )
+        for function in functions if function["name"] not in excludes
+    ])
+
+    separator = ", \\\n" if args.concat_lines else ",\n"
```
neverpanic, line 122: Same here, I'd expect `",\\\n"` as separator to avoid the trailing space before the newline.

neverpanic (commented):
Can you elaborate why you chose python source files parsed by the ast module to store the templates, and what advantage that provides over, for example, using YAML, TOML, or JSON?

p11-kit/gen-fixed-closures.py
ueno: I was told [not](https://jmmv.dev/2016/09/env-considered-harmful.html) to use `/usr/bin/env python` at shebang if we want to enforce a specific version of Python interpreter. Anyway, the python version is checked in the top-level [meson.build](https://github.com/p11-glue/p11-kit/blob/91bec95e9ddbf9ba89dce13513e939bebd92ffce/meson.build#L438).

neverpanic: That advise does apply for a typical Linux system, but it'll fail for example on macOS with MacPorts or Homebrew, since `/usr/bin/python` no longer exists on macOS Ventura (only `/usr/bin/python3` does), and that Python may be older than what other instances installed in `$PATH` are.

I'd agree with this advice if there was anything in your scripts that required modules or a specific version of Python, but from what I've seen you only use builtin modules, so any installation >= 3.6 should work.

It doesn't necessarily have to be `/usr/bin/env python3`, but at least `/usr/bin/python` will get you bug reports that the build doesn't work on modern macOS almost immediately.

Do you not still support the autoconf build system, too? In that case, does that also check the Python version?

ueno:
> Can you elaborate why you chose python source files parsed by the ast module to store the templates, and what advantage that provides over, for example, using YAML, TOML, or JSON?

Since the template values are directly consumed by `str.format()`, I thought it would be the most natural to use Python literals. The other formats also have drawbacks: YAML (some characters needs different escaping, e.g., a colon), TOML (the parser is part of the standard library only after 3.11), and JSON (no possibility to add comments).

p11-kit/gen-wrappers.py
```diff
@@ -0,0 +1,145 @@
...
+
+import ast
+import json
+import sys
+
+INDENT = "        "
+
+
+def emit_wrapper_function(function, templates, concat_lines=False):
+    arglist_separater = ", \\\n" if concat_lines else ",\n"
```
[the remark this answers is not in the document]
ueno: Typo: fixed.

As for the space after the comma: the intended use-case of this is to emit macros, e.g.,
```c
static CK_RV \
fixed ## fixed_index ## _C_GetSlotList (unsigned char token_present, \
                                        CK_SLOT_ID_PTR slot_list, \
                                        long unsigned int * ulCount) \
{ \
...
} \
```
Without a space, we would have to rely on the next line starts with an indent

p11-kit/gen-fixed-closures.py
ueno: > Do you not still support the autoconf build system, too? In that case, does that also check the Python version?

Good point, added a check through `AM_PATH_PYTHON`.

p11-kit/gen-fixed-closures.py
```diff
@@ -0,0 +1,46 @@
...
+    parser.add_argument("--outfile", type=argparse.FileType("w"),
+                        default=sys.stdout)
+    args = parser.parse_args()
+
+    args.outfile.write(fixed_instantiate(args.closures))
```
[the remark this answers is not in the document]
ueno: Thanks; yes, we can directly write to outfile here. Fixed.

ZoltanFridrich (approved):
The changes look fine by me. The threads have been addressed except for one.

p11-kit/gen-fixed-closures.py
ueno: Not for now, but it would make future extension easier I guess.

ueno:
Thank you for the reviews!

Issue #539, opened 2023-08-13
0.25.0 breaks gnutls build


lazka:
it looks like something regarding d3d333f9aa037badd360036da3766e9f58a45dbc

gcc fails with:

```
 In file included from C:/_/B/src/gnutls-3.8.1/lib/pkcs11_int.h:27,
                   from ../../../gnutls-3.8.1/lib/x509/verify.c:37:
  D:/M/msys64/mingw64/include/p11-kit-1/p11-kit/pkcs11.h:1445:46: error: expected ';', ',' or ')' before 'struct'
   1445 |                        struct ck_interface **interface,
        |                                              ^~~~~~~~~
  D:/M/msys64/mingw64/include/p11-kit-1/p11-kit/pkcs11.h:1160:32: note: in definition of macro '_CK_DECLARE_FUNCTION'
   1160 | typedef ck_rv_t (*CK_ ## name) args;            \
        |                                ^~~~
  D:/M/msys64/mingw64/include/p11-kit-1/p11-kit/pkcs11.h:1445:46: error: expected ';', ',' or ')' before 'struct'
   1445 |                        struct ck_interface **interface,
        |                                              ^~~~~~~~~
  D:/M/msys64/mingw64/include/p11-kit-1/p11-kit/pkcs11.h:1161:22: note: in definition of macro '_CK_DECLARE_FUNCTION'
   1161 | ck_rv_t CK_SPEC name args
        |                      ^~~~
  D:/M/msys64/mingw64/include/p11-kit-1/p11-kit/pkcs11.h:1660:3: error: unknown type name 'CK_C_GetInterface'
   1660 |   CK_C_GetInterface C_GetInterface;
        |   ^~~~~~~~~~~~~~~~~
  In file included from C:/_/B/src/gnutls-3.8.1/lib/pkcs11_int.h:27,
                   from ../../../gnutls-3.8.1/lib/x509/x509.c:38:
  D:/M/msys64/mingw64/include/p11-kit-1/p11-kit/pkcs11.h:1445:46: error: expected ';', ',' or ')' before 'struct'
   1445 |                        struct ck_interface **interface,
        |                                              ^~~~~~~~~
  D:/M/msys64/mingw64/include/p11-kit-1/p11-kit/pkcs11.h:1160:32: note: in definition of macro '_CK_DECLARE_FUNCTION'
   1160 | typedef ck_rv_t (*CK_ ## name) args;            \
        |                                ^~~~
[...]

PR #541, opened 2023-08-14
common: Avoid naming clash on MinGW

ueno:
The symbol "interface" is defined differently in a header included by default on MinGW, which prevents the use of it in a function signature in pkcs11.h.  To mitigate that, this renames it to "interface_".

In the source there are more places where "interface" is used, though it shouldn't be a problem as WIN32_LEAN_AND_MEAN is set in compat.h.

Fixes: #539


lazka:
I can confirm that this fixes the issue.

ueno:
Thank you for checking!

PR #543, opened 2023-08-24
Add command to export objects matching PKCS#11 URI

ZoltanFridrich:
Adds the ability to print X.509 certificate objects format to stdout.
`p11-kit export-object pkcs11:objectUri`


ueno (commented):
I would love to see some tests to prove the behavior of this command. I remember we didn't add tests for other commands like `delete-object`, just because they have side effects and it is not easy to test (though still possible by e.g., using a temporary file as storage). Here, `export-object` does not have any side-effect, you could simply run this against a mock module and compare the expected output in a shell script.

ZoltanFridrich:
> 

Yes, but I could add that in a separate MR.

ueno (approved):
LGTM!

PR #544, opened 2023-08-28
Allow all types of URIs to be used in p11-kit commands

ZoltanFridrich:
This makes it possible to specify an object with the URI in commands like p11-kit delete-object.


p11-kit/delete-object.c
```diff
@@ -70,7 +70,7 @@ delete_object (const char *token_str)
 		goto cleanup;
 	}
 
-	if (p11_kit_uri_parse (token_str, P11_KIT_URI_FOR_TOKEN, uri) != P11_KIT_URI_OK) {
+	if (p11_kit_uri_parse (token_str, P11_KIT_URI_FOR_ANY, uri) != P11_KIT_URI_OK) {
```
ueno, line 73: Why not `P11_KIT_URI_FOR_OBJECT_ON_TOKEN`? ANY also matches tokens/modules which is not very useful in this context.

ZoltanFridrich: I thought that it might be good to allow deleting all objects across multiple tokens etc. I can change that.

ueno: That is what `P11_KIT_URI_FOR_OBJECT_ON_TOKEN` means. PKCS#11 can also be used to match against tokens or modules, which are not objects, and thus irrelevant for these commands.

p11-kit/list-objects.c
```diff
@@ -170,9 +236,10 @@ print_object (p11_list_printer *printer,
 	CK_MECHANISM_TYPE mechanism_type;
 	CK_BBOOL trusted, local, token, private, modifiable, copyable, destroyable;
 	CK_DATE start_date, end_date;
-	char label[128] = { '\0' };
-	char application[128] = { '\0' };
-	char id[128] = { '\0' };
+	char label[BUFFER_SIZE] = { '\0' };
```
ueno, line 239: Not originated in this PR, but shouldn't we allow those attributes of arbitrary length?

p11-kit/list-objects.c
```diff
@@ -55,10 +57,19 @@
...
 #endif
 
+#define BUFFER_SIZE 128
+
 int
 p11_kit_list_objects (int argc,
 		      char *argv[]);
 
+static inline bool
+is_storage_object (CK_OBJECT_CLASS klass)
```
ueno, line 67: Just out of curiosity: what are the criteria between storage and non-storage objects? If the former means objects with exportable data, I would name this function as "is_exportable_object".

ZoltanFridrich: From PKCS #11 specification version 3.1: "This is not an object class; hence no CKO_ definition is required. It is a category of object classes with common attributes for the object classes that follow." and the object classes that follow are the same that I have listed in the function.

Also the URI can only be specified for storage objects, so I would rather keep the function named is_storage_object.

p11-kit/list-objects.c
ZoltanFridrich: We probably should, yes.

ZoltanFridrich: done

p11-kit/add-profile.c
```diff
@@ -161,9 +161,11 @@ add_profile (const char *token_str,
 
 cleanup:
 	p11_kit_iter_free (iter);
-	p11_kit_modules_finalize (modules);
-	p11_kit_modules_release (modules);
 	p11_kit_uri_free (uri);
+	if (modules != NULL) {
+		p11_kit_modules_finalize (modules);
+		p11_kit_modules_release (modules);
```
ueno, line 167: `p11_kit_modules_finalize_and_release` might be handier?

p11-kit/list-objects.c
```diff
@@ -43,8 +43,10 @@
 #include "message.h"
 #include "print.h"
 #include "tool.h"
+#include "uri.h"
 
 #include <assert.h>
+#include <stdbool.h>
```
ueno, line 49: Isn't this already included through "compat.h", which also provide fallback definitions of true / false?

p11-kit/list-objects.c
```diff
@@ -170,9 +235,17 @@ print_object (p11_list_printer *printer,
 	CK_MECHANISM_TYPE mechanism_type;
 	CK_BBOOL trusted, local, token, private, modifiable, copyable, destroyable;
 	CK_DATE start_date, end_date;
-	char label[128] = { '\0' };
-	char application[128] = { '\0' };
-	char id[128] = { '\0' };
+	char *uri_str;
+	size_t i;
+
+	CK_ATTRIBUTE arrs[] = {
```
ueno, line 241: `arrs` is not very intuitive acronym I would say.

p11-kit/list-objects.c
```diff
@@ -202,27 +270,59 @@ print_object (p11_list_printer *printer,
 	};
 	CK_ULONG n_attrs = sizeof (attrs) / sizeof (attrs[0]);
 
+	/* get the attributes */
+	p11_kit_iter_get_attributes (iter, arrs, n_arrs);
+	for (i = 0; i < n_arrs; ++i) {
+		if (arrs[i].ulValueLen == CK_UNAVAILABLE_INFORMATION)
+			continue;
+		arrs[i].pValue = malloc (arrs[i].ulValueLen + 1);
```
ueno, line 278: I guess we can safely assume that ulValueLen accounts for NUL byte (otherwise we would need a overflow check here).

p11-kit/list-objects.c
ZoltanFridrich: Any suggestion? Otherwise I will go with `arr_attrs`.

p11-kit/list-objects.c
ZoltanFridrich: I dont know about that. It feels safer to use the overflow check. I didnt find anywhere that a RFC2279 string would be NULL terminated.

ueno: Yeah, you are right according to the PKCS#11 spec. Let's add a check.

p11-kit/list-objects.c
ueno: I would rather not split the attrs array, but just loop over the elements and if `pValue` is `NULL`, allocate memory as necessary, something like (without error handling):
```c
CK_ATTRIBUTE attrs[] = {
  { CKA_CERTIFICATE_CATEGORY, &cert_category, sizeof (cert_category) },
  ...
  { CKA_LABEL, NULL_PTR, 0 },
  ...
};

/* track allocated memory */
p11_array *allocated = p11_array_new (free);
...
/* first call to determine ulValueLen */
p11_kit_iter_get_attributes (iter, attrs, n_attrs);

for (i = 0; i < n_attrs; i++) {
  if (attrs[i].pValue == NULL && attrs[i].ulValueLen != CK_UNAVAILABLE_INFORMATION) {
    p11_buffer buf;
    size_t len = attrs[i].ulValueLen;

    p11_buffer_init (&buf, len);

    if (... this is a string attribute ...)
      p11_buffer_add (&buf, "\0", 1);

    attrs[i].pValue = p11_buffer_steal (&buf, &len);
    p11_array_push (allocated, attrs[i].pValue);
  }
}

/* second call to actually get the attributes */
p11_kit_iter_get_attributes (iter, arrs, n_arrs);

p11_array_free (allocated);
```

p11-kit/list-objects.c
```diff
@@ -202,8 +266,45 @@ print_object (p11_list_printer *printer,
...
+		return;
+	}
+
+	p11_kit_iter_get_attributes (iter, attrs, n_attrs);
+	for (i = 0; i < n_attrs; ++i) {
+		if (attrs[i].pValue != NULL_PTR ||
+		    attrs[i].ulValueLen == CK_UNAVAILABLE_INFORMATION)
+			continue;
+
+		if (attrs[i].ulValueLen >= SIZE_MAX) {
```
ueno, line 281: `p11_buffer` has the overflow check. What about using it instead of doing that manually?

p11-kit/list-objects.c
ueno: Sorry I couldn't find any reference of the term "storage object" in the specification, and I still don't understand the difference between your definition of storage and non-storage objects. For example, CKO_X_CERTIFICATE_EXTENSION is *stored* on the token, but cannot be exported individually. Is it storage object or non-storage object?

ZoltanFridrich: I have seen the term both here https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/pkcs11-spec-v3.1.pdf and here https://datatracker.ietf.org/doc/html/rfc7512
Thats why I decided to use the term storage object.

p11-kit/list-objects.c
ZoltanFridrich: Because the intent is much clearer and the code is simpler then using the abstraction p11_buffer. Also the example you provided would not work because after calling `p11_buffer_init (&buf, len);` the `buffer.len` would still be `0` and therefore you would first need to fill in the buffers and then call `p11_buffer_add (&buf, "\0", 1);` otherwise you are just adding '\0' at `buf.data[0]`. I dont see the point in using `p11_buffer` in this case.

p11-kit/list-objects.c
ueno: I see; sorry for the oversight.

p11-kit/list-objects.c
ueno: Ah, that's correct; it would be nice if the `p11_buffer` provides an API to reserve some more room, without appending data.

ueno (approved):

PR #547, opened 2023-08-31
Minor refactoring on p11-kit list-objects subcommand

ueno:
This avoids spelling out array indices and having overflow checks in multiple places.

### DIFF
--- a/p11-kit/list-objects.c
+++ b/p11-kit/list-objects.c
@@ -36,6 +36,7 @@
 
 #include "config.h"
 
+#include "buffer.h"
 #include "constants.h"
 #include "debug.h"
 #include "hex.h"
@@ -274,21 +275,24 @@ print_object (p11_list_printer *printer,
 
 	p11_kit_iter_get_attributes (iter, attrs, n_attrs);
 	for (i = 0; i < n_attrs; ++i) {
+		p11_buffer buffer;
+
 		if (attrs[i].pValue != NULL_PTR ||
 		    attrs[i].ulValueLen == CK_UNAVAILABLE_INFORMATION)
 			continue;
 
-		if (attrs[i].ulValueLen >= SIZE_MAX) {
-			p11_message (_("allocation would result in overflow"));
-			goto cleanup;
-		}
-		attrs[i].pValue = malloc (attrs[i].ulValueLen + 1);
-		if (attrs[i].pValue == NULL) {
+		if (!p11_buffer_init_null (&buffer, attrs[i].ulValueLen) ||
+		    !p11_buffer_append (&buffer, attrs[i].ulValueLen)) {
 			p11_message (_("failed to allocate memory"));
 			goto cleanup;
 		}
-		((char *)attrs[i].pValue)[attrs[i].ulValueLen] = '\0';
-		p11_array_push (allocated, attrs[i].pValue);
+
+		attrs[i].pValue = p11_buffer_steal (&buffer, NULL);
+		if (!p11_array_push (allocated, attrs[i].pValue)) {
+			free (attrs[i].pValue);
+			p11_message (_("failed to allocate memory"));
+			goto cleanup;
+		}
 	}
 	p11_kit_iter_get_attributes (iter, attrs, n_attrs);
 
--- a/p11-kit/list-objects.c
+++ b/p11-kit/list-objects.c
@@ -36,6 +36,7 @@
 
 #include "config.h"
 
+#include "attrs.h"
 #include "buffer.h"
 #include "constants.h"
 #include "debug.h"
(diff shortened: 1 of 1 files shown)


common/buffer.c
```diff
@@ -77,8 +77,14 @@ bool
...
                       size_t reserve)
 {
+	bool ok;
+
+	return_val_if_fail (reserve < SIZE_MAX, false);
+
 	p11_buffer_init_full (buffer, NULL, 0, P11_BUFFER_NULL, realloc, free);
-	return buffer_realloc (buffer, reserve);
+	ok = buffer_realloc (buffer, reserve + 1);
+	((unsigned char *)buffer->data)[reserve] = '\0';
```
ZoltanFridrich, line 86: I dont think this is an intuitive behavior. This probably deserves a function documentation comment.

p11-kit/list-objects.c
```diff
@@ -274,21 +275,22 @@ print_object (p11_list_printer *printer,
 
 	p11_kit_iter_get_attributes (iter, attrs, n_attrs);
 	for (i = 0; i < n_attrs; ++i) {
+		p11_buffer buffer;
```
ZoltanFridrich, line 275: I would prefer to have this variable declaration at the start of the function with the rest of the declarations.

p11-kit/list-objects.c
```diff
@@ -274,21 +275,22 @@ print_object (p11_list_printer *printer,
...
+		if (!p11_buffer_init_null (&buffer, attrs[i].ulValueLen)) {
+			p11_message (_("failed to allocate memory"));
 			goto cleanup;
 		}
-		attrs[i].pValue = malloc (attrs[i].ulValueLen + 1);
-		if (attrs[i].pValue == NULL) {
+
+		attrs[i].pValue = p11_buffer_steal (&buffer, NULL);
+		if (!p11_array_push (allocated, attrs[i].pValue)) {
 			p11_message (_("failed to allocate memory"));
```
ZoltanFridrich, line 288: `Attrs[i].pValue` memory is leaked here.

p11-kit/list-objects.c
```diff
@@ -192,16 +189,16 @@ get_object_uri (P11KitIter *iter,
 		goto cleanup;
 	}
 
-	if (label != NULL) {
-	        ret = p11_kit_uri_set_attribute (uri, &attr_label);
+	if (attr_label != NULL) {
```
ZoltanFridrich, line 192: These will always be true in this use case. The idea was to not set attributes which have `pValue == NULL`. I would suggest adding a check for `attr->pValue != NULL` as well. However I dont know if setting a `NULL` value attr matters for URI.

ZoltanFridrich (changes requested):
The `p11_buffer` changes need to be documented somehow as the behavior is not at all intuitive.
The usage of `p11_attrs_findn` looks good but is quite an inefficient solution. However it shouldn't as its still a constant time.

p11-kit/list-objects.c
ueno: I don't think that is beneficial; there are many other projects suggest the opposite (i.e., limit the variable scope, [example](https://github.com/libreswan/libreswan/blob/main/docs/ProgrammingConventions.txt#L80))

p11-kit/list-objects.c
ueno: I think in this case `pValue` is also always non-NULL (because we allocated). I've added `ulValueLen != CK_UNAVAILABLE_INFORMATION` check instead.

common/buffer.c
ueno: In the end I reverted this change and switched to use `p11_buffer_init_null` followed by `p11_buffer_append` (which does append NUL byte as necessary).

p11-kit/list-objects.c
ZoltanFridrich: I was violating this principle all the time then. Good to know

p11-kit/list-objects.c
```diff
@@ -274,56 +272,59 @@ print_object (p11_list_printer *printer,
...
+		p11_buffer buffer;
+
 		if (attrs[i].pValue != NULL_PTR ||
 		    attrs[i].ulValueLen == CK_UNAVAILABLE_INFORMATION)
 			continue;
 
-		if (attrs[i].ulValueLen >= SIZE_MAX) {
-			p11_message (_("allocation would result in overflow"));
+		if (!p11_buffer_init_null (&buffer, attrs[i].ulValueLen) ||
+		    !p11_buffer_append (&buffer, attrs[i].ulValueLen)) {
```
ZoltanFridrich, line 282: I get it now. The `p11_buffer_append` function doesn't actually append data, it just moves the length pointer... :smiling_face_with_tear:

p11-kit/list-objects.c
```diff
@@ -191,16 +189,16 @@ get_object_uri (P11KitIter *iter,
...
+	        ret = p11_kit_uri_set_attribute (uri, attr_label);
 	        if (ret != P11_KIT_URI_OK) {
 		        p11_message (_("failed to set attribute for URI: %s"), p11_kit_uri_message (ret));
 		        goto cleanup;
 	        }
 	}
 
-	if (id != NULL) {
-		ret = p11_kit_uri_set_attribute (uri, &attr_id);
+	if (attr_id->ulValueLen != CK_UNAVAILABLE_INFORMATION) {
```
ZoltanFridrich, line 200: This should be fine as well as the `pValue` should always be allocated in case `ulValueLen` is set

ZoltanFridrich (approved):
OK.

PR #549, opened 2023-09-01
Add ability to export public keys

ZoltanFridrich:
- adds the ability to export public key object to stdout in PEM format
- adds a test for this feature


p11-kit/export-object.c
```diff
@@ -60,12 +60,14 @@ p11_kit_export_object (int argc,
 		       char *argv[]);
 
 static void
-export_x509_certificate (P11KitIter *iter)
+export_attr (P11KitIter *iter,
+	     CK_ATTRIBUTE_TYPE type,
```
ueno, line 64: nit: I would spell `export_attribute` out, as we don't abbreviate `export_certificate` as `export_cert`.

PR #550, opened 2023-09-02
proxy: Generate wrappers with script

ueno:
This also applies code generation to the proxy module.


ZoltanFridrich (approved):
Looks nice.

PR #551, opened 2023-09-06
Add command to generate keypair on a PKCS#11 token

ZoltanFridrich:
When the user builds a solution around PKCS#11, such as setting up an HTTPS server with NGINX or Apache HTTPD, they first would need to create a key pair on a PKCS#11 token, where the private key can be used to sign the certificate and TLS handshake.

The generate-keypair subcommand has been added to generate a private key and the corresponding public key. This takes a PKCS#11 URI of a token, where the generated keys are stored, and a label through the --label option. It can also take other options, such as --type to specify the key type and --bits to specify the key size.

### DIFF
--- a/common/compat.c
+++ b/common/compat.c
@@ -1051,3 +1051,13 @@ p11_ascii_toupper (int c)
 		return 'A' + (c - 'a');
 	return c;
 }
+
+bool
+p11_ascii_strcaseeq (const char *s1,
+		     const char *s2)
+{
+	while (p11_ascii_tolower (*s1) == p11_ascii_tolower (*s2++))
+		if (*s1++ == '\0')
+			return true;
+	return false;
+}
--- a/p11-kit/generate-keypair.c
+++ b/p11-kit/generate-keypair.c
@@ -1,0 +1,457 @@
+/*
+ * Copyright (c) 2023, Red Hat Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the
+ *       following disclaimer.
+ *     * Redistributions in binary form must reproduce the
+ *       above copyright notice, this list of conditions and
+ *       the following disclaimer in the documentation and/or
+ *       other materials provided with the distribution.
+ *     * The names of contributors to this software may not be
+ *       used to endorse or promote products derived from this
+ *       software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * Author: Zoltan Fridrich <zfridric@redhat.com>
+ */
+
+#include "config.h"
+
+#include "attrs.h"
+#include "compat.h"
+#include "debug.h"
+#include "iter.h"
+#include "message.h"
+#include "tool.h"
+
+#ifdef P11_KIT_TESTABLE
+#include "mock.h"
+#endif
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef ENABLE_NLS
+#include <libintl.h>
+#define _(x) dgettext(PACKAGE_NAME, x)
+#else
+#define _(x) (x)
+#endif
+
+int
+p11_kit_generate_keypair (int argc,
+			  char *argv[]);
+
+static CK_MECHANISM
+get_mechanism (const char *type)
+{
+	CK_MECHANISM m = { CKA_INVALID, NULL_PTR, 0 };
+
+#ifdef P11_KIT_TESTABLE
+	if (p11_ascii_strcaseeq (type, "mock")) {
+		m.mechanism = CKM_MOCK_GENERATE;
+		m.pParameter = "generate";
+		m.ulParameterLen = 9;
+		return m;
+	}
+#endif
+	if (p11_ascii_strcaseeq (type, "rsa"))
+		m.mechanism = CKM_RSA_PKCS_KEY_PAIR_GEN;
+	else if (p11_ascii_strcaseeq (type, "ecdsa"))
+		m.mechanism = CKM_ECDSA_KEY_PAIR_GEN;
+	else if (p11_ascii_strcaseeq (type, "ed25519"))
+		m.mechanism = CKM_EC_EDWARDS_KEY_PAIR_GEN;
+
+	return m;
+}
+
+static const uint8_t *
+get_ec_params (const char *curve,
+	       size_t *ec_params_len)
+{
+	static const uint8_t OID_SECP256R1[] = { 0x06, 0x08, 0x2a, 0x86, 0x48, 0xce, 0x3d, 0x03, 0x01, 0x07 };
+	static const uint8_t OID_SECP384R1[] = { 0x06, 0x05, 0x2b, 0x81, 0x04, 0x00, 0x22 };
+	static const uint8_t OID_SECP521R1[] = { 0x06, 0x05, 0x2b, 0x81, 0x04, 0x00, 0x23 };
+
+	if (p11_ascii_strcaseeq (curve, "secp256r1")) {
+		*ec_params_len = sizeof (OID_SECP256R1);
+		return OID_SECP256R1;
+	} else if (p11_ascii_strcaseeq (curve, "secp384r1")) {
+		*ec_params_len = sizeof (OID_SECP384R1);
+		return OID_SECP384R1;
+	} else if (p11_ascii_strcaseeq (curve, "secp521r1")) {
+		*ec_params_len = sizeof (OID_SECP521R1);
+		return OID_SECP521R1;
+	}
+
+	return NULL;
+}
+
+static bool
+check_args (CK_MECHANISM_TYPE type,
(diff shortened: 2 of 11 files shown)


p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,355 @@
...
+#else
+#define _(x) (x)
+#endif
+
+int
+p11_kit_generate_keypair (int argc,
+			  char *argv[]);
+
+static inline int
+p11_strcasecmp (const char *a,
```
ueno, line 62: Let's move this to `common/compat.h` and use `p11_ascii_tolower` instead of `tolower` to avoid issues like https://developers.redhat.com/articles/2022/06/15/openssl-30-dealing-turkish-locale-bug 

We could use `strcasecmp_l` as well.

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,355 @@
...
+static bool
+get_templates (const char *label,
+	       const CK_MECHANISM *m,
+	       const CK_ULONG *bits,
+	       CK_ATTRIBUTE *a,
+	       CK_ULONG *a_len,
+	       CK_ATTRIBUTE *b,
+	       CK_ULONG *b_len)
+{
+	CK_BBOOL tval = 1, fval = 0;
```
ueno, line 99: Using `CK_TRUE` / `CK_FALSE` would be more consistent to read if you use `CK_BBOOL`.

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,355 @@
...
+	m->ulParameterLen = 0;
+
+	return true;
+}
+
+static bool
+get_templates (const char *label,
+	       const CK_MECHANISM *m,
+	       const CK_ULONG *bits,
+	       CK_ATTRIBUTE *a,
```
ueno, line 94: Shouldn't we better name the parameter: `a` → `pubkey`, `b` → `privkey`?

ueno (commented):
I guess one tricky thing is that, if we want to support EC keys, we need to encode ECParamters in ASN.1 and the tool probably need to link to libtasn1, which is currently an optional build dependency for the trust module. I guess some tweaks in the build system is needed, e.g., check libtasn1 is available, and if yes, link p11-kit tool to it; otherwise we shouldn't but we can still support RSA.

common/compat.c
```diff
@@ -1051,3 +1051,13 @@ p11_ascii_toupper (int c)
 		return 'A' + (c - 'a');
 	return c;
 }
+
+int
+p11_strcasecmp (const char *s1,
```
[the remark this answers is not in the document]
ZoltanFridrich: I will add "ascii" to the name, but I am not going to reduce the functionality of the function by making it return bool when its supposed to be a general string comparison function. Otherwise whats the point of having it in compat.h when its supposed to be a single use only? In such case I could have left it local to the generate-keypair.c file.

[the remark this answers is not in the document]
ueno: There is no point in making the internal function as generic as possible, either. If we make it a general string comparison function, it needs to work correctly (e.g., it works based on the ordering axioms as checked by [sortcheck](https://github.com/yugr/sortcheck)); it can be overly complicated and out of scope of our use-case.

Limiting the responsibility of the function is also beneficial to avoid programming errors (e.g., https://github.com/libreswan/libreswan/blob/a4bed3681646f19bc3ee2783f88a3257e92e0659/docs/ProgrammingConventions.txt#L65).

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,373 @@
...
+	m->ulParameterLen = 0;
+
+	return true;
+}
+
+static const char *
+get_curve (const char *curve)
+{
+	if (p11_ascii_strcaseeq (curve, "SECP256R1"))
+		return "\x13\x05P-256";
```
ueno, line 84: This is an encoded PrintableString in ASN.1, right? In that case we should treat it as a binary blob rather than a nul-terminated string (i.e., `uint8_t *` with explicit length, instead of `char *`). Also where are those curve names defined? It might be safer to use OIDs instead as defined in [SEC 2](https://www.secg.org/sec2-v2.pdf#page=33) as typical HSMs would recognize them.

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,412 @@
...
+#include <string.h>
+
+#ifdef ENABLE_NLS
+#include <libintl.h>
+#define _(x) dgettext(PACKAGE_NAME, x)
+#else
+#define _(x) (x)
+#endif
+
+#define cleanup_if_fail(X, S)          \
```
ueno, line 59: Let's not invent these macros similar to `return_if_fail` etc., which [originated](https://developer-old.gnome.org/glib/stable/glib-Warnings-and-Assertions.html) in GNOME to assert pre-conditions (programming error), while in the following code `cleanup_if_fail` and `cleanup_if_fail_rv` are used for regular runtime errors (e.g., token doesn't support key generation); in general I suggest not using macros for such but just spelling out `p11_message` and `goto`.

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,412 @@
...
+		goto cleanup;          \
+	} } while (false)
+
+#define cleanup_if_fail_rv(X, S, R)    \
+	do { if (X) { } else {         \
+		p11_message (_(S), R); \
+		goto cleanup;          \
+	} } while (false)
+
+const CK_BBOOL TVAL = CK_TRUE;
```
ueno, line 71: As `TVAL` and `FVAL` are only used in `get_templates`, can we make these as a local variable? Otherwise `static` is missing.

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,412 @@
...
+
+int
+p11_kit_generate_keypair (int argc,
+			  char *argv[]);
+
+static bool
+get_mechanism (CK_MECHANISM *m,
+	       const char *type)
+{
+	if (p11_ascii_strcaseeq (type, "mock")) {
```
ueno, line 82: Can we conditionalize this `if` at build time (i.e., only compile in when building a `-testable` binary)?

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,412 @@
...
+
+	return true;
+}
+
+static bool
+get_ec_params (const char *curve,
+	       p11_buffer *ec_params)
+{
+	bool ok = true;
+	const char *data = NULL;
```
ueno, line 109: I would suggest declaring it as `const uint8_t *` and:
```c
static const uint8_t OID_secp256r1[] = { 0x06, 0x08, ... };
...
data = OID_secp256r1;
size = sizeof(OID_secp256r1);
```

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,412 @@
...
+	    CK_ULONG bits,
+	    const p11_buffer *ec_params)
+{
+	if (type == CKA_INVALID) {
+		p11_message (_("no type specified"));
+		return false;
+	}
+
+	switch (type) {
+	case CKM_MOCK_GENERATE:
```
ueno, line 139: Let's compile out from a production code.

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,412 @@
...
+			return false;
+		}
+		break;
+	default:
+		p11_message (_("unkwnown mechanism type in %s"), __func__);
+		return false;
+	}
+
+	if (bits != 0 && ec_params->data != NULL) {
+		p11_message (_("both bits and curve cannot be specified at once"));
```
ueno, line 160: Since this message is translated, making this a little more generic would reduce translators burden: `p11_message (_("both %s and %s cannot be specified at once"), "--bits", "--curves")`, so no new message will need to be translated even after we add another options mutually exclusive.

ueno (changes requested):
No test, while `CKM_MOCK_GENERATE` is supported?

p11-kit/generate-keypair.c
ZoltanFridrich: If I make them into local variables inside `get_templates` then after return the addresses become invalid and I don't want to add them as additional arguments to `get_templates`. I will make them `static`.

p11-kit/generate-keypair.c
ZoltanFridrich: Maybe, I can check.

p11-kit/generate-keypair.c
ZoltanFridrich: I have added these to shorten the code inside `generate_keypair` and make it easier to read.

p11-kit/generate-keypair.c
ZoltanFridrich: ok

p11-kit/generate-keypair.c
ueno: In that case I would use `p11_attrs` API, which internally keeps a copy. That would also make `get_templates` simpler at a cost of some heap allocations (it's not a problem here as this is not a performance critical code).

ZoltanFridrich: I have added a commit with second `get_templates` function where I used the `attrs.h` API. I dislike it.

ueno: I don't understand why you dislike it; the amount of code is smaller and you don't need to maintain attribute index by yourself (i and j).

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,467 @@
...
+}
+
+int
+p11_kit_generate_keypair (int argc,
+			  char *argv[])
+{
+	int opt, ret = 2;
+	char *label = NULL;
+	CK_ULONG bits = 0;
+	p11_buffer ec_params;
```
ueno, line 368: `p11_buffer` is used here just because it's a convenient structure with data and size? It's confusing as it gives the reader an impression that this data will be modified afterwards. Let's just use `const uint8_t *` with size.

p11-kit/generate-keypair.c
```diff
@@ -0,0 +1,467 @@
...
+	CK_ATTRIBUTE attr_private_false = { CKA_PRIVATE, &fval, sizeof (fval) };
+	CK_ATTRIBUTE attr_sign = { CKA_SIGN, &tval, sizeof (tval) };
+	CK_ATTRIBUTE attr_verify = { CKA_VERIFY, &tval, sizeof (tval) };
+	CK_ATTRIBUTE attr_decrypt = { CKA_DECRYPT, &tval, sizeof (tval) };
+	CK_ATTRIBUTE attr_encrypt = { CKA_ENCRYPT, &tval, sizeof (tval) };
+	CK_ATTRIBUTE attr_modulus = { CKA_MODULUS_BITS, (void *)bits, sizeof (*bits) };
+	CK_ATTRIBUTE attr_ec_params = { CKA_EC_PARAMS, (void *)bits, sizeof (*bits) };
+
+	tmp = p11_attrs_build (pub, &attr_token, &attr_private_false, &attr_sign, &attr_verify, NULL);
+	if (tmp == NULL) {
```
ueno, line 183: This is the place you could use assertion macros like `return_val_if_fail`, without doing any cleanup, as allocation failure is fatal and the program cannot proceed.

p11-kit/test-objects.sh
```diff
@@ -212,3 +212,19 @@ else
 	sed 's/^/# /' list.diff
 	exit 1
 fi
+
+cat > list.exp <<EOF
+EOF
+
+"$abs_top_builddir"/p11-kit/p11-kit-testable generate-keypair -q --type=mock "pkcs11:" > list.out
+
+echo 7..7
```
ueno, line 221: Maybe you might want to check #553 and possibly rebase on it; then you wouldn't need to maintain the number of tests by hand anymore.

p11-kit/generate-keypair.c
ZoltanFridrich: I think I prefer doing the cleanup anyway.

p11-kit/generate-keypair.c
ZoltanFridrich: done

p11-kit/test-objects.sh
ZoltanFridrich: done

ueno (approved):
Looks good to me, please remove "[WIP]" from the title when merging.

PR #557, opened 2023-09-18
list-modules: Do not embed module-name and module-path in URL

ueno:
module-name and module-path are query attributes, which are to modify
the default settings (RFC 7512 section 2.4).  As p11-kit manages the
default configuration, it shouldn't print them as part of the token
URL.  This instead prints the absolute path of each module in a
separate field of the output.


ZoltanFridrich (approved):
The change looks good.

PR #558, opened 2023-09-18
Make add-profile command to create token objects instead of session objects

### LINKED ISSUE #559: `p11-kit add-profile` checks existing profile on all tokens provided by a module
As profile objects are per token, we should only check the existence on a specific token, while the current implementation (`profile_exists`) [checks](https://github.com/p11-glue/p11-kit/blob/79fe36c8b2d8880a4fbde4cad0c3781d0f3917c4/p11-kit/add-profile.c#L62) all tokens provided by a module.

ZoltanFridrich:
- rework add-profiles
- make add-profile command to create token objects instead of session objects
- add some more error information

Closes #559

### DIFF
--- a/p11-kit/add-profile.c
+++ b/p11-kit/add-profile.c
@@ -109,9 +109,11 @@ add_profile (const char *token_str,
 	CK_FUNCTION_LIST **modules = NULL;
 	P11KitUri *uri = NULL;
 	P11KitIter *iter = NULL;
+	CK_BBOOL token = CK_TRUE;
 	CK_OBJECT_CLASS klass = CKO_PROFILE;
 	CK_ATTRIBUTE template[] = {
 	    { CKA_CLASS, &klass, sizeof (klass) },
+	    { CKA_TOKEN, &token, sizeof (token) },
 	    { CKA_PROFILE_ID, &profile, sizeof (profile) }
 	};
 	CK_ULONG template_len = sizeof (template) / sizeof (template[0]);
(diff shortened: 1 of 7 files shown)


p11-kit/add-profile.c
```diff
@@ -150,7 +152,7 @@ add_profile (const char *token_str,
 		session = p11_kit_iter_get_session (iter);
 		rv = module->C_CreateObject (session, template, template_len, &object);
 		if (rv != CKR_OK) {
-			p11_message (_("failed to create the profile object: %s"), p11_kit_strerror (rv));
+			p11_message (_("failed to create profile object: %s"), p11_kit_strerror (rv));
```
ueno, line 155: Wouldn't it be more sensible to return an error if there are no matching writable tokens?

p11-kit/add-profile.c
```diff
@@ -109,16 +109,18 @@ add_profile (const char *token_str,
 	CK_FUNCTION_LIST **modules = NULL;
 	P11KitUri *uri = NULL;
 	P11KitIter *iter = NULL;
+	CK_BBOOL token = CK_TRUE;
 	CK_OBJECT_CLASS klass = CKO_PROFILE;
 	CK_ATTRIBUTE template[] = {
 	    { CKA_CLASS, &klass, sizeof (klass) },
+	    { CKA_TOKEN, &token, sizeof (token) },
```
ueno, line 116: Good but I guess we should add a test, presumably using SoftHSM?

p11-kit/add-profile.c
ZoltanFridrich: that reminds me that I should change it to iterate only the tokens and open the session manually

ueno (approved):

Issue #555, opened 2023-09-16
`p11-kit export-object` doesn't work for public keys without SPKI stored


ueno:
Steps to reproduce:
```console
$ cat softhsm2.conf
directories.tokendir = /tmp/softhsm/
$ mkdir /tmp/softhsm
$ export SOFTHSM2_CONF=$PWD/softhsm2.conf
$ softhsm2-util --init-token --free --label test
$ GNUTLS_PIN=12345 p11tool --login --generate-privkey=ecdsa --label foo 'pkcs11:model=SoftHSM%20v2'
...
$ p11tool --export 'pkcs11:model=SoftHSM%20v2;token=test;object=foo;type=public'
...
$ p11-kit export-object 'pkcs11:model=SoftHSM%20v2;token=test;object=foo;type=public'
p11-kit: ../common/pem.c:277: p11_pem_write: Assertion `len > 0' failed.
```

The current implementation only retrieves CKA_PUBLIC_KEY_INFO, but the spec [says](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/pkcs11-spec-v3.1.pdf#page=66) about the attribute: "DER-encoding of the SubjectPublicKeyInfo for this public key. (MAY be empty, DEFAULT derived from the underlying public key data)".
I guess we would need to fallback to the algorithm specific method.

PR #568, opened 2023-09-21
export-object: Recover public key from algorithm specific attributes

ueno:
Some tokens such as SoftHSMv2 omit CKA_PUBLIC_KEY_INFO when it can be derived from algorithm specific attributes, such as CKA_MODULUS and CKA_PUBLIC_EXPONENT for RSA.  This adds support for it if libtasn1 is available at compile time.

Fixes: #555


p11-kit/export-object.c
```diff
@@ -163,28 +419,37 @@ export_object (const char *token_str)
...
+	switch (klass) {
+	case CKO_CERTIFICATE:
+		if (!export_certificate (iter, &buf))
 			goto cleanup;
-		}
+		break;
+	case CKO_PUBLIC_KEY:
+		if (!export_pubkey (iter, &buf))
+			goto cleanup;
+	default:
```
ZoltanFridrich, line 440: I would say we should return an error if we match an object that is neither a certificate not public key

p11-kit/export-object.c
```diff
@@ -163,28 +419,37 @@ export_object (const char *token_str)
 	}
 
 	p11_kit_iter_begin (iter, modules);
-	while (p11_kit_iter_next (iter) == CKR_OK) {
-		rv = p11_kit_iter_get_attributes (iter, &attr, 1);
-		if (rv != CKR_OK) {
-			p11_message (_("failed to retrieve attribute of an object"));
+	if (p11_kit_iter_next (iter) != CKR_OK) {
+		p11_message (_("no matching object"));
```
ZoltanFridrich, line 423: I would prefer this error message to be consistent across other commands. Either use the one I have been using in other commands like in delete-object or change the other ones to this one.

p11-kit/export-object.c
```diff
@@ -59,85 +67,333 @@ int
...
+
+	ok = true;
+
+cleanup:
+	p11_attrs_free (attrs);
+	return ok;
+}
+
+static bool
+export_certificate (P11KitIter *iter, p11_buffer *buf)
```
ZoltanFridrich, line 342: each function parameter should be on a separate line judging by the code format of other commands.

p11-kit/export-object.c
```diff
@@ -59,85 +67,333 @@ int
...
-		p11_message (_("failed to retrieve attribute of an object"));
+	spk = p11_asn1_encode (asn, &n_spk);
+	if (!spk) {
+		p11_message (_("unable to encode %s"), "subjectPublicKey");
+		goto cleanup;
+	}
+	p11_asn1_free (asn);
+
+	asn = p11_asn1_create (defs, "PKIX1.SubjectPublicKeyInfo");
+	return_val_if_fail (asn, false);
```
ZoltanFridrich, line 149: Wouldn't this leak `spk`?

p11-kit/export-object.c
```diff
@@ -59,85 +67,333 @@ int
...
+	if (!ec_params || !ec_point) {
+		p11_message (_("failed to retrieve attributes"));
+		goto cleanup;
+	}
+
+	defs = p11_asn1_defs_load ();
+	return_val_if_fail (defs, false);
+
+	asn = p11_asn1_create (defs, "PKIX1.SubjectPublicKeyInfo");
+	return_val_if_fail (asn, false);
```
ZoltanFridrich, line 229: This would leak `defs`

p11-kit/export-object.c
```diff
@@ -59,85 +67,333 @@ int
...
+
+	result = asn1_write_value (asn, "algorithm.parameters",
+				   ec_params->pValue, ec_params->ulValueLen);
+	if (result != ASN1_SUCCESS) {
+		p11_message (_("unable to write value"));
 		goto cleanup;
 	}
 
+	/* Strip 2 octes of DER header for OCTET STRING */
+	return_val_if_fail (ec_point->ulValueLen > 2, false);
```
ZoltanFridrich, line 246: also this would leak if fails.

p11-kit/export-object.c
```diff
@@ -59,85 +67,333 @@ int
...
 	}
 
-	attr.pValue = malloc (attr.ulValueLen);
-	if (attr.pValue == NULL) {
-		p11_message (_("failed to allocate memory"));
+	defs = p11_asn1_defs_load ();
+	return_val_if_fail (defs, false);
+
+	asn = p11_asn1_create (defs, "PKIX1.RSAPublicKey");
+	return_val_if_fail (asn, false);
```
ZoltanFridrich, line 126: would leak `defs`

ZoltanFridrich (changes requested):
Found some small things, other then that looks good.

p11-kit/export-object.c
ueno: It leaks, but I would say it's ok because this is not a library code. We use this macro in such occasions (untestable error in commands) already.

ZoltanFridrich:
I have tested the command manually together with valgrind. Seems to work fine.

p11-kit/export-object.c
ZoltanFridrich: I would prefer to still `goto cleanup` in such cases.

ueno: Note that those macros are designed as a counterpart `g_return_if_*`, which are used like an assert: https://developer-old.gnome.org/glib/stable/glib-Warnings-and-Assertions.html

As they will cause the program abort if `P11_KIT_STRICT` is set (similar to `G_DEBUG=fatal_*` is set), we should already have a sufficient test coverage even if we dont't do the unrealistic cleanup. I believe other low-level commands (e.g., coreutils) do similar, e.g., they use `xmalloc` which simply aborts the program when memory allocation fails.

ZoltanFridrich: As you say, they are used like an assert, which guards against programming errors, not normal runtime errors like memory allocation failure and such. From `g_return_if_*` _"If expr evaluates to [FALSE](https://developer-old.gnome.org/glib/stable/glib-Standard-Macros.html#FALSE:CAPS), the current function should be considered to have undefined behaviour (a programmer error)"_. If `asn = p11_asn1_create (defs, "PKIX1.SubjectPublicKeyInfo");` fails, thats not a programming error. I think we should cleanup in such cases.

p11-kit/export-object.c
ZoltanFridrich: I have opened an MR to adjust the other error messages

p11-kit/export-object.c
ueno: I argue against that. The failure in `asn = p11_asn1_create (defs, "PKIX1.SubjectPublicKeyInfo");`  really shouldn't happen and we can't do anything on that. That is a kind of undefined behavior in this case. You can see the exact same pattern in the `trust extract` command: https://github.com/p11-glue/p11-kit/blob/e8872bfdc4b12735bbc0433a59fc0a7ddfd68bbb/trust/extract-openssl.c#L299

ueno: In the end I refactored the code by splitting the function into smaller ones, so we don't need `goto` for some cases, and proper error handling even in unrealistic cases. In the future maybe we should consider using automatic cleanup as suggested on #577.

ZoltanFridrich (approved):
Looks good!

p11-kit/export-object.c
```diff
@@ -59,73 +68,359 @@ int
...
+		*padded = 0x00;
+		free (attr->pValue);
+		attr->pValue = padded;
+		attr->ulValueLen++;
 	}
+}
 
-	if (p11_kit_iter_get_attributes (iter, &attr, 1) != CKR_OK) {
-		p11_message (_("failed to retrieve attribute length of an object"));
+unsigned char *
```
ZoltanFridrich, line 101: `static` keyword missing

PR #575, opened 2023-09-22
p11-kit tool: make "no matching object/token" error user friendly 

### LINKED ISSUE #574: `p11-kit` commands expecting a single token/object prints confusing error upon no match
If there are no matching objects, the `p11-kit delete-object` command shows the following error:
```console
p11-kit delete-object "pkcs11:token=nonexistent"
p11-kit: failed to find the object: The operation was cancelled
```
This is confusing, because there is no user interaction involved, but `CKR_CANCEL` is only used as a convention to tell that there is no match:
https://p11-glue.github.io/p11-glue/p11-kit/manual/p11-kit-Future.html#p11-kit-iter-next

ZoltanFridrich:
- don't check slot == 0 as 0 is a valid slot ID
-  make the error message human friendly when no token/object is matched

Closes #574

### DIFF
--- a/p11-kit/add-profile.c
+++ b/p11-kit/add-profile.c
@@ -117,12 +117,7 @@ add_profile (const char *token_str,
 		p11_message (_("failed to obtain module"));
 		goto cleanup;
 	}
-
 	slot = p11_kit_iter_get_slot (iter);
-	if (slot == 0) {
-		p11_message (_("failed to obtain slot"));
-		goto cleanup;
-	}
 
 	rv = module->C_OpenSession (slot, CKF_SERIAL_SESSION | CKF_RW_SESSION, NULL, NULL, &session);
 	if (rv != CKR_OK) {
--- a/p11-kit/add-profile.c
+++ b/p11-kit/add-profile.c
@@ -108,7 +108,10 @@ add_profile (const char *token_str,
 	p11_kit_iter_begin (iter, modules);
 	rv = p11_kit_iter_next (iter);
 	if (rv != CKR_OK) {
-		p11_message (_("failed to find the token: %s"), p11_kit_strerror (rv));
+		if (rv == CKR_CANCEL)
+			p11_message (_("no matching token"));
+		else
+			p11_message (_("failed to find token: %s"), p11_kit_strerror (rv));
 		goto cleanup;
 	}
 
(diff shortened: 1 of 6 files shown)


p11-kit/add-profile.c
```diff
@@ -106,9 +106,8 @@ add_profile (const char *token_str,
 	}
 
 	p11_kit_iter_begin (iter, modules);
-	rv = p11_kit_iter_next (iter);
-	if (rv != CKR_OK) {
-		p11_message (_("failed to find the token: %s"), p11_kit_strerror (rv));
+	if (p11_kit_iter_next (iter) != CKR_OK) {
```
ueno, line 109: I guess we could special case `CKR_CANCEL` in this case, as there might be other errors?

ueno (changes requested):
And of course, please add a test for the issues this PR is fixing.

p11-kit/test-objects.sh
```diff
@@ -258,5 +258,58 @@ EOF
...
+		sed 's/^/# /' list.diff
+		assert_fail "output contains wrong result"
+	fi
+        : ${ECHO=echo}
+        : ${TAIL=tail}
+        ${ECHO} "$(${TAIL} -n1 err.out)" > err.out
+	if ! ${DIFF} err.exp err.out > err.diff; then
+		sed 's/^/# /' err.diff
+		assert_fail "stderr contains wrong result"
+	fi
```
ueno, line 284:
```suggestion
        assert_contains err.out "no matching token"
```

p11-kit/test-objects.sh
```diff
@@ -258,5 +258,58 @@ EOF
...
```
ueno, line 310: Same suggestion above, using `assert_contains`.

p11-kit/test-profiles.sh
```diff
@@ -33,4 +33,31 @@ EOF
...
+		sed 's/^/# /' list.diff
+		assert_fail "output contains wrong results"
+	fi
+        : ${ECHO=echo}
+        : ${TAIL=tail}
+        ${ECHO} "$(${TAIL} -n1 err.out)" > err.out
+	if ! ${DIFF} err.exp err.out > err.diff; then
+		sed 's/^/# /' err.diff
+		assert_fail "stderr contains wrong result"
+	fi
```
ueno, line 59: Same suggestion above, using `assert_contains`.

p11-kit/test-objects.sh
ZoltanFridrich: that looks nice, thanks!

Issue #563, opened 2023-09-20
`p11-kit` command: add means to list mechanisms a token supports


ueno:
It would be useful if a command is provided to list mechanisms supported by a token. p11tool (GnuTLS) provides this functionality with `--list-mechanisms` option.

PR #576, opened 2023-09-22
WIP: list-mechanisms: initial implementation

t184256:
Resolves: #563

To be addressed:
* is mixing APIs OK?
* which printing API should I use?
* there are no tests


p11-kit/list-mechanisms.c
```diff
@@ -0,0 +1,303 @@
...
+		goto cleanup;
+	}
+
+	iter = p11_kit_iter_new (uri, P11_KIT_ITER_WITH_LOGIN | P11_KIT_ITER_WITH_TOKENS | P11_KIT_ITER_WITHOUT_OBJECTS);
+	if (iter == NULL) {
+		p11_message (_("failed to initialize iterator"));
+		goto cleanup;
+	}
+
+	p11_list_printer_init (&printer, stdout, 0);
```
ueno, line 187: There are some duplicated lines between line 139 to 187 (`p11_kit_modules_load_and_initialize (0)` is called twice, etc.). Maybe a copy and paste error?

p11-kit/list-mechanisms.c
```diff
@@ -0,0 +1,303 @@
...
+	int ret = 1;
+	CK_FUNCTION_LIST **modules = NULL;
+	CK_FUNCTION_LIST *module = NULL;
+	P11KitUri *uri = NULL;
+	P11KitIter *iter = NULL;
+	CK_SESSION_HANDLE session = 0;
+	CK_SLOT_ID slot = 0;
+	CK_MECHANISM_INFO mechanism_info;
+	CK_MECHANISM_TYPE_PTR mechanisms = NULL;
+	CK_MECHANISM_TYPE mechanism = CKP_INVALID_ID;
```
ueno, line 133: Type mismatch: `CKP_INVALID_ID` is defined as CK_PROFILE_ID

p11-kit/list-mechanisms.c
```diff
@@ -0,0 +1,303 @@
...
+		}
+		p11_list_printer_start_section (&printer, "Token", "#%lu", slot);
+
+		rv = module->C_GetMechanismList (slot, NULL, &mechanisms_count);
+		if (rv != CKR_OK) {
+			p11_message (_("querying amount of mechanisms failed: %s"), p11_kit_strerror (rv));
+			goto cleanup;
+		}
+
+		mechanisms = malloc(sizeof(CK_MECHANISM_TYPE) * mechanisms_count);
```
ueno, line 209:
```suggestion
		mechanisms = reallocarray (NULL, mechanisms_count, sizeof (CK_MECHANISM_TYPE));
```
Maybe it would also be possible to reuse the previously allocated `mechanisms` by resizing.

p11-kit/list-mechanisms.c
t184256: side-stepped

p11-kit/list-mechanisms.c
t184256: indeed, removed

p11-kit/list-mechanisms.c
t184256: right, and it was also leaking. used reallocarray

t184256:
addressed feedback, added a test against mock data, added printing token labels

I don't see how reasonably use the printing API for what I'm doing, so I just hope mixing it with printfs is OK.

p11-kit/list-mechanisms.c
```diff
@@ -0,0 +1,288 @@
...
+	p11_kit_iter_begin (iter, modules);
+	while ((rv = p11_kit_iter_next (iter)) == CKR_OK) {
+		module = p11_kit_iter_get_module (iter);
+		if (module == NULL) {
+			p11_message (_("failed to obtain module"));
+			goto cleanup;
+		}
+
+		slot = p11_kit_iter_get_slot (iter);
+		if (slot == 0) {
```
ueno, line 174: 0 is a valid slot ID; we can safely assume that slot is always valid.

p11-kit/test-list-mechanisms.sh
```diff
@@ -0,0 +1,47 @@
...
+0x80000002 (unknown): sign verify key-size=2048-2048
+token: TEST LABEL
+0x80000001 (unknown): encrypt decrypt key-size=512-4096
+0x80000002 (unknown): sign verify key-size=2048-2048
+token: TEST LABEL
+0x80000001 (unknown): encrypt decrypt key-size=512-4096
+0x80000002 (unknown): sign verify key-size=2048-2048
+EOF
+
+	if ! "$abs_top_builddir"/p11-kit/p11-kit-testable list-mechanisms -q pkcs11: > list.out; then
```
ueno, line 36: This URL matches all tokens? I would rather limit it to the first match, as `p11tool --list-mechanism` does.

p11-kit/list-mechanisms.c
```diff
@@ -0,0 +1,288 @@
...
+
+	modules = p11_kit_modules_load_and_initialize (0);
+	if (modules == NULL) {
+		p11_message (_("failed to load and initialize modules"));
+		goto cleanup;
+	}
+
+	iter = p11_kit_iter_new (uri, P11_KIT_ITER_WITH_LOGIN | P11_KIT_ITER_WITH_TOKENS | P11_KIT_ITER_WITHOUT_OBJECTS);
+	if (iter == NULL) {
+		p11_message (_("failed to initialize iterator"));
```
ueno, line 160: As this is an internal error, it might make sense to use `p11_debug` instead.

p11-kit/list-mechanisms.c
t184256: removed the check

p11-kit/list-mechanisms.c
t184256: changed to `p11_debug`

p11-kit/test-list-mechanisms.sh
t184256: > This URL matches all tokens?

Yes.

> I would rather limit it to the first match, as p11tool --list-mechanism does.

You mean, limit `list-mechanisms` subcommand to inspecting the first matching token? I'd disagree on that. First, I think it might introduce confusion in a scenario where the user accidentally matches several tokens. Second, AFAICS this behaviour won't be in line with the existing `list-*` subcommands that iterate over all matches.

p11-kit/list-mechanisms.c
```diff
@@ -0,0 +1,285 @@
...
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * Author: Alexander Sosedkin <asosedkin@redhat.com>
+ */
+
+#include "config.h"
+
+#include "constants.h"
+#define P11_DEBUG_FLAG P11_DEBUG_LIB
```
ueno, line 40:
```suggestion
#define P11_DEBUG_FLAG P11_DEBUG_TOOL
```

p11-kit/list-mechanisms.c
```diff
@@ -0,0 +1,285 @@
...
+	p11_list_printer printer;
+	CK_RV rv;
+
+	uri = p11_kit_uri_new ();
+	if (uri == NULL) {
+		p11_message (_("failed to allocate memory"));
+		goto cleanup;
+	}
+
+	if (p11_kit_uri_parse (token_str, P11_KIT_URI_FOR_OBJECT_ON_TOKEN, uri) != P11_KIT_URI_OK) {
```
ueno, line 148: We iterate over tokens, not objects on them, right?
```suggestion
	if (p11_kit_uri_parse (token_str, P11_KIT_URI_FOR_TOKEN, uri) != P11_KIT_URI_OK) {
```

p11-kit/test-list-mechanisms.sh
ueno: > AFAICS this behaviour won't be in line with the existing list-* subcommands that iterate over all matches

Not really:
- `p11-kit list-modules`: doesn't take URL
- `p11-kit list-profiles`: takes a token URL, uses first match only, because profiles are capabilities of a particular token
- `p11-kit list-objects`: takes an *object* URL, returns all matches

t184256: Changed to stop after first match. Testing 3 scenarios now: "pkcs11:", specific existing and specific non-existing token. Maybe when we have list-tokens we should also test the iteration you're suggesting.

p11-kit/list-mechanisms.c
```diff
@@ -0,0 +1,281 @@
...
+
+	iter = p11_kit_iter_new (uri, P11_KIT_ITER_WITH_LOGIN | P11_KIT_ITER_WITH_TOKENS | P11_KIT_ITER_WITHOUT_OBJECTS);
+	if (iter == NULL) {
+		p11_debug ("failed to initialize iterator");
+		goto cleanup;
+	}
+
+	p11_list_printer_init (&printer, stdout, 0);
+	p11_kit_iter_begin (iter, modules);
+	rv = p11_kit_iter_next (iter);  // we will stop at the first token
```
ueno, line 166:
```suggestion
	rv = p11_kit_iter_next (iter);
	if (rv != CKR_OK) {
		if (rv == CKR_CANCEL)
			p11_message (_("no matching token"));
		else
			p11_message (_("failed to find token: %s"), p11_kit_strerror (rv));
		goto cleanup;
	}
```

t184256: incorporated

ueno (approved):
Thank you, looks good to me!

PR #580, opened 2023-09-25
Add tests for p11-kit profile commands

### LINKED ISSUE #548: Add tests for p11-kit add-profile, delete-profile, and delete-object commands
We didn't add tests for those commands, as it's not straightforward to preserve object states across program invocations. That said we could store the states into a file, possibly in a .p11-kit format using [p11_persist_write](https://github.com/p11-glue/p11-kit/blob/11394d91ec05ff63f5f7201bb9c2098d707fe4e4/trust/persist.h#L64).

ZoltanFridrich:
- added module that is able to load objects from and store result in a .p11-kit file
- moved persist.c into common
- added actually useful tests for profile commands

Closes #548


p11-kit/Makefile.am
```diff
@@ -607,6 +612,10 @@ mock_twelve_la_SOURCES = p11-kit/mock-module-ep10.c
 mock_twelve_la_LDFLAGS = $(mock_one_la_LDFLAGS)
 mock_twelve_la_LIBADD = $(mock_one_la_LIBADD)
 
+mock_thirteen_la_SOURCES = p11-kit/mock-module-ep11.c
+mock_thirteen_la_LDFLAGS = $(mock_one_la_LDFLAGS) $(LIBTASN1_CFLAGS)
```
ueno, line 616: Don't add CFLAGS to LDFLAGS:
```suggestion
mock_thirteen_la_LDFLAGS = $(mock_one_la_LDFLAGS)
mock_thirteen_la_CFLAGS = $(LIBTASN1_CFLAGS)
```

ueno (commented):
Maybe good to test without libtasn1-devel installed; I see a couple of test failures.

ueno (approved):
Looks good to me!

PR #582, opened 2023-10-01
p11-kit generate-keypair: Support EdDSA key generation

ueno:
Also add tests for the p11-kit generate-keypair command using SoftHSM2.


ZoltanFridrich (approved):
Looks nice.

PR #586, opened 2023-10-05
Fix CI warnings and errors

ZoltanFridrich:
The numeric values that were passed into `p11_hash_murmur3`, a variable argument function, were smaller then `size_t` which resulted in a memory access error when they were read inside the function as variable argument of length `size_t` .
Other then that, this MR fixes some deprecated meson usage.


ueno (approved):
Looks good to me, though the variable initializations look mostly false-positive (for cppcheck).

Issue #570, opened 2023-09-21
p11-kit list-objects should support interactive prompts for PIN


Jakuje:
Writing PIN inside of the URI might not be the best security practice (while it is handy and common). The `pin-source` should already provide a way to get pin from file or from some command so it should be easy to spawn some prompt, but having a separate option (`--login`?) to log in and ask interactively for pin would be preferred.

Additionally, I was not able to make the `list-objects` login to the token with the uri like this -- prints just the publicly accessible private key:
```
p11-kit list-objects --verbose pkcs11:type=private?pin-source=file:~/pin
```
(the `--verbose` really does not print anything new and I really do not see if the log was attempted, if if failed or what happened)

Issue #564, opened 2023-09-20
`p11-kit list-objects` doesn't return any matches when a query attribute is in URL


ueno:
After generating a key, when listing objects with a query attribute, it doesn't show any matches, while it does if the query attribute is omitted:

```console
$ p11-kit generate-keypair --type=rsa --bits=2048 --label=foo2 'pkcs11:model=SoftHSM%20v2?pin-value=12345'

# OK
$ p11-kit list-objects 'pkcs11:model=SoftHSM%20v2'                
Object: #0
    uri: pkcs11:model=SoftHSM%20v2;manufacturer=SoftHSM%20project;serial=4d5c2be2819673ce;token=test;id=;object=foo2;type=public
    class: public-key
    key-type: rsa
    label: foo2

# NG
$ p11-kit list-objects 'pkcs11:model=SoftHSM%20v2?pin-value=12345'
```

Ideally, it should also perform login if pin-value is specified, as the objects with CKA_PRIVATE flag enabled.

ZoltanFridrich:
This has been fixed and the test exists `test_list_private`

PR #587, opened 2023-10-06
p11-kit commands: Add --login option

ueno:
Previously, those tools determined whether a login is necessary by
checking the presence of "pin-value" query attribute in the URI.  It
was too implicit and against modern security practice.  This instead
asks users to specify --login option and if no "pin-value" is given,
it tries to read a PIN from the terminal.

Fixes: #570

### DIFF
--- a/common/compat.h
+++ b/common/compat.h
@@ -381,6 +381,27 @@ int        isatty           (int fd);
 
 #endif /* HAVE_ISATTY */
 
+#ifdef HAVE_READPASSPHRASE
+
+#include <readpassphrase.h>
+
+#else
+
+#define RPP_ECHO_OFF    0x00		/* Turn off echo (default). */
+#define RPP_ECHO_ON     0x01		/* Leave echo on. */
+#define RPP_REQUIRE_TTY 0x02		/* Fail if there is no tty. */
+#define RPP_FORCELOWER  0x04		/* Force input to lower case. */
+#define RPP_FORCEUPPER  0x08		/* Force input to upper case. */
+#define RPP_SEVENBIT    0x10		/* Strip the high bit from input. */
+#define RPP_STDIN       0x20		/* Read from stdin, not /dev/tty */
+
+char      *readpassphrase   (const char *prompt,
+                             char *buf,
+                             size_t bufsiz,
+                             int flags);
+
+#endif /* HAVE_READPASSPHRASE */
+
 /* If either locale_t or newlocale() is not available, strerror_l()
  * cannot be used */
 #if !defined(HAVE_LOCALE_T) || !defined(HAVE_NEWLOCALE)
--- a/p11-kit/add-profile.c
+++ b/p11-kit/add-profile.c
@@ -43,6 +43,10 @@
 #include "message.h"
 #include "tool.h"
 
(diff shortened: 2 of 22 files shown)


p11-kit/add-profile.c
```diff
@@ -115,27 +117,11 @@ add_profile (const char *token_str,
 		goto cleanup;
 	}
 
+	/* Module and session should always set at this point.  */
```
ZoltanFridrich, line 130: change `always set` to `always be set`

p11-kit/delete-profile.c
```diff
@@ -115,27 +117,11 @@ delete_profile (const char *token_str,
 		goto cleanup;
 	}
 
+	/* Module and session should always set at this point.  */
```
ZoltanFridrich, line 130: missing `be`. Its missing in the rest of the comments as well

p11-kit/delete-profile.c
```diff
@@ -115,27 +117,11 @@ delete_profile (const char *token_str,
...
-
-	pin = p11_kit_uri_get_pin_value (uri);
-	if (pin != NULL) {
-		rv = module->C_Login (session, CKU_USER, (unsigned char *)pin, strlen (pin));
-		if (rv != CKR_OK) {
-			p11_message (_("failed to login: %s"), p11_kit_strerror (rv));
-			goto cleanup;
-		}
-	}
+	return_val_if_fail (module, 1);
```
ZoltanFridrich, line 132: nit: I would prefer to have all of these specify `!= NULL` so its consistent with the rest of the code. But it doesnt really matter

common/compat.h
```diff
@@ -381,6 +381,29 @@ int        isatty           (int fd);
 
 #endif /* HAVE_ISATTY */
 
+#ifdef HAVE_READPASSPHRASE
+
+#ifdef HAVE_READPASSPHRASE_H
```
ZoltanFridrich, line 386: just a question: so there might be a situation where `HAVE_READPASSPHRASE` is defined but no header `HAVE_READPASSPHRASE_H`? Should we error in such case?

p11-kit/add-profile.c
```diff
@@ -171,25 +181,29 @@ p11_kit_add_profile (int argc,
...
 	CK_ULONG profile = CKA_INVALID;
 	p11_dict *profile_nicks = NULL;
+	bool login = false;
 
 	enum {
 		opt_verbose = 'v',
 		opt_quiet = 'q',
 		opt_help = 'h',
 		opt_profile = 'p',
+		opt_login = CHAR_MAX + 1,
```
ZoltanFridrich, line 191: dont we want an abbreviation of `--login` `-l`?

ZoltanFridrich (changes requested):
Overall looks nice. I didn't see any problems. Just a few nits and questions.

ZoltanFridrich:
does this also resolve #564?

ZoltanFridrich (approved):
LGTM

common/compat.h
ueno: Good point; I just removed the header check so it should fail at compile time.

PR #588, opened 2023-10-16
Add command for importing objects into a PKCS#11 token

### LINKED ISSUE #394: Add basic key and object manipulation subcommands to p11-kit

ZoltanFridrich:
The import-object command allows users to import a certificates or public keys into a pkcs#11 token by providing a PEM file.

`$ p11-kit import-object --login --label=MyCert --file=MyCert.pem 'pkcs11:token=MyToken'`

Closes #394

### DIFF
--- a/p11-kit/import-object.c
+++ b/p11-kit/import-object.c
@@ -1,0 +1,615 @@
+/*
+ * Copyright (c) 2023, Red Hat Inc.
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the
+ *       following disclaimer.
+ *     * Redistributions in binary form must reproduce the
+ *       above copyright notice, this list of conditions and
+ *       the following disclaimer in the documentation and/or
+ *       other materials provided with the distribution.
+ *     * The names of contributors to this software may not be
+ *       used to endorse or promote products derived from this
+ *       software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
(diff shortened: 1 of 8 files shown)


common/compat.c
```diff
@@ -308,6 +308,8 @@ p11_mmap_open (const char *path,
 void
 p11_mmap_close (p11_mmap *map)
 {
+	if (map == NULL)
```
ueno, line 311: I wouldn't introduce free-like semantics to `p11_mmap`; it should be an programming error if called on NULL.

p11-kit/import-object.c
```diff
@@ -0,0 +1,606 @@
...
+
+#ifdef WITH_ASN1
+
+typedef struct {
+	CK_FUNCTION_LIST *module;
+	CK_SESSION_HANDLE session;
+	const char *label;
+} import_data;
+
+static void
```
ueno, line 81: Should this return error code?

p11-kit/import-object.c
```diff
@@ -0,0 +1,606 @@
...
+import_x509_cert(const unsigned char *der,
```
ueno, line 82: Put a space before `(`.

p11-kit/import-object.c
```diff
@@ -0,0 +1,606 @@
...
+
+cleanup:
+	free (attr_ec_params.pValue);
+	free (attr_ec_point.pValue);
+	free (ec_point);
+
+	return result;
+}
+
+static void
```
ueno, line 320: Should this return error code?

p11-kit/import-object.c
```diff
@@ -0,0 +1,606 @@
...
+	CK_ATTRIBUTE attr_private = { CKA_PRIVATE, &fval, sizeof (fval) };
+	CK_ATTRIBUTE attr_cert_type = { CKA_CERTIFICATE_TYPE, &cert_type, sizeof (cert_type) };
+	CK_ATTRIBUTE attr_value = { CKA_VALUE, (void *)der, der_len };
+	CK_ATTRIBUTE attr_subject = { CKA_SUBJECT, };
+
+	return_if_fail (data != NULL);
+	return_if_fail (data->module != NULL);
+	return_if_fail (data->session != 0);
+
+	defs = p11_asn1_defs_load ();
```
ueno, line 106: Can this be moved to `import_pem` to avoid code duplication?

ZoltanFridrich: yes

p11-kit/import-object.c
ZoltanFridrich: I guess I could propagate the error code via the `ImportData` structure or a global variable and return 1 in the main function in such case. Otherwise there is no point returning anything as the parsing function for `p11_pem_parse` doesnt return anything.

p11-kit/import-object.c
```diff
@@ -0,0 +1,623 @@
...
+
+typedef struct {
+	bool parse_error;
+	CK_FUNCTION_LIST *module;
+	CK_SESSION_HANDLE session;
+	const char *label;
+} import_data;
+
+static bool
+import_x509_cert (p11_dict *defs,
```
ueno, line 83: What about making `defs` part of `import_data`?

ueno (approved):
LGTM, except a minor nit.

Issue #589, opened 2023-10-17
Mark P11KitIter interface stable


ueno:
Currently the P11KitIter interface is only exposed when `P11_KIT_FUTURE_UNSTABLE_API` is defined. Since the interface has been there for many years, it might make sense to declare it to be stable.

A couple of things to consider before doing so:
- `P11_KIT_ITER_WITHOUT_OBJECTS` was added as a compatibility measure in 3d54011b0d0bf1b31fbab8d7025b7201722d61c3, as P11KitIter previously only supported enumerating objects. Maybe we could make it no-op
- There is a recent behavior [change](fd7c819eff1de40619ba0231dc8819af76df72a6) (not part of any release yet) regarding the timing to stop iteration when `P11_KIT_ITER_WITH_TOKENS` is specified; previously we stopped before opening a session, now we stop after that. If we consider this is an ABI break, we might want to bump SONAME

ZoltanFridrich:
|* P11_KIT_ITER_WITHOUT_OBJECTS was added as a compatibility measure in https://github.com/p11-glue/p11-kit/commit/3d54011b0d0bf1b31fbab8d7025b7201722d61c3, as P11KitIter previously only supported enumerating objects. Maybe we could make it no-op
Why would we make it no-op? I thought that when I specify something like `WITH_TOKENS` it will enumerate both objects and tokens. What if I dont want to enumerate the objects but only the tokens? Unless I imagined this behavior and actually it works differently.

| There is a recent behavior [change](https://github.com/p11-glue/p11-kit/issues/fd7c819eff1de40619ba0231dc8819af76df72a6) (not part of any release yet) regarding the timing to stop iteration when P11_KIT_ITER_WITH_TOKENS is specified; previously we stopped before opening a session, now we stop after that. If we consider this is an ABI break, we might want to bump SONAME
Is it an ABI break though? The session is opened automatically now but that shouldnt break anything. I believe that anyone who uses such iterator and manually opens the session, their program should still work.

ueno:
> Why would we make it no-op? I thought that when I specify something like WITH_TOKENS it will enumerate both objects and tokens. What if I dont want to enumerate the objects but only the tokens? Unless I imagined this behavior and actually it works differently.

If we don't care about backward compatibility, we could have `WITH_OBJECTS` instead of `WITHOUT_OBJECTS`, which would be more intuitive.

[...]

PR #590, opened 2023-10-19
iter: Add option to stop at token with active session

ueno:
The commit fd7c819eff1de40619ba0231dc8819af76df72a6 introduced a slight backward incompatibility to the P11KitIter behavior when iterating with P11_KIT_ITER_WITH_TOKENS: previously it stopped before opening a token, while now it stops after that.

This adds a compatibility measure so those two are distinguishable with an explicit flag P11_KIT_ITER_WITH_SESSIONS and a kind P11_KIT_ITER_KIND_SESSION.

Related: #589

### DIFF
--- a/p11-kit/iter.c
+++ b/p11-kit/iter.c
@@ -116,6 +116,7 @@ struct p11_kit_iter {
 	unsigned int with_tokens : 1;
 	unsigned int with_objects : 1;
 	unsigned int with_login : 1;
+	unsigned int with_sessions : 1;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


p11-kit/iter.c
```diff
@@ -636,6 +642,11 @@ move_next_session (P11KitIter *iter)
 		if (rv != CKR_OK || !p11_match_uri_token_info (&iter->match_token, &iter->token_info))
```
ZoltanFridrich, line 642: the indentation seems to be a bit messed up in this file

ZoltanFridrich (approved):
Looks ok

p11-kit/iter.c
ueno: I've wrapped long lines (if you mean that).

PR #591, opened 2023-10-25
revise documentation of object and profile commands

ZoltanFridrich:
This is to adjust the p11-kit documentation after all of the behavioral changes done to the object and profile sub-commands.


ueno:
Btw s/rewise/revise/ ?

PR #594, opened 2023-10-26
pkcs11-json: Make regeneration robuster

ueno:
Previously, when common/pkcs11.h is updated and castxml is not installed on the system, an empty pkcs11.json is created.  This makes the build process abort in that case.

This also fixes the path of pkcs11.json referred to by p11-kit/Makefile.am, and only includes necessary files in the distribution.


ZoltanFridrich (approved):
LGTM

Issue #593, opened 2023-10-25
Make separation of unit tests / integration tests clearer


ueno:
Currently some unit tests rely on exotic packages like softhsm and are skipped if they are not available. That makes it [hard](https://bugzilla.redhat.com/show_bug.cgi?id=2246048) to isolate issues from the test execution environment. I suggest:

- making the unit tests self-contained to only use mock modules
- exercising the integration tests after installation (i.e., in Meson we could install them in `installed-tests`, in Autotools we could also run them at `installcheck`).

PR #596, opened 2023-10-27
tests: Run tests using SoftHSM after installation

ueno:
This makes the unit tests self-contained so all possible code paths are exercised without optional packages installed.  The integration tests against SoftHSM can still run after installation, through "make installcheck" or running the scripts manually.

Fixes: #593


ZoltanFridrich (approved):
Works and looks good!

ueno:
Thanks for the review!

Issue #598, opened 2023-11-04
Meson build failure in macOS Catalina (appleframeworks not found, tried framework) in p11-kit 0.25.1+


aminehaddad:
I've been having compilation failures since [p11-kit 0.25.1/0.25.2](https://github.com/p11-glue/p11-kit/commit/daabb9f2f6e2443e4ce3155ac75341c924175b50?diff=split) in macOS Catalina (with xcode-select installed). It was working fine with 0.25.0. Should this be reported in Meson or perhaps I am missing something on my end?

```bash
==> meson setup build -Dsystem_config=/usr/local/etc -Dmodule_config=/usr/local/etc/pkcs11/modules -Dtrust_paths=/usr/local/etc/ca-certificates/cert.pem" -Dsystemd=disabled
Last 15 lines from /Users/username/Library/Logs/Homebrew/p11-kit/01.meson:
Build type: native build
Project name: p11-kit
Project version: 0.25.2
C compiler for the host machine: clang (clang 12.0.0 "Apple clang version 12.0.0 (clang-1200.0.32.29)")
C linker for the host machine: clang ld64 609.8
Host machine cpu family: x86_64
Host machine cpu: x86_64
Has header "libintl.h" : YES 
Library intl found: YES
Checking for function "dgettext" with dependency -lintl: YES 
Run-time dependency appleframeworks found: NO (tried framework)

meson.build:70:26: ERROR: Dependency "appleframeworks" not found, tried framework
```

Side note: it is used for dependencies in QEMU -> gnutls -> p11-kit. Thanks!

ueno:
Given the CI [passes](https://github.com/p11-glue/p11-kit/actions/runs/6704529806/job/18217113330) on macOS 12 with "Run-time dependency appleframeworks found: YES (foundation)", I suspect there might be something missing on your build setup, though I can't point what's exactly causing this. @neverpanic do you have any thought?

aminehaddad:
Quick test with check_appleframeworks.c and meson.build:

```c
#include <stdio.h>

int main() {
    printf("Apple Frameworks Test\n");
    return 0;
}
```

```meson
project('check_appleframeworks', 'c')
src = 'check_appleframeworks.c'
appleframeworks = dependency('appleframeworks', modules: 'foundation')

if appleframeworks.found()
  message('appleframeworks with "foundation" module found')
else
  message('appleframeworks with "foundation" module not found')
endif

[...]

PR #600, opened 2023-11-13
"Foudation" Apple Framework should be case sensitive

claviola:
As per issue #598, the Foundation AppleFramework should be searched case-sensitively, to accomodate for case-sensitive APFS partitions.


meson.build
```diff
@@ -67,7 +67,7 @@ if get_option('nls') and cc.has_header('libintl.h')
     if cc.has_function('dgettext', dependencies : libintl)
       libintl_deps += libintl
       if ['darwin', 'ios'].contains(host_system)
-        appleframeworks = dependency('appleframeworks', modules : 'foundation')
+        appleframeworks = dependency('appleframeworks', modules : 'Foundation')
```
ueno, line 70: I suspect we might only need CoreFoundation, as the libintl only checks CF* functions:
https://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=blob;f=m4/intlmacosx.m4;h=3141bf8c5e4c949f03d4757701ca9c5fc3c6c9a1;hb=HEAD

ZoltanFridrich:
Fixed with #603

PR #601, opened 2023-11-14
rpc: Fix serialization of NULL mechanism pointer

ueno:
A NULL mechanism pointer is valid for C_*Init functions to cancel the
operation.  Since 852ccd8d we encoded it with a CK_MECHANISM_TYPE 0 as
an indicator, though it clashes with CKM_RSA_PKCS_KEY_PAIR_GEN (0).
This patch changes the encoding to use a special value (0xffffffff) to
indicate that and also properly advance the offset when reading.


ZoltanFridrich (approved):
I can confirm that this patch fixes the remote-option test.

PR #604, opened 2023-11-15
Add the flags required for libtasn1 to libp11-asn1.la

MaxHearnden:
These were not copied when libp11-asn1.la was seperated from libtrust-data.la as part of d8cc241187775e18de86caf446e9160299e6f157


ueno (approved):
Good catch, thanks!

PR #605, opened 2023-11-16
Fix issues found by static analysis

ZoltanFridrich:
This fixes all important errors that are not false positives and new errors reported by covscan


ueno (approved):
LGTM

PR #614, opened 2023-12-17
Fixes for GNU/Hurd

ametzler:
Hello,

we have been patching p11-kit on Debian for GNU/Hurd and kfreebsd compatibility for a long time. Find attached the latest iteration of the (small) patchset. kfreebsd-compatibilty is untested nowadays since we have dropped the port on Debian. - Please tell me if I should drop the respective code ("kfreebsd*-gnu |")

cu Andreas

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -206,7 +206,8 @@ if host_system != 'windows'
     'mkstemp',
     'readpassphrase',
     'secure_getenv',
-    'strndup'
+    'strndup',
+    'strnstr'
   ]
   if ['linux'].contains(host_system)
     functions += [
--- a/meson.build
+++ b/meson.build
@@ -59,6 +59,14 @@ endif
 
 conf.set10('WITH_STRICT', get_option('strict'))
 
+system_deps = []
+
+if ['gnu', 'gnu/kfreebsd'].contains(host_system)
+  libbsd_overlay_dep = dependency('libbsd-overlay')
+  system_deps += libbsd_overlay_dep
+  add_project_dependencies(libbsd_overlay_dep, language: 'c')
+endif
+
 libintl_deps = []
 if get_option('nls') and cc.has_header('libintl.h')
   conf.set('ENABLE_NLS', 1)
@@ -180,7 +188,7 @@ if host_system != 'windows'
   ]
 
   foreach h : headers
-    if cc.has_header(h)
+    if cc.has_header(h, dependencies: system_deps)
       conf.set('HAVE_' + h.underscorify().to_upper(), 1)
     endif
   endforeach
@@ -203,7 +211,7 @@ if host_system != 'windows'
   ]
 
   foreach f : functions
-    if cc.has_function(f)
+    if cc.has_function(f, dependencies: system_deps)
       conf.set('HAVE_' + f.underscorify().to_upper(), 1)
     endif
   endforeach
--- a/meson.build
+++ b/meson.build
@@ -195,7 +195,6 @@ if host_system != 'windows'
 
   functions = [
     'fdwalk',
-    'getauxval',
     'getexecname',
     'getpeereid',
     'getpeerucred',
@@ -209,6 +208,11 @@ if host_system != 'windows'
     'secure_getenv',
     'strndup'
   ]
+  if ['linux'].contains(host_system)
+    functions += [
+      'getauxval'
+    ]
+  endif
 
   foreach f : functions
     if cc.has_function(f, dependencies: system_deps)
--- a/common/compat.c
+++ b/common/compat.c
@@ -859,7 +859,7 @@ mkdtemp (char *template)
 #ifndef HAVE_GETAUXVAL
 
 unsigned long
-getauxval (unsigned long type)
+_p11_getauxval (unsigned long type)
 {
 	static unsigned long secure = 0UL;
 	static bool check_secure_initialized = false;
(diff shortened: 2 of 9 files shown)


ueno:
Thank you; I see no problem in merging this. One nit: the second commit has an ill-formatted log message "[use_libbsd-overlay on GNU/hurd and kfreebsd.](https://github.com/p11-glue/p11-kit/pull/614/commits/042a49fb03e93ecae2acfe3396a882a427147c33)"; maybe the first "_" should be replaced with a space?

Btw, is Meson available and supposed to work on GNU/Hurd?

ametzler:
ueno wrote
> Thank you; I see no problem in merging this. One nit: the second commit has an ill-formatted log message "[use_libbsd-overlay on GNU/hurd and kfreebsd.](https://github.com/p11-glue/p11-kit/pull/614/commits/042a49fb03e93ecae2acfe3396a882a427147c33)"; maybe the first "_" should be replaced with a space?

I have force-pushed this change.

> Btw, is Meson available and supposed to work on GNU/Hurd?

Yes, it should work. I played a little bit but this seems to need somebody familiar with meson. I did not get beyond
```
if ['gnu', 'gnu/kfreebsd'].contains(host_system)
  libbsd_overlay = dependency('libbsd-overlay')
endif
```
and adding this to dependencies: of some targets. Which is not enough. This test
```
  foreach h : headers
    if cc.has_header(h)
      conf.set('HAVE_' + h.underscorify().to_upper(), 1)
    endif
  endforeach
```
would also need to use libbsd-overlay's CFLAGS to find its ucred.h

ueno:
Afaik `has_header` also takes [`dependencies`](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerhas_header) so you can write something like:
```meson
  foreach h : headers
    if cc.has_header(h, dependencies: libbsd_overlay)
      conf.set('HAVE_' + h.underscorify().to_upper(), 1)
    endif
  endforeach
```

ametzler:
> Afaik `has_header` also takes [`dependencies`](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerhas_header) so you can write something like:
> 
> ```meson
>   foreach h : headers
>     if cc.has_header(h, dependencies: libbsd_overlay)
>       conf.set('HAVE_' + h.underscorify().to_upper(), 1)
>     endif
>   endforeach
> ```

Thanks, I have used this and pushed an updated version that also changes the meson build.

meson.build
```diff
@@ -180,14 +187,13 @@ if host_system != 'windows'
   ]
 
   foreach h : headers
-    if cc.has_header(h)
+    if cc.has_header(h, dependencies: libbsd_overlay_dep)
```
ueno, line 190: Do we still need this even with `add_project_dependencies(libbsd_overlay_dep, language: 'c')`? If yes, I would suggest having a separate array of dependencies:
```meson
system_deps = []

if ['gnu', 'gnu/kfreebsd'].contains(host_system)
  libbsd_overlay_dep = dependency('libbsd-overlay')
  system_deps += libbsd_overlay_dep
endif

if cc.has_header(h, dependencies: system_deps)
...
endif
```

ametzler: @ueno wrote:
> Do we still need this even with `add_project_dependencies(libbsd_overlay_dep, language: 'c')`?

Yes, both ```add_project_dependencies``` and ```add_project_arguments``` seem to be ignored by ```cc.has_header()``` et al. (Not setting '-D_GNU_SOURCE' from project_arguments when testing also makes a difference but did not cause obvious breakage)

>  If yes, I would suggest having a separate array of dependencies:
> 
> ```meson
> system_deps = []
> 
> if ['gnu', 'gnu/kfreebsd'].contains(host_system)
>   libbsd_overlay_dep = dependency('libbsd-overlay')
>   system_deps += libbsd_overlay_dep
> endif
> 
> if cc.has_header(h, dependencies: system_deps)
> ...
> endif
> ```

In my proposed change I had added libbsd-overlay only to the test-loops where it was necessary but not to e.g. ```cc.has_function('pthread_create``` or ```cc.has_function('dlopen')```. - Would you like me to basically do a s/libbsd_overlay_dep/system_deps/ or to extend the change to add system_deps to **all** instances of _cc.has..._?

cu Andreas

ueno: The point was that the current code uses an unusual dependency declaration in the fallback code path: `libbsd_overlay_dep =  dependency('', required:false)`, which wouldn't be necessary if you build an array of dependencies (extending with the `+=` operator).

ametzler: No worries. I have switched to
```meson
system_deps = []

if ['gnu', 'gnu/kfreebsd'].contains(host_system)
  libbsd_overlay_dep = dependency('libbsd-overlay')
  system_deps += libbsd_overlay_dep
  add_project_dependencies(libbsd_overlay_dep, language: 'c')
endif
```

ueno (approved):
Thank you!

meson.build
```diff
@@ -1,6 +1,6 @@
 project('p11-kit', 'c',
         version: '0.25.3',
-        meson_version: '>= 0.51')
+        meson_version: '>= 0.63')
```
ueno, line 3: CodeQL CI is failing because they only have 0.61.2 installed:
```console
meson.build:1:0: ERROR: Meson version is 0.61.2 but project requires >= 0.63
```
I guess we could stick with `>= 0.51`.

ametzler: I have removed the last commit bumping the meson version requirement. Recent meson ssems to be smart enough to not throw
> meson.build:67: WARNING: Project targets '>= 0.51' but uses feature introduced in '0.63.0': add_project_dependencies.
for inactive code, i.e. the warning is only shown on hosts matching
```
if ['gnu', 'gnu/kfreebsd'].contains(host_system)
```

PR #615, opened 2024-01-05
Add option to specify CKA_ID in generate-keypair and import-object

### LINKED ISSUE #607: p11-kit generate-keypair should allow setting CKA_ID
Sometimes it is useful to keep the ID assignment on the PKCS#11 module, but sometimes the application needs to have the control over the ID so I would provide this option, similarly as we can set the label already.

ZoltanFridrich:
--id option allows users to specify CKA_ID attribute to be associated with imported/generated objects when using p11-kit command. The ID attribute should be specified in hexadecimal format without the "0x" prefix.

Example usage:
`p11-kit import-object --file=cert.pem --id="54:62:70:63:f1:75:84:43:58:8e:d1:16:20:b1:c6:ac:1a:bc:f6:89" --label=cert --login "pkcs11:token=test"`

Closes #607

### DIFF
--- a/common/hex.c
+++ b/common/hex.c
@@ -30,13 +30,18 @@
  * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
  * DAMAGE.
  *
- * Author: Stef Walter <stefw@collabora.co.uk>, Daiki Ueno
+ * Authors: Stef Walter <stefw@collabora.co.uk>,
+ *          Daiki Ueno,
+ *          Zoltan Fridrich <zfridric@redhat.com>
  */
 
 #include "config.h"
+
+#include "debug.h"
 #include "hex.h"
-#include <stdint.h>
+
 #include <stdlib.h>
+#include <string.h>
 
 static const char HEXC_LOWER[] = "0123456789abcdef";
 
@@ -47,6 +52,8 @@ hex_encode (const unsigned char *data,
 	char *result;
 	size_t i;
 	size_t o;
+
+	return_val_if_fail (data != NULL, NULL);
 
 	if ((SIZE_MAX - 1) / 3 < n_data)
 		return NULL;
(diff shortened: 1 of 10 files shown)


common/hex.c
```diff
@@ -64,3 +66,44 @@ hex_encode (const unsigned char *data,
...
+
+		if (i + 1 >= hex_len) {
+			free (bin);
+			return NULL;
+		}
+
+		hex2[0] = hex[i++];
+		hex2[1] = hex[i++];
+
+		val = strtoul (hex2, NULL, 16);
```
ueno, line 98: Using `strtoul` for this simple task seems a bit too much for me, as it requires additional setup (`hex2`) and error handling which would unlikely to happen. Maybe you could write expand something like:
```c
/* Calculate bin_len accurately */
return_val_if_fail (SIZE_MAX - 1 >= hex_len, NULL);
return_val_if_fail ((hex_len + 1) % 3 == 0, NULL);

bin_len = (hex_len + 1) / 3;
...
for (i = 0; i < bin_len; i++) {
  int n = 0;
  for (j = 0; j < 2; j++, n <<= 4) {
    int c = hex[i * 3 + j];
    switch (c) {
    case '0': ... case '9':
      n |= c - '0';
      break;
    case 'a': ... case 'f':
      n |= c - 'a';
      break;
    default:
      error;
    }
  c = hex[i * 3 + j];
  if (c != ':' || c != '\0')
    error;
  bin[i] = n;
}
```

common/hex.c
```diff
@@ -64,3 +66,44 @@ hex_encode (const unsigned char *data,
...
+        size_t hex_len = strlen (hex);
+
+	bin = malloc (hex_len / 2);
+	if (bin == NULL)
+		return NULL;
+
+        for (i = j = 0; i < hex_len;) {
+		if (hex[i] == ':') {
+			++i;
+			continue;
```
ueno, line 87: This is too lax as it accepts colons in an invalid position, e.g., `00::::::::11`.

common/hex.c
ZoltanFridrich: This way it doesnt accept strings without `:`

ueno (approved):
Looks good to me, thanks!

PR #617, opened 2024-01-10
Fix a bug where eddsa mechanism isnt recognized in generate-keypair

ZoltanFridrich:


p11-kit/test-generate-keypair.sh
```diff
@@ -54,16 +54,16 @@ test_generate_keypair_ecdsa() {
 test_generate_keypair_eddsa() {
 	for curve in ed25519 ed448; do
 		if "$abs_top_builddir"/p11-kit/p11-kit-testable generate-keypair --login --label="eddsa-$curve" --type=eddsa --curve="$curve" "pkcs11:token=PUBKEY%20LABEL?pin-value=booo" 2> err.out; then
-			assert_fail "unable to run: p11-kit generate-keypair"
+			assert_fail "expected to fail: p11-kit generate-keypair"
 		fi
 	done
 	assert_contains err.out "key-pair generation failed: The crypto mechanism is invalid or unrecognized"
```
ueno, line 60: I guess this assert should be in the loop, or it will only check the result for the last curve (ed448)? Same applies to the ecdsa test.

ueno (approved):
Good catch, thanks!

PR #624, opened 2024-02-28
rpc: add support for recursive attributes

ZoltanFridrich:
Allow usage of recursive attributes like CKA_WRAP_TEMPLATE inside RPC protocol.

### DIFF
--- a/common/persist.c
+++ b/common/persist.c
@@ -562,6 +562,12 @@ field_to_attribute (p11_persist *persist,
 		return false;
 	}
 
+	/* Ignore recursive attributes */
+	if (IS_ATTRIBUTE_ARRAY (&attr)) {
+		free (attr.pValue);
+		return true;
+	}
+
 	*attrs = p11_attrs_take (*attrs, attr.type,
 	                         attr.pValue, attr.ulValueLen);
 	return true;
--- a/p11-kit/rpc-server.c
+++ b/p11-kit/rpc-server.c
@@ -247,21 +247,15 @@ proto_write_ulong_array (p11_rpc_message *msg,
 }
 
 static CK_RV
-proto_read_attribute_buffer (p11_rpc_message *msg,
-                             CK_ATTRIBUTE_PTR *result,
-                             CK_ULONG *n_result)
+proto_read_attribute_buffer_array (p11_rpc_message *msg,
+				   CK_ATTRIBUTE_PTR *result,
+				   CK_ULONG *n_result)
 {
-	CK_ATTRIBUTE_PTR attrs;
+	CK_RV rv;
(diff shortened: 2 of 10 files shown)


ZoltanFridrich:
Still in progress. But at least the test_wrap_template test works together with other C_GetAttributeValue tests

ZoltanFridrich:
The basic functionality seems to be working but it will probably require more rigorous testing. Especially in test-attrs

p11-kit/rpc-server.c
```diff
@@ -1175,7 +1086,7 @@ rpc_C_GetAttributeValue (CK_X_FUNCTION_LIST *self,
 	BEGIN_CALL (GetAttributeValue);
 		IN_ULONG (session);
 		IN_ULONG (object);
-		IN_ATTRIBUTE_BUFFER (template, count);
```
ueno, line 1178: I'm not sure if it is a good idea to replace "buffer" with "array". The difference between those is that "array" needs all elements to be known and initialized prior to the call, while "buffer" can contain uninitialized element, e.g., `attr->pValue` is set to NULL to retrieve `attr->ulValueLen`.

In any case, their RPC signatures are also different, so if we change "buffer" to "array", we might need a version bump of the RPC protocol and a proper fallback to the previous behavior.

ZoltanFridrich: I have kept the IN_ATTRIBUTE_BUFFER and made it to work with recursive attributes.

common/persist.c
```diff
@@ -562,6 +562,13 @@ field_to_attribute (p11_persist *persist,
 		return false;
 	}
 
+	/* Ignore recursive attributes */
+	if (IS_ATTRIBUTE_ARRAY (&attr)) {
```
ueno, line 566: Can we simply return `false` rather than proceeding with a NULL attribute?

ueno (approved):
Looks like a clean fix, thank you!

common/persist.c
ZoltanFridrich: No, because the p11-kit-testable binary is using module-ep12 which at the beginning loads and at exit persists all objects into .p11-kit file. This is used in some tests to preserve the objects across multiple calls to p11-kit-testable. However as the mock_reset_objects function always creates the object with CKA_WRAP_TEMPLATE, the p11-kit-testable will always write this object at exit into .p11-kit file. If I were to make it fail here, it would break the tests because module setup will fail during reading of .p11-kit file. So I made it to just ignore the CKA_WRAP_TEMPLATE value.

ueno: In that case, can we return `true` without adding it to the array?

ZoltanFridrich: yes, we can.

PR #627, opened 2024-03-04
Fix some compiler warnings

ZoltanFridrich:


p11-kit/rpc-client.c
```diff
@@ -788,7 +788,7 @@ rpc_C_Initialize (CK_X_FUNCTION_LIST *self,
 		if (ret != CKR_OK) {
 			assert (module->vtable->disconnect != NULL);
 			(module->vtable->disconnect) (module->vtable, reserved);
-			ret = (module->vtable->connect) (module->vtable, reserved);
+			(module->vtable->connect) (module->vtable, reserved);
```
ueno, line 791: Should we actually check the return value?

ZoltanFridrich: if the connect wont succeed then the next call fails anyway so I am not sure. maybe.

ZoltanFridrich: actually, lets check the return value

ueno (approved):

PR #628, opened 2024-03-12
log: Avoid buffer overflow

Jakuje:
There was attempt to fix this in bb3a3cb1, but it fixed only the case when this value was actually provided by the token. When the token does not support clock, it would print behind the "" buffer, causing random constant string to be printed instead of the actual value, for example

	utcTime: ../common/attrs

This simplifies the long line for readability and prints additional information about this value being unsupported by the token.


ueno (approved):
Good catch, thanks!

PR #629, opened 2024-03-14
rpc: check value length before writing the value

ZoltanFridrich:
This is to prevent buffer overflow in case the value buffer is too short to store the decoded value.


ueno (approved):
Looks good to me!

PR #631, opened 2024-03-18
rpc: remove length assert in proto_read_attribute_buffer_array

ZoltanFridrich:


ueno (approved):
Should we try to avoid integer overflow (wrap around this case) at multiplication? Maybe adding a check like `ULONG_MAX / length < sizeof (CK_ATTRIBUTE)` might help.

ZoltanFridrich:
> Should we try to avoid integer overflow (wrap around this case) at multiplication? Maybe adding a check like `ULONG_MAX / length < sizeof (CK_ATTRIBUTE)` might help.

I think you meant `ULONG_MAX / n_array < sizeof (CK_ATTRIBUTE)`. Thats a good idea. Fixed.

PR #633, opened 2024-03-21
Fix division by zero in proto_read_attribute_buffer_array

ZoltanFridrich:


ueno (approved):
Thanks for the quick fix!

Issue #523, opened 2023-06-25
FR: API for getting the code version.


bcsgh:
Many libraries provide an API for getting the version (or an approximation there of) of the library that is being used. Often this comes in two flavors; the static version (e.g. a `#define P11_VERSION` macro that allows inspecting what version of the headers a build included) and a dynamic version (i.e. a `p11_check_version()` function that allows inspecting what version of the library was linked in or dynamically loaded).

---
Allowing client code to know what version of the library is being used can enable a number of nice-to-have capabilities;
- Better status reporting.
- Better ability to identify if a dependent build or installation includes know issue.
- Better visibility of of what includes, libraries and shared libraries are actual selected in a specific case.

Issue #518, opened 2023-06-20
Can p11-kit be used in a hermetic build?


bcsgh:
I found [this](https://github.com/p11-glue/p11-kit/issues/355) which implies but doesn't prove the answer is no. I'm hoping that's not the case or that there is another API comparable library I can use in it's place. 

For context, I am trying to use p11-kit in a context where I must not make any assumptions about the environment that are not guaranteed by the fact the OS got to the point of attempting to start my binary. There is a begrudging allowance for the dynamic library that contains the OS sys-calls, but that's only because there is no way to even get to trying to run my binary if that's not usable and sys-calls have to interact with the *current* OS version so it can't be statically linked anyway.

Nether of these applies to something like p11-kit; using a dynamic library could result in load failures if the binary file is copied onto a system without that library or could result in it running on a version other than the one build as part of the hermetic build if another version is present. I need to eliminate both these possibilities.

If it makes any difference in my case, I'd be willing to manually call an officially supported public init API to make things work.

---
**Side note:** I'm guessing that the current state is a result of module registration and the requirement for external libraries to call into a registration function simply by being linked in as well as some ordering constraints between them and p11-kit's internal initialization. As it happens, at least in C++11 (thought presumably the details would would different for C), I personally know this is 100% possible to do in a static library because I've implemented it myself. It ends up looking something like this:

```
struct Registrar { ... };

Registrar* get_registrar() {
  static Registrar* ret = construct_init_registrar();  // See C++11 standard 6.7.4 re concurrent execution
  return ret;
}

template<void (*fn)(Registrar*)>
class Register {
 public:
  Register() { fn(get_registrar()); }
};

// in module:
namespace { Register<my_init> init; } 
```

The only remaining trick then is to get the linker to not throw away the module, but I have reason to believe this is [a problem with know but finiky solutions](https://stackoverflow.com/questions/1202494/why-doesnt-attribute-constructor-work-in-a-static-library).

PR #635, opened 2024-04-26
p11-kit: Expose version information through macros

ueno:
This provides P11_KIT_VERSION_{MAJOR,MINOR,MICRO} from the <p11-kit/version.h> header, along with a helper macro P11_KIT_CHECK_VERSION(major, minor, micro) for checking the version requirement with `#if` directive.

Fixes: #523

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,7 @@ frob-*
 /p11-kit/pkcs11.conf.example
 /p11-kit/p11-kit-server.service
 /p11-kit/proxy-generated.h
+/p11-kit/version.h
 /p11-kit/virtual-*-generated.h
 /p11-kit/virtual-fixed-wrappers.h
 /p11-kit/virtual-fixed-closures.h
--- a/configure.ac
+++ b/configure.ac
@@ -638,11 +638,20 @@ AC_SUBST(GENHTML)
 P11KIT_LT_RELEASE=$P11KIT_CURRENT:$P11KIT_REVISION:$P11KIT_AGE
 AC_SUBST(P11KIT_LT_RELEASE)
 
-echo $PACKAGE_VERSION | tr '.' ' ' | while read major minor unused; do
-	AC_DEFINE_UNQUOTED(PACKAGE_MAJOR, $major, [Major version of package])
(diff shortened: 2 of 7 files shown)


ZoltanFridrich (approved):
Looks good.

ueno:
Thanks for the review!

bcsgh:
Why is `P11_KIT_CHECK_VERSION` a macro rather than a function? That kind of removes one of the reason for including this sort of API which is to check _at runtime_ that the library linked in (either linked statically at build time or dynamically from whatever environment things are run in) matches the expectation of the calling code. This macro at most allows calling code the verify that the set of headers resolved at build time are the expected ones but doesn't, by itself, provide a way to know what version of the p11 library the code is then linked with. 

Given that [p11 seems to require non-static builds](https://github.com/p11-glue/p11-kit/issues/518), it seems like being able to verify that the (unavoidable) `.so` that is chosen at run time is what the calling code expects it to be (or at least having a way to know what it actually is) would be at least as desirable here as it is for other libraries that provide that sort of function:

- Asn1 [`asn1_check_version`](https://linux.die.net/man/3/asn1_check_version)
- IDN2 [`idn2_check_version`](https://linux.die.net/man/3/idn2_check_version)
- Nettle [`nettle_version_major`/`nettle_version_minor`](https://github.com/gnutls/nettle/blob/master/version.h.in)
- Curl [`curl_version_info`](https://curl.se/libcurl/c/curl_version_info.html)
- Enki [`ev_version_major`/`ev_version_mnor`](http://pod.tst.eu/http://cvs.schmorp.de/libev/ev.pod#GLOBAL_FUNCTIONS)
- Lua [`lua_version`](https://www.lua.org/manual/5.4/manual.html#lua_version)
- And others (those are just the ones I found in a few minutes in code I wrote).

ueno:
I actually thought that run-time detection was already possible through PKCS#11 API (because `libp11-kit.so` is the same binary as `p11-kit-proxy.so`), but it turns out [not](https://github.com/p11-glue/p11-kit/blob/652927da18abecedf3ab21728252ec73e8214174/p11-kit/proxy.c#L106) the case. Maybe we could embed the actual version.

PR #637, opened 2024-05-07
Add function to check run-time version of the library

ueno:
This adds p11_kit_check_version function, which can be used to check the run-time version of p11-kit.

Fixes: #523

### DIFF
--- a/p11-kit/util.c
+++ b/p11-kit/util.c
@@ -44,6 +44,7 @@
 #include "message.h"
 #include "p11-kit.h"
 #include "private.h"
+#include <p11-kit/version.h>
 
 #include <assert.h>
 #include <stdarg.h>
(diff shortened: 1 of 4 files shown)


p11-kit/util.c
```diff
@@ -236,3 +237,22 @@ _p11_get_progname_unlocked (void)
...
+ * @minor: minor version
+ * @micro: micro version
+ *
+ * Check if the run-time version of p11-kit meets the requirement
+ * given by a triple: @major, @minor, and @micro.
+ *
+ * Returns: %CKR_OK if the version requirement meets; %CKR_ARGUMENTS_BAD otherwise.
+ * Since: 0.25.4
+ */
+CK_RV
```
ZoltanFridrich, line 253: Would it be better if the return value was `bool` as it literally returns a binary state?

ueno: Afaik we intentionally do not expose `bool` from the public API for compatibility reasons.

ZoltanFridrich: how about int?

ueno: Yeah, we use `int` in many places so we can do that here too.

ZoltanFridrich (approved):
Looks good

ueno:
Thanks for the review!

ametzler:
Looking at 0.25.5 `p11_kit_check_version()` is part of the source but afaict it is dead code. version.h with the function prototype is not installed and the function is not part of the exported API either since it is not listed in p11-kit/libp11-kit.map.

ueno:
@ametzler indeed; thank you for the report. Could you check #648?

PR #639, opened 2024-06-28
fix format warning with 32-bit platforms on musl

neheb:
musl uses 64-bit time_t, even on 32-bit platforms. Cast to avoid the warning.


ueno (approved):
Thank you!

Issue #641, opened 2024-07-02
Test failure on 32-bit


heftig:
Since 0.25.4, a build using `gcc -m32` (on x86_64) fails the `test-enumerate` test: 

```
1..17
ok 1 /extract/test_file_name_for_label
ok 2 /extract/test_file_name_for_class
ok 3 /extract/test_comment_for_label
ok 4 /extract/test_comment_not_enabled
ok 5 /extract/test_info_simple_certificate
ok 6 /extract/test_info_limit_purposes
ok 7 /extract/test_info_invalid_purposes
p11-kit: 'templ[i].ulValueLen != 0' not true at prepare_recursive_attribute
```

backtrace:

```c++
#0  0xf7ddc579 in __kernel_vsyscall ()
#1  0xf7a5a88f in __pthread_kill_implementation (threadid=threadid@entry=4154181888, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:43
#2  0xf7a5a91f in __pthread_kill_internal (threadid=4154181888, signo=6) at pthread_kill.c:78
#3  0xf7a00287 in __GI_raise (sig=6) at ../sysdeps/posix/raise.c:26
#4  0xf79e7121 in __GI_abort () at abort.c:79
#5  0xf7d461e6 in p11_debug_precond (format=0xf7d4900d "p11-kit: '%s' not true at %s\n") at ../common/debug.c:190
#6  0xf7c4eaa5 in prepare_recursive_attribute (iter=0x5f020530, attr=0x5f022310, templ=0x92395010, templ_len=1700946284) at ../p11-kit/iter.c:1045
#7  0xf7c4ebd1 in prepare_recursive_attribute (iter=0x5f020530, attr=0x5f022310, templ=0x5f025cb0, templ_len=13) at ../p11-kit/iter.c:1052
#8  0xf7c4ebd1 in prepare_recursive_attribute (iter=0x5f020530, attr=0x5f022310, templ=0x5f027800, templ_len=36) at ../p11-kit/iter.c:1052
#9  0xf7c4f05d in p11_kit_iter_load_attributes (iter=0x5f020530, templ=0x5f021f20, count=149) at ../p11-kit/iter.c:1141
#10 0x5e1882e7 in extract_info (ex=0x5e1b4d7c <test+284>) at ../trust/enumerate.c:309
[...]

PR #642, opened 2024-07-02
iter: Fix recursive attribute loading

heftig:
Fixes: https://github.com/p11-glue/p11-kit/issues/641


ZoltanFridrich:
Thank you for the MR. I will test it on fedora before merging.

ZoltanFridrich (approved):

PR #647, opened 2024-07-05
build: Set LC_ALL=C for test suite

xry111:
Otherwise the error messages may be translated, causing a match failure.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -83,7 +83,9 @@ AM_TESTS_ENVIRONMENT = \
 	abs_top_srcdir="$(abs_top_srcdir)"; \
 	export abs_top_srcdir; \
 	P11_MODULE_PATH="$(abs_top_builddir)/.libs"; \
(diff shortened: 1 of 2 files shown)


p11-kit/test-generate-keypair.sh
```diff
@@ -10,6 +10,9 @@ test "${abs_top_builddir+set}" = set || {
 : ${P11_MODULE_PATH="$abs_top_builddir"/.libs}
 
 setup() {
+	LC_ALL=C
+	export LC_ALL
```
ueno, line 14: It might make more sense to set this globally, e.g., in `p11_kit_tests_env` in Meson and `AM_TESTS_ENVIRONMENT` in Autotools?

xry111: Let me try...

ueno (approved):
Thank you!

Issue #649, opened 2024-07-20
On PPC Tiger, old Mac OS X 10.4.11, test target does not build with GCC 4.2


ballapete:
Hello!

Other compiler option would be GCC 7.5.

Version of p11-kit is 0.25.5.

The problem is that CK_ULONG and CK_SESSION_HANDLE get redefined. I prepared the attached patch set. With it applied the default and the test target build.

See also: https://trac.macports.org/ticket/70399.

--

Pete

[Patches-for-test-target-on-Tiger.diff.zip](https://github.com/user-attachments/files/16321420/Patches-for-test-target-on-Tiger.diff.zip)

ueno:
Thank you for the report; it indeed looks like a bug. I've filed #653 to fix it.

PR #653, opened 2024-08-02
p11-kit: Move CRYPTOKI_GNU definitions to <p11-kit/p11-kit.h>

ueno:
Previously, the same fallback types are defined in `<p11-kit/uri.h>` and `<p11-kit/iter.h>` if `CRYPTOKI_GNU` is defined. This moves those to `<p11-kit/p11-kit.h>` so they are not redefined.

Fixes: #649

### DIFF
--- a/p11-kit/iter.h
+++ b/p11-kit/iter.h
@@ -45,16 +45,6 @@ extern "C" {
 
 #ifdef P11_KIT_FUTURE_UNSTABLE_API
 
-/*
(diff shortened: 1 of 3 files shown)


ZoltanFridrich (approved):
I can't verify whether it works, but it looks good.

ueno:
> I can't verify whether it works, but it looks good.

Me neither, but with `CFLAGS="-std=gnu99 -pedantic"`, the warning is gone.

ueno:
Thanks for the review!

PR #656, opened 2024-08-28
token: sort paths for reproducible extract

katexochen:
There is no defined order in which readdir will return the entries of a directory. In practice, order can depend on inode number or similar. If we run p11-kit on different files systems with similar directory structure but different inode order the output of extract can change.

To get a stable and reproducible output, sort the paths returned by readdir before extracting.

### DIFF
--- a/trust/token.c
+++ b/trust/token.c
@@ -253,15 +253,27 @@ loader_load_if_file (p11_token *token,
 }
 
 static int
+compar_strings (const void *one,
+                const void *two)
+{
+	const char **p1 = (const char **)one;
+	const char **p2 = (const char **)two;
+	return strcmp (*p1, *p2);
+}
+
+
+static int
 loader_load_directory (p11_token *token,
                        const char *directory,
                        p11_dict *present)
 {
 	p11_dictiter iter;
+	p11_array *paths;
 	struct dirent *dp;
 	char *path;
 	int total = 0;
 	int ret;
+	int i;
 	DIR *dir;
 
 	/* First we load all the modules */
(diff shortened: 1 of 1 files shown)


trust/token.c
```diff
@@ -272,11 +283,25 @@ loader_load_directory (p11_token *token,
 		return 0;
 	}
 
+	paths = p11_array_new (NULL);
```
ZoltanFridrich, line 286: This should have an error guard as this can return NULL.

trust/token.c
```diff
@@ -252,12 +252,23 @@ loader_load_if_file (p11_token *token,
 	return 0;
 }
 
+static int
+compar_strings (const void *one,
+                const void *two)
+{
+	char **p1 = (char **)one;
```
ZoltanFridrich, line 259: Can p1 and p2 be const?

trust/token.c
```diff
@@ -272,11 +283,25 @@ loader_load_directory (p11_token *token,
...
 
+	paths = p11_array_new (NULL);
+
 	while ((dp = readdir (dir)) != NULL) {
 		path = p11_path_build (directory, dp->d_name, NULL);
 		return_val_if_fail (path != NULL, -1);
 
-		ret = loader_load_if_file (token, path);
+		if (!p11_array_push (paths, path)) {
+			free (path);
```
ZoltanFridrich, line 293: A simple `return_val_if_fail` will suffice as `p11_array_push` will only fail in case of OOM in which case the program dies anyway.

trust/token.c
```diff
@@ -272,11 +283,25 @@ loader_load_directory (p11_token *token,
...
+			return -1;
+		}
+	}
+
+    closedir(dir);
+
+    qsort(paths->elem, paths->num, sizeof(char *), compar_strings);
+
+    for (int i = 0; i < paths->num; i++) {
```
ZoltanFridrich, line 302: Please move the declaration of `i` at the start of the function.

trust/token.c
```diff
@@ -272,11 +283,25 @@ loader_load_directory (p11_token *token,
...
 		return_val_if_fail (path != NULL, -1);
 
-		ret = loader_load_if_file (token, path);
+		if (!p11_array_push (paths, path)) {
+			free (path);
+			return -1;
+		}
+	}
+
+    closedir(dir);
```
ZoltanFridrich, line 298: Formatting needs fixing. You are mixing TABs and spaces + put space between function name and brackets.

ZoltanFridrich (changes requested):
Thank you for your contribution!
The change looks good overall, but I would like to see some things fixed before I can merge it.

katexochen:
> The change looks good overall, but I would like to see some things fixed before I can merge it.

Thanks for your feedback, I tried to incorporate the requested changes. Let me know if I overlooked something. :)

ZoltanFridrich (approved):
Looks good!

ZoltanFridrich:
@katexochen please rebase it to master so I can merge.

PR #657, opened 2024-09-11
doc/manual: Fix chapter id collision: proxy/sharing

jca-klk:
The chapter id in p11-kit-proxy.xml collides with the one in p11-kit-sharing.xml, which makes the HTML generated page overwriting each other.

This results in duplicate page content for both "Sharing PKCS#11 modules" section and "Proxy Module" section. Both provide content from p11-kit-proxy.xml and make p11-kit-sharing.xml content unavailable to the reader.

Changing the chapter id fixes the issue.


ueno (approved):
Good catch, thanks!

PR #659, opened 2024-10-03
trust: don't create file names longer then 255

ZoltanFridrich:
`trust extract` creates file names based on CKA_LABEL attribute of trusted certificates. Some of these certificates contain labels that are longer then file name limit on linux which causes file creation to fail.
This MR makes `make_unique_name` to trim names that are equal or longer then 255.

### DIFF
--- a/trust/save.c
+++ b/trust/save.c
@@ -61,6 +61,8 @@
 #define O_DIRECTORY 0
 #endif
 
+#define MAX_FILE_NAME 255
+
 struct _p11_save_file {
 	char *bare;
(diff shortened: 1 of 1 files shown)


ueno (commented):
Do we have any assurance that the truncated filename doesn't collide with others? Perhaps we could fall back to other naming scheme, e.g., hash, if the filename would be longer than 255?

ueno (approved):
LGTM, looks like the uniqueness is ensured and also the generated file name ends with a NUL character. Thanks!

PR #661, opened 2024-10-11
server: check SHELL if (and only if) neither --sh nor --csh is specified

alanc:
The previous check was checking if the enums returned by getopt were non-zero, which they always are, regardless of whether the option has been specified or not.

Fixes: e7d1c91 ("server: Check SHELL only when neither --sh nor --csh is specified") from #438


ueno (approved):
Good catch, thank you!

PR #667, opened 2024-12-31
avoid using already defined thread_local as variable name

brechtsanders:
Building p11-kit 0.25.5 with GCC15 on MinGW-w64 failed because `thread_local` is already defined for this platform.

Resolved by changing the variable name from `thread_local` to `threadlocal`.

### DIFF
--- a/common/library.c
+++ b/common/library.c
@@ -124,7 +124,7 @@ _p11_library_get_thread_local (void)
 	return &local;
 }
 #else
-static pthread_key_t thread_local = 0;
+static pthread_key_t threadlocal = 0;
 
 static p11_local *
 _p11_library_get_thread_local (void)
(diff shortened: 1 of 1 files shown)


kalvdans:
GCC 15 has not been released yet, could you confirm the compiler version?

kalvdans (approved):
`thread_local` was added as a keyword in C23:

https://en.wikipedia.org/wiki/C23_(C_standard_revision)#Keywords

Instead of specifying `-std=c17` I agree renaming a variable is best. (I have no merge rights)

kalvdans:
Thinking more about it, we should instead (or in addition to the rename) make sure `P11_TLS_KEYWORD` gets defined on GCC15.

brechtsanders:
Yes, I was testing a snapshot version of GCC15 ad it will default to `-std=c23` causing some changes that break a lot of software builds.

See also: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118283

ueno (approved):
LGTM, thank you!

PR #669, opened 2025-01-29
Add IBM specific mechanisms

fcallies:
This series adds support for three IBM specific mechanisms and four IBM specific attributes.
Additionally for the rpc_C_DeriveKey function support for size queries for potential output fields of the mechanisms parameters is implemented.


fcallies:
@ueno can one of the maintainers please approve the check workflows?

p11-kit/rpc-message.h
```diff
@@ -224,7 +225,7 @@ static const p11_rpc_call p11_rpc_calls[] = {
 	{ P11_RPC_CALL_C_GenerateKeyPair,      "C_GenerateKeyPair",      "uMaAaA",  "uu"                   },
 	{ P11_RPC_CALL_C_WrapKey,              "C_WrapKey",              "uMuufy",  "ay"                   },
 	{ P11_RPC_CALL_C_UnwrapKey,            "C_UnwrapKey",            "uMuayaA", "u"                    },
-	{ P11_RPC_CALL_C_DeriveKey,            "C_DeriveKey",            "uMuaA",   "u"                    },
+	{ P11_RPC_CALL_C_DeriveKey,            "C_DeriveKey",            "uMuaA",   "uPu"                    },
```
ueno, line 228: This breaks compatibility at the protocol level; I would suggest bumping the protocol version and provide a proper fallback, as in b2dd8d5c73dab8e0f074b32d7bb393817964fb1d.

ueno (commented):
Do you have any documentation available for those mechanisms?

p11-kit/rpc-message.c
```diff
@@ -1660,6 +1668,386 @@ p11_rpc_buffer_get_ibm_attrbound_wrap_mechanism_value (p11_buffer *buffer,
...
+						   void *value,
+						   CK_ULONG *value_length)
+{
+	uint64_t val1;
+	uint64_t val2;
+	size_t len;
+	const unsigned char *data;
+	uint64_t val3;
+
+	if (!p11_rpc_buffer_get_uint64 (buffer, offset, &val1))
```
ZoltanFridrich, line 1905: You can combine these into a single `if` which would look much nicer, something like:
```
if (!p11_rpc_buffer_get_uint64 (buffer, offset, &val1) ||
    !p11_rpc_buffer_get_uint64 (buffer, offset, &val2) ||
     ...)
      return false;
```
Same can be done in `p11_rpc_buffer_get_ibm_kyber_mechanism_value`

ZoltanFridrich (changes requested):
Looks nice overall except the compatibility break reported by Daiki. That needs to be addressed.

fcallies:
You can find the documentation for the mechanisms [here](https://www.ibm.com/docs/en/linux-on-systems?topic=opencryptoki-specific-mechanisms).

fcallies:
I pumped the RPC protocol version to 2 and included the requested changes by @ZoltanFridrich and I hope I fixed all the Ci bugs. It looks like they were originating in the meson build failing.

common/pkcs11x.h
```diff
@@ -188,6 +188,8 @@ typedef CK_ULONG                        CK_TRUST;
 #define CKK_IBM_PQC_DILITHIUM    CKK_VENDOR_DEFINED + 0x10023
 
 #define CKA_IBM_OPAQUE                         (CKA_VENDOR_DEFINED + 1)
+#define CKA_IBM_OPAQUE_REENC                   (CKA_VENDOR_DEFINED + 3)
+#define CKA_IBM_OPAQUE_OLD                     (CKA_VENDOR_DEFINED + 4)
```
kalvdans, line 192: I fail to see these mentioned in the documentation on https://www.ibm.com/docs/en/linux-on-systems?topic=opencryptoki-specific-mechanisms , remove.

p11-kit/rpc-client.c
```diff
@@ -706,6 +765,17 @@ proto_read_sesssion_info (p11_rpc_message *msg,
...
+#define PREP_EEOF \
+	CK_RV err;
+
+#define PROCESS_CALL_NO_CHECK \
+		_ret = call_run (_mod, &_msg);
+
+#define CHECK_EEOF \
+	if(err != CKR_OK) \
+		_ret = err;
+
```
kalvdans, line 778: These macros are only used in one place, code will be more readable without the use of macros.

fcallies: Since all of the rpc_C_ Functions make heavy use of macros I actually think this is more readable because mixing macros and plain code where almost exclusively macros are used is not helpful. Additionally it eases possible extensions of other rpc_C_ functions that could use the param update feature because they can just make use of the existing macros. If of course the majority agrees with you I will adjust the code.

common/pkcs11x.h
fcallies: Unfortunately the documentation is not up to date here but an update is currently worked on. As reference here the opencryptoki commit that added those: [link](https://github.com/opencryptoki/opencryptoki/commit/20f3cafe5c75b3ab6a4ef0a2e74888d2dea5996b)

kalvdans: Thanks for the link, I'll yell at the people who reviewed that PR.

kalvdans: Yelled at https://github.com/opencryptoki/opencryptoki/pull/659/files#r1946274303 but it would be nice if you could paste that documentation in here as a comment.

fcallies:
The documentation requested by @kalvdans is added as a comment and code should be ready for another CI run.

ZoltanFridrich:
@fcallies check the cppcheck failure. It seems to be complaining about uninitialized variables.
The autotools-no-asn1 failure should not be relevant and will probably be fixed after I rerun the test.

ZoltanFridrich:
> This should be a false positive as these values are set using the `p11_rpc_buffer_get_` functions. As requested by @ZoltanFridrich I put them all in one `if`, maybe thats what confuses the test... The first of the values set in the `if` is not reported, but all the values after that are reported as uninitialized.

Indeed that might be the cause because I dont think the cppcheck understands that all of the `p11_rpc_buffer_get_` functions must succeed in order to continue to the assignment of the values. You can expand it to multiple `if`s as previously and I can run the tests again.

fcallies:
I split it again in different `if`s @ZoltanFridrich

fcallies:
Please restart the CI tests.

ZoltanFridrich:
> Please restart the CI tests.

seems like it wasnt the issue afterall. Maybe just initialize the struct with initializer like `params = { 0 };`?

fcallies:
I initialized params and on my local machine the cppcheck succeeds. Ready for another CI run.

fcallies:
Lgtm, is there anything else to be done by me?

ZoltanFridrich (approved):
LGTM

PR #671, opened 2025-02-03
Update pkcs11.h to version 3.1

ZoltanFridrich:
Spec: https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/cs01/pkcs11-spec-v3.1-cs01.html


p11-kit/test-mock.c
```diff
@@ -2737,7 +2737,7 @@ test_mock_add_tests (const char *prefix, const CK_VERSION *version)
 	p11_test (test_derive_key, "%s/test_derive_key", prefix);
 	p11_test (test_random, "%s/test_random", prefix);
 	/* PKCS #11 3.0 tests */
-	if (version && version->major == 3 && version->minor == 0) {
```
kalvdans, line 2740: The comment is inaccurate now, because you removed the test for 3.0 and added tests for 3.1.

That said, shouldn't we test that both 3.0 and 3.1 works?

ZoltanFridrich: The 3.0 test is testing functions that were introduced in the pkcs11 3.0 spec. The 3.1 spec did not introduce any new functions. I have just removed the minor version restriction as 3.1 has the same functions as 3.0 and therefore it makes sense to run the same tests.

kalvdans (approved):

ueno (approved):
Looks good to me, given it can compile GnuTLS with CRYPTOKI_GNU.

PR #672, opened 2025-02-04
Update to v4 of actions/upload-artifact

ZoltanFridrich:


ueno (approved):
LGTM, might also want to enable dependabot, as in https://github.com/latchset/crypto-auditing/blob/efa33c56431063c98aceedfb076df4bb4d926ef2/.github/dependabot.yml#L13

PR #674, opened 2025-03-04
Add zsh completions

rpigott:
So far only bash completions were defined. Uses trust list to complete pkcs11 uris for --filter.


ZoltanFridrich (approved):
Looks nice, thanks!

PR #681, opened 2025-03-28
Improve reliability of header file generation using context manager

Doekin:
This PR addresses an issue where the `gen-fixed-closures.py` script fails to write to `virtual-fixed-closures.h` on Windows, causing `p11-kit` to not build properly. The problem occurs when using Meson 1.7.0 and Python 3.13.2.

The following minimal reproduction example demonstrates the issue:

```python
import argparse

def foo(output):
    output.write("OK\n")

if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--outfile", type=argparse.FileType("w"))
    args = parser.parse_args()
    foo(args.outfile)
```

Running `meson runpython foo.py --outfile t.h` does not write "OK" to `t.h`.


kalvdans (approved):
This is in line with the recommendation at https://docs.python.org/3/library/argparse.html#type
Approve! (but I have no merge rights)

ueno (approved):
Thank you; I wonder if other scripts (e.g., `p11-kit/gen-wrappers.py`) need the same treatment?

Doekin:
The other scripts seem to work fine at the moment, but I'm not entirely confident they'll remain consistent. I'll go ahead and update them as well.

PR #682, opened 2025-04-04
Avoid segfault if p11_library_init_impl/p11_library_uninit are called multiple times

duramuss:
On my test environment test-server.sh fails with a SIGSEGV because p11_library_init_impl/p11_library_uninit are called multiple times. This PR allows test-server.sh to pass.


kalvdans (approved):

ueno (approved):
Thank you; the change makes sense.

PR #683, opened 2025-04-17
Fix definition of NEED_READPASSPHRASE on non-unix

myzhang1029:
On non-unix platform (e.g. mingw), the Automake conditional is not defined `NEED_READPASSPHRASE` by `configure`, which causes
```
configure: error: conditional "NEED_READPASSPHRASE" was never defined.
Usually this means the macro was only invoked conditionally.
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -229,6 +229,9 @@ if test "$os_unix" = "yes"; then
 	                AC_MSG_RESULT([yes])],
 	               [AC_MSG_RESULT([no])])
 else
+    # ($os_unix != "yes")
+    # termios-based passphrase query is only supported on Unix platforms
+    # See common/Makefile.am
     AM_CONDITIONAL([NEED_READPASSPHRASE], [false])
 fi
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -228,7 +228,8 @@ if test "$os_unix" = "yes"; then
 	                          [Whether vsock support available])
 	                AC_MSG_RESULT([yes])],
 	               [AC_MSG_RESULT([no])])
-
+else
+    AM_CONDITIONAL([NEED_READPASSPHRASE], [false])
```
ueno, line 232: I would add some comment explaining why this can be set to `false` on Windows (i.e., passphrase query on terminal is only supported on Unix platforms).

ueno (approved):
Thank you; looks good to me.

myzhang1029:
Thank you. I have added a justification.
Unfortunately, even after this patch, cross-compiling with autotools for a mingw host is still failing. As I do not have a straightforward solution, I'll open an issue for this.

PR #685, opened 2025-04-30
server: Remove libsystemd dependency for socket activation

ueno:
While libsystemd provides rich facilities to interact with systemd, the server uses it only for socket activation, which is now also supported by other service managers. This avoids the dependency by embedding the minimal implementation of socket activation.


ZoltanFridrich (approved):
I think it looks ok. I haven't found anything wrong.

PR #701, opened 2025-08-18
Update sl.po

filmsi:
Updated Slovenian translation as they do not get updated from transifex.


filmsi:
@ueno Can you please help?

ueno (approved):
Thanks; I can't really review the translation update, but trust you :-)

PR #705, opened 2025-09-04
rpc: Correctly map Mozilla certificate distrust attributes

ueno:
The attributes used for expressing certificate distrust in NSS were incorrectly mapped to a byte array.


FrantisekKrenzelok (approved):
Everything looks good.
Only one concern: not being for the commit message written so well, I wouldn't expect there to be a default switch statement, i understand the placement but can we add a comment to the end that it is indeed present.

debarshiray:
> I suppose you are using different p11-kit versions inside/outside the container as this breaks the RPC protocol compatibility

Oops!  Yes, I haven't updated my host OS for a few weeks, and I was only playing with the updates inside the container.

PR #730, opened 2025-10-26
p11-kit-client.dll for Windows

### LINKED ISSUE #405: p11-kit-client.dll for Windows
Hi,

I would like to use the p11-kit-client.dll with native Windows. it is included only in the Cygwin build but not within the native MSYS2 builds. As far as I can see, this is by intention. The p11-kit/meson.build skips it on line 90.
Is there any specific reason?
I would like to access a central Hardware Token containing several SSH Keys on a Windows Workstation. Microsofts own Windows SSH does support PKCS11 providers and this is working directly with native Windows Softhsm2 or Yubikey PKCS11 Dlls. It would be great to have it with p11-kit PKCS11 as well.

I've attached a patch file, I can compile the dll with a MSYS2 MinGW 64 environment: [p11kitclient-win32-diff.txt](https://github.com/p11-glue/p11-kit/files/7833564/p11kitclient-win32-diff.txt)

TSlivede:
This fixes #405.

The change to `p11-kit/client.c` is exactly as originally suggested by @spfendtner in #405.

The changes in `p11-kit/rpc-message.c` change the serialization of attributes on systems where size of unsigned long is not 8 bytes. As far as I know, this should only be the case for Windows and some uncommon linux variants (32 bit MIPS?).
With this change the wireprotocol is always identical and allows to connect with a p11-kit-client.dll to a p11-kit server running on Linux.

The length field for ULONG attributes is now always serialized as value 8.
In principle there is no need to serialize the length at all for ULONG attributes,
as the only allowed length value here is sizeof(CK_ULONG) anyway.
But by always serializing as 8 the wire protocoll does only change on the
few platforms where `sizeof(CK_ULONG)!=8`.
As there never was a p11-kit-client on Windows, I hope this is ok.

As was commented in the issue, `LDFLAGS+=" -lintl"` was required to build on windows, so I added `libintl_deps` to the dependencies of the mock_sources in `p11-kit/meson.build`, as some of those were failing without.

The `shared_module('p11-kit-client',` call in `p11-kit/meson.build` is the same as before, just two spaces indentation removed, as I completely removed the `if host_system != 'windows'` condition.

### DIFF
--- a/p11-kit/rpc-message.c
+++ b/p11-kit/rpc-message.c
@@ -1333,6 +1333,36 @@ p11_rpc_buffer_get_byte_array_value (p11_buffer *buffer,
 	return true;
 }
 
+static inline bool
+p11_rpc_buffer_get_attr_val_len_by_val_type(p11_buffer *buf,
+                                            size_t *offset,
+                                            p11_rpc_value_type value_type,
+                                            uint32_t *length)
+{
+	if (!p11_rpc_buffer_get_uint32 (buf, offset, length))
+		return false;
+
+	/* Length of ULONG attribures is always sizeof(CK_ULONG) which can be 4 or 8 depending on the platform.
+	 * For interoperability always override it to the size on the current platform when deserializing. */
+	if (length != NULL && *length != 0 && value_type == P11_RPC_VALUE_ULONG){
+		assert (*length == 4 || *length == 8);
+		*length = sizeof(CK_ULONG);
+	}
+
+	return true;
+}
+
+bool
+p11_rpc_buffer_get_attr_val_length(p11_buffer *buf,
+                                   size_t *offset,
+                                   CK_ATTRIBUTE_TYPE attr_type,
+                                   uint32_t *length)
+{
+	p11_rpc_value_type value_type;
+	value_type = map_attribute_to_value_type (attr_type);
+	return p11_rpc_buffer_get_attr_val_len_by_val_type (buf, offset, value_type, length);
+}
+
 bool
 p11_rpc_message_get_attribute (p11_rpc_message *msg,
 			       p11_buffer *buffer,
@@ -1361,7 +1391,8 @@ p11_rpc_message_get_attribute (p11_rpc_message *msg,
 		return true;
 	}
 
-	if (!p11_rpc_buffer_get_uint32 (buffer, offset, &length))
+	value_type = map_attribute_to_value_type (type);
+	if (!p11_rpc_buffer_get_attr_val_len_by_val_type (buffer, offset, value_type, &length))
 		return false;
 
 	if (length == 0) {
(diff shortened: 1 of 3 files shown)


p11-kit/rpc-message.c
```diff
@@ -244,13 +244,17 @@ p11_rpc_message_verify_part (p11_rpc_message *msg,
...
+map_attribute_to_value_type(CK_ATTRIBUTE_TYPE type);
+
 static void
 p11_rpc_message_write_attribute_buffer_array (p11_rpc_message *msg,
 					      CK_ATTRIBUTE_PTR arr,
 					      CK_ULONG num)
 {
 	CK_ATTRIBUTE_PTR attr;
 	CK_ULONG i;
+	uint32_t valueLenSerialized;
```
ueno: Use snail case instead of camel case: `value_len_serialized`.

p11-kit/rpc-message.c
```diff
@@ -264,7 +268,11 @@ p11_rpc_message_write_attribute_buffer_array (p11_rpc_message *msg,
 		p11_rpc_buffer_add_uint32 (msg->output, attr->type);
 
 		/* And the attribute buffer length */
-		p11_rpc_buffer_add_uint32 (msg->output, attr->pValue ? attr->ulValueLen : 0);
+		valueLenSerialized = attr->pValue ? attr->ulValueLen : 0;
+		/* For ULONG attribures we always serialize the lenght as 8 even if it is not 8 on the current platform */
+		if (sizeof(CK_ULONG) != 8 && valueLenSerialized == sizeof(CK_ULONG) && map_attribute_to_value_type(attr->type) == P11_RPC_VALUE_ULONG)
+			valueLenSerialized = 8;
```
ueno: Maybe good to have a helper function for this? Something like:
```c
static uint32_t
p11_rpc_buffer_add_value_length (p11_rpc_buffer *buffer, p11_rpc_value_type type, CK_ULONG value_length);
```

ueno (commented):
Thank you; the logic looks correct. I just added minor nit picks which are nice to be fixed before merging it.

TSlivede:
Short update:
I changed this PR such that only on receiving the ULONG attribute length is deserialized as the correct value of the platform. In my test this did not yet work, though I did not yet debug why. Will post update in a few days.

Though at least this version does (as intended) seem to not break any existing working configuration based on my short tests.


I kept the previous version of this PR [in this branch](https://github.com/p11-glue/p11-kit/compare/90dba4d...TSlivede:p11-kit:feature_p11_kit_client_ulong_len_normalized) for reference.

TSlivede:
Ok, seems to work now based on my testing.

@ueno I completely changed this PR's implementation of the attribute value length handling. Short summary of why and what i changed:

Instead of normalizing the attribute length to 8 when serializing and doing the opposite on deserializing, I modified this PR to only change how the attribute value length is deserialized. For attributes of value type ULONG the length is now always overwritten to the current platform's `sizeof(CK_ULONG)`.

I changed it to not normalize on serialization to avoid becoming incompatible to existing clients/servers on linux x32 (as that apparently also has `sizeof(long)==4`). This way the wire protocol does not change, but if both server and client have the version ignoring the sent length for ULONG attributes, then communication is possible between systems with different ULONG sizes. (Though 32bit ULONG client only works with 64bit ULONG server if the used pkcs11 module does not generate to big CK_OBJECT_HANDLE values.)

PR #744, opened 2026-03-09
Fix client compatibility with older servers

ozbenh:
We changed the format of the mechanism parameter without actually checking the server version. This broke SignInit() when using older servers.

### DIFF
--- a/p11-kit/rpc-client.c
+++ b/p11-kit/rpc-client.c
@@ -419,7 +419,8 @@ mechanism_list_purge (CK_MECHANISM_TYPE_PTR mechs,
 
 static CK_RV
 proto_write_mechanism (p11_rpc_message *msg,
-                       CK_MECHANISM_PTR mech)
+                       CK_MECHANISM_PTR mech,
+		       int rpc_version)
 {
 	assert (msg != NULL);
 	assert (msg->output != NULL);
@@ -433,7 +434,10 @@ proto_write_mechanism (p11_rpc_message *msg,
 	 * marker to indicate that.
 	 */
 	if (mech == NULL) {
-		p11_rpc_buffer_add_uint32 (msg->output, 0xffffffff);
+		if (rpc_version < 1)
+			p11_rpc_buffer_add_uint32 (msg->output, 0);
+		else
+			p11_rpc_buffer_add_uint32 (msg->output, 0xffffffff);
 		return p11_buffer_failed (msg->output) ? CKR_HOST_MEMORY : CKR_OK;
 	}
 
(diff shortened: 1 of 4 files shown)


ozbenh:
This is my attempt at fixing it. I havent messed around with p11-kit much in the past so bear with me if this isn't the right approach. 

It fixes our Secure Boot signing infrastructure in Amazon Linux (we use an older server on our builders with the newer client potentially inside the buildroot and a bind mounted socket between the two).

ozbenh:
The gcov test seems to be failing for unrelated reasons...

ueno (commented):
Have you identified which commit changed the behavior? If yes, good to mention that in the commit message. I hope that, it would be nice if the different implementations are provided through different functions so they can be compile out with `#ifdef` if desired. Otherwise, I would make the RPC version part of `p11_rpc_message` to avoid passing it as a parameter.

ozbenh:
> Have you identified which commit changed the behavior? If yes, good to mention that in the commit message.

You are right, I forgot to do that.  The main culprits are:

6449cfe59b80344c87832c9e38afcb4df3f61c56 "Rework rpc_C_DeriveKey to allow size queries"
and
66f1fc7b8f60693e0947d5a9d38cd0d1bc306baa "rpc: Fix serialization of NULL mechanism pointer"

As far as I can tell. I'll fix up the commit message in my next spin

> I hope that, it would be nice if the different implementations are provided through different functions so they can be compile out with `#ifdef` if desired. Otherwise, I would make the RPC version part of `p11_rpc_message` to avoid passing it as a parameter.

Any particular reason you dislike passing it as a parameter ?

One option could be to have two variants of p11_rpc_buffer_add_mechanism() and two variants of proto_write_mechanism() and put the selection of which one to call inside IN_MECHANISM(). I though that was a bit more bloated ... but I'm happy to respin along those lines if that's what you want.

ozbenh:
actually on second thought ... there are two changes here and one is from protocol 0 to 1 (roughly, the change happened in the middle of 1) and the other one is 1 to 2. So we would need *3* variants of both functions above. It becomes a bit crazy don't you think ? IMHO it's cleaner to pass the version down, but we could add ifdef's for conditional compilation inside them if you think it matters. I doubt in practice that it makes any significant difference in either performance or code size though.

ozbenh:
Looking at putting the version in the message, that doesn't really help with `p11_rpc_buffer_add_mechanism`...

I suppose one "compromise" could be to have the version in the message, which avoids the argument from `IN_MECHANISM` to `proto_write_mechanism`, and then two variants of `p11_rpc_buffer_add_mechanism`.
I'm not convinced the end result is much nicer though :-) What do you reckon ?

ozbenh:
FYI. I will be away for 3 weeks so don't worry if I don't respond for a bit to your comments when you post them :)

PR #747, opened 2026-03-27
fuzz: add three new harnesses

DavidKorczynski:


ueno:
@DavidKorczynski thank you for the PR and sorry for the shameless delay. As for the CI failures: the one in "test / coverage" is known, while the other in "CIFuzz / Fuzzing" looks new and probably needs a fix in p11-kit itself?

fuzz/conf_fuzzer.c
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (c) 2026, Red Hat Inc.
```
ueno: I suspect you are not assigning copyright to Red Hat, or are you? Could you credit yourself otherwise?

ueno (changes requested):
This looks very useful; could you update the copyright notice if needed, so we can merge it?

fuzz/conf_fuzzer.c
DavidKorczynski: Adjusted!

ueno (approved):
Thank you!

PR #752, opened 2026-05-20
fuzz: build conf, uri, and persist fuzzers with meson

ueno:


meson.build
```diff
@@ -480,13 +480,13 @@ pkcs11_json = pkcs11_json_gen.process('common/pkcs11.h')
...
 subdir('p11-kit')
-subdir('fuzz')
 if with_trust_module
   subdir('trust')
 endif
 subdir('doc/manual')
 if get_option('nls')
   subdir('po')
 endif
+subdir('fuzz')
```
ZoltanFridrich, line 490: What does this change do?

ueno: This ordering change makes sure that `libtrust-testable` is built before linking fuzzers under `fuzz/`.

ZoltanFridrich: ah, ok then

ZoltanFridrich (approved):
LGTM

PR #753, opened 2026-05-20
common/asn1.c: memory management hardening

ZoltanFridrich:
Fixes a bunch of potential memory leaks and overall hardens the memory management in common/asn1.c


ueno (approved):
LGTM

PR #757, opened 2026-05-21
ci: Update CI image to Fedora 44

ueno:


ueno:
CI only change; merging without approval

PR #762, opened 2026-05-26
persist: fix parse_string corner-case

ZoltanFridrich:
Function parse_string incorrectly interpreted the
value (") as string, which triggered the
assert (value <= end) in function p11_url_parse


ueno (commented):
Can we have a test exercising the issue?

ZoltanFridrich:
> Can we have a test exercising the issue?

Isn't the fuzzer test enough for this?

ueno:
In that case, we might want to update the corpora in git repo, like https://gitlab.com/gnutls/gnutls/-/blob/2cb190c8c6ea5876c8cfd6a2840323959e7a1a89/fuzz/README.md#L13 ?

ueno (approved):
LGTM, thanks for updating the corpus.

PR #763, opened 2026-06-09
p11-kit/uri.c: memory management hardening

ZoltanFridrich:
Hardens the memory management + fixes int overflow and timeout issues found by the fuzzer


ueno (approved):
LGTM. Although all changes are small, it would be nice if you could summarize them in the commit log.

ZoltanFridrich:
added description to commit message

PR #765, opened 2026-06-12
Reject numerical attribute type that matches CKA_INVALID

ZoltanFridrich:
* fixes a memory leak when later on attrs_build did not free the memory because type was CKA_INVALID
* added new fuzzer testcase


ueno (approved):
LGTM

PR #766, opened 2026-06-15
Free previously allocated objects when parsing .p11-kit file fails

ZoltanFridrich:
- added destructor when calling `p11_array_new()` in `p11_parser_format_persist()`. This way the array is properly freed when error occurs during parsing
- added fuzzer testcase


ueno (approved):
Looks like a nice way to fix the leak. LGTM.

PR #769, opened 2026-07-08
parser: fix memory leak in build_openssl_extensions

ZoltanFridrich:
Fixes a new OSS-fuzz find.


trust/parser.c
```diff
@@ -414,6 +414,8 @@ attached_eku_attrs (p11_parser *parser,
 	return attrs;
 }
 
+/* Call to this function always invalidates pointer to `cert`
```
ueno, line 417: OK, so this function amends `cert` with trust attributes and thus needs to take the ownership. What about `public_key_info`? Maybe it's clearer if it is declared as `const CK_ATTRIBUTE *public_key_info` so it doesn't take ownership.

ueno (approved):
LGTM, with a minor nit.

Issue #778, opened 2026-08-22
p11-kit server hides every post-quantum mechanism a module offers


afchine1337:
## The short version

`p11_rpc_mechanism_is_supported()` is an allow-list of mechanism types, and it
has not kept up with PKCS#11. The effect is not a degraded transport; it is a
mechanism that is simply invisible to the client, with no diagnostic anywhere.

You can confirm the shape of this without running anything. In
`p11-kit/rpc-message.c`, `mechanism_has_no_parameters()` on current master
lists:

- `CKM_SHA224`, `CKM_SHA512_256` — standard digests
- `CKM_IBM_SHA3_224` … `CKM_IBM_SHA3_512`, `CKM_IBM_SHA3_*_HMAC`
- `CKM_IBM_DILITHIUM`, `CKM_IBM_KYBER`

and does not list `CKM_SHA3_224`, `CKM_SHA3_256`, `CKM_SHA3_384`,
`CKM_SHA3_512`, `CKM_ML_DSA`, `CKM_ML_KEM` or `CKM_SLH_DSA`.

So IBM's vendor SHA-3 crosses the socket and the standard SHA-3 does not.
SHA-3 has been in PKCS#11 since v3.0; ML-DSA, ML-KEM and SLH-DSA since v3.2.

The comment above the switch already says `/* This list is incomplete */`.
This is a report of what that incompleteness now costs, not a complaint that
it exists.

## What it looks like from a client

Measured with a PKCS#11 module that implements post-quantum signatures,
p11-kit 0.24.0 (Ubuntu jammy):

| | mechanisms reported by `C_GetMechanismList` |
|---|---|
| module loaded directly | 72 |
| same module through `p11-kit server` | **20** |

52 dropped. Among them every mechanism the module exists to provide.
`C_SignInit` with one of them returns `CKR_MECHANISM_INVALID`, and the module
never receives the call — confirmed from the server side, where the module's
audit log records the login and nothing after it.

Two properties of this that make it awkward to diagnose:

1. **`C_GetMechanismList` is filtered, so the mechanism is not merely
   unusable — it is invisible.** An application enumerating mechanisms to
   decide what it can do concludes, correctly as far as it can tell, that the
   token does not offer it.
2. **`C_GetMechanismInfo` also answers `CKR_MECHANISM_INVALID`**, through
   `IN_MECHANISM_TYPE`, even though that call carries a bare mechanism type
   and serialises no parameter at all.

## Why now rather than at any point in the last five years

[...]

PR #779, opened 2026-08-22
rpc: carry mechanisms invoked without a parameter

afchine1337:
Follows #778.

The RPC layer decides what it can serialise from the mechanism **type**. This
changes it to decide from the **call**, which is a different and easier
question, and lets every parameterless invocation across — post-quantum
mechanisms and vendor mechanisms included — without loosening what the
allow-list was built to protect.

One commit, four parts. The last one changes behaviour for existing users and
I am happy to split it into its own commit, or drop it, on request; the first
three stand without it.

## The encoder already does the hard part

Worth saying first, because it makes the change smaller than it sounds.
`p11_rpc_buffer_add_mechanism()` already encodes an absent parameter as a
presence byte of `0`, and `p11_rpc_buffer_get_mechanism()` already reads that
byte back for *any* mechanism, recognised or not. The two halves are symmetric
today.

The only thing between them is `assert (mechanism_has_sane_parameters (...))`,
which aborts for a type the allow-list does not name — including one invoked
with no parameter, where the encoding would be that same presence byte. This
patch lets that case take the path that already exists. **The decoder is not
touched.**

## 1. Mechanisms that never take a parameter

`CKM_SHA3_224/256/384/512`, their `_HMAC` and `_KEY_GEN` forms, and
`CKM_ML_DSA_KEY_PAIR_GEN`, `CKM_ML_KEM_KEY_PAIR_GEN`,
`CKM_SLH_DSA_KEY_PAIR_GEN`, added to `mechanism_has_no_parameters()`.

Safe by inspection: none takes a parameter in any mode, so nothing is written
after the mechanism type and nothing is read back — exactly what every entry
already in that switch does.

**Deliberately not added: `CKM_ML_DSA`, `CKM_SLH_DSA`, and the `CKM_HASH_*`
family.** Those take an *optional* `CK_SIGN_ADDITIONAL_CONTEXT`. Listing them
as parameterless would encode a supplied context as absent, and the server
would return a signature computed over an empty context: well-formed, verifying
against nothing the caller expects, and silent. **A wrong signature is a worse
failure than a refused call.** Part 2 carries them instead.

I mention this because it is the trap in the obvious version of this patch, and
the reason it is not simply "add the new mechanisms to the list".

## 2. Decide from the call, not the type

```c
bool
p11_rpc_mechanism_call_is_supported (const CK_MECHANISM *mech)
{
        if (mech == NULL)
                return false;
        if (p11_rpc_mechanism_is_supported (mech->mechanism))
                return true;
        return mech->pParameter == NULL && mech->ulParameterLen == 0;
}
```

The allow-list exists because a `CK_MECHANISM` parameter is a `void *` whose
layout depends on the mechanism, so an unrecognised type cannot be encoded.
That reasoning does not reach a call which supplied no parameter: the encoding
is the presence byte the decoder already understands for every mechanism, and
there is nothing left to misencode.

`proto_write_mechanism()` uses it.

This is also strictly safer than enumerating a mechanism with an optional
parameter: with `CKM_ML_DSA` on the list of part 1, a context would be dropped;
with this test, the call is refused.

Demonstrated in both directions against a module implementing `CKM_ML_DSA`:

```
  CKM_ML_DSA, no parameter      C_SignInit -> CKR_OK
  CKM_ML_DSA, with a parameter  C_SignInit -> CKR_MECHANISM_INVALID
```

— refused at the proxy while the module itself accepts it. That is the
boundary this patch draws, and I think it is the right one.

## 3. The encoder stops asserting on an unrecognised type

Two lines: where `p11_rpc_buffer_add_mechanism()` currently asserts, a type
that is in neither list now writes the presence byte `0` and returns. It can
only have reached that point through part 2, which means the call carried no
parameter, which means that byte is the whole of it.

## 4. `C_GetMechanismList` stops being trimmed — the debatable part

`mechanism_list_purge()` removes mechanisms this file cannot serialise. That
was sound while the answer depended only on the type. After part 2 it no
longer does: a mechanism this file has never heard of may be entirely usable,
and trimming hides it.

Measured before the change: a module advertising 72 mechanisms reported 20
through the socket.

The trade is between two imperfect behaviours. Advertising a mechanism whose
parameterised form will fail at `C_*Init` is ordinary PKCS#11 — modules refuse
advertised mechanisms for key type, session state and policy reasons already.
Advertising nothing, for a mechanism the module does offer and the proxy can
carry, is a false statement about the token that the application has no way to
question.

`IN_MECHANISM_TYPE` is relaxed for a narrower reason: it carries a bare
mechanism type for `C_GetMechanismInfo`, serialises no parameter, and so had
nothing to protect.

**If you would rather keep the trimming, or make it conditional, say so and I
will split it out** — parts 1 to 3 stand on their own: `C_SignInit` would work
for a caller that knows the mechanism value, and enumeration would keep today's
behaviour.

## Testing

- `meson test` on this branch — **44 tests, 44 pass, 0 failures**, and the same
  44 pass on unmodified master in the same build directory. Base:
  `120050e` (Release 0.26.5), Debian 13, gcc.
- Through the patched socket, a module advertising 72 mechanisms reports 72.
- A composite ML-DSA-65 + Ed25519 signature produced through the socket
  **verifies against the same module loaded directly**, and a single flipped
  byte in it is rejected — so the verification is discriminating.
- The negative case above: unknown mechanism plus parameter is still refused.

## Notes

- The end-to-end measurements (72 mechanisms, the composite signature) were
  taken against 0.24.0, which is what the environment where this showed up
  could build. This branch is against master and its suite is green there.
- No test added to your suite yet. The natural shape is a mock module offering
  a vendor mechanism, asserted to cross without a parameter and be refused with
  one — I will write it in whatever style you prefer, and would rather match
  your conventions than guess.

### DIFF
--- a/p11-kit/rpc-client.c
+++ b/p11-kit/rpc-client.c
@@ -396,25 +396,25 @@ static void
 mechanism_list_purge (CK_MECHANISM_TYPE_PTR mechs,
                       CK_ULONG *n_mechs)
 {
-	CK_ULONG i;
-
 	assert (mechs != NULL);
 	assert (n_mechs != NULL);
 
-	/* Trim unsupported mechanisms at the end */
-	for (; *n_mechs > 0 && !p11_rpc_mechanism_is_supported (mechs[*n_mechs - 1]); --*n_mechs)
-		;
-
-	for (i = 0; i < *n_mechs; ++i) {
-		if (!p11_rpc_mechanism_is_supported (mechs[i])) {
-			/* Remove the mechanism from the list */
-			memmove (&mechs[i], &mechs[i + 1],
-				 (*n_mechs - (i + 1)) * sizeof (CK_MECHANISM_TYPE));
-
-			--(*n_mechs);
-			--i;
-		}
-	}
+	/*
+	 * C_GetMechanismList used to be trimmed to the mechanisms this file
+	 * knows how to serialise. That was sound while the decision could be
+	 * taken from the type alone. It cannot be any more: whether a call can
+	 * be carried depends on whether the caller supplies a parameter, so a
+	 * mechanism this file has never heard of may be entirely usable.
+	 *
+	 * Trimming therefore hid working mechanisms. Measured against a module
+	 * offering post-quantum signatures: 72 advertised, 20 reported through
+	 * the socket, and the 52 hidden included every one the module existed
+	 * to provide.
+	 *
+	 * A mechanism that is advertised and then fails at C_*Init is ordinary
+	 * PKCS#11. A mechanism the module offers and the proxy makes invisible
+	 * is a false statement about the token.
+	 */
 }
 
 static CK_RV
@@ -437,7 +437,7 @@ proto_write_mechanism (p11_rpc_message *msg,
 		return p11_buffer_failed (msg->output) ? CKR_HOST_MEMORY : CKR_OK;
 	}
 
-	if (!p11_rpc_mechanism_is_supported (mech->mechanism))
+	if (!p11_rpc_mechanism_call_is_supported (mech))
 		return CKR_MECHANISM_INVALID;
 
 	/*
@@ -683,9 +683,9 @@ proto_read_sesssion_info (p11_rpc_message *msg,
 	if (!p11_rpc_message_write_attribute_array (&_msg, (arr), (num))) \
 		{ _ret = CKR_HOST_MEMORY; goto _cleanup; }
 
+/* Carries a bare mechanism type, for C_GetMechanismInfo and friends:
+ * no parameter is serialised here, so there was nothing to protect. */
 #define IN_MECHANISM_TYPE(val) \
-	if(!p11_rpc_mechanism_is_supported (val)) \
-		{ _ret = CKR_MECHANISM_INVALID; goto _cleanup; } \
 	if (!p11_rpc_message_write_ulong (&_msg, val)) \
 		{ _ret = CKR_HOST_MEMORY; goto _cleanup; }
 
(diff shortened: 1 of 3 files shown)


afchine1337:
I have just found #745, which has been open since March and covers a good deal
of this. That is my fault for not searching before opening; apologies to
@mingulov for the duplication.

Read together, I think the overlap is smaller than it looks and the two are
complementary:

**#745 does better** — the mechanism enumeration itself, and far more besides:
`generate-keypair`, import/export, attribute value-type mapping, and real
integration tests against kryoptic and NSS. Nothing here improves on that.

**What is left in this PR that #745 does not cover:**

1. **Standard `CKM_SHA3_*`.** #745 adds the PQC families; the SHA-3 digests and
   HMACs, in PKCS#11 since 3.0, are still absent from the list — while
   `CKM_IBM_SHA3_*` is present.

2. **Mechanisms nobody enumerated.** This is the substantive difference.
   #745 extends the allow-list; this PR changes the *question* the allow-list
   answers, from "is this type known" to "can this call be serialised", which
   is true whenever the caller supplied no parameter. That carries vendor
   mechanisms — the case that motivated me, a module using a private mechanism
   value for composite ML-DSA + Ed25519 signatures — and it will keep carrying
   the next standard mechanism without anyone editing a list.

3. **`C_GetMechanismList` stops being trimmed** (part 4), which is separable
   and which I am happy to drop.

I have also left a comment on #745 about `CKM_ML_DSA` and the `CKM_HASH_*`
family, which I believe should not be on the no-parameter list. That is a
question about that PR rather than a claim about this one.

Happy to rebase this on top of #745 once it lands, reduced to parts 1, 2 and 3,
if the maintainers prefer that order. #745 is the bigger piece of work and
should go first.

PR #782, opened 2026-09-18
Fix list-objects garbage print via RPC

ZoltanFridrich:
In list-objects, any attribute that returned CK_UNAVAILABLE_INFORMATION was not dropped and instead the macro was used as length in the second call to p11_kit_iter_get_attributes which resulted in an out of memory error and garbage output in the printing.


ueno (approved):
Good catch, LGTM.

